Issue #5, opened 2019-08-30
Stuck At Please log in under Azure


bariscaglar:
I'm not sure where to go from here. 

![image](https://user-images.githubusercontent.com/1233208/64048060-7b0a1e80-cb25-11e9-8886-4f7c65a32b38.png)

MarkusRodler:
Same problem & no documentation

ilya-spv:
The issue is still legit, took time to figure out what I had to do. Ran it in terminal as well :)

There probably can be a bit better explanation or clickable link or something in there, though it seems it's already handled in the linked PR

PR #6, opened 2019-09-11
Refactor Azure Cloud Explorer Contribution

philliphoff:
Updates the Azure contribution the Kubernetes extension's Cloud Explorer to be based on the common Azure subscription node UX, making sign-in, subscription selection, and visualization consistent across the various Azure-based explorers.

Should help with #5.

Sign in UX:

Screen Shot 2019-09-11 at 2 59 04 PM

Select subscriptions UX:

Screen Shot 2019-09-11 at 2 54 07 PM

Enumeration and filter UX:

Screen Shot 2019-09-11 at 2 53 35 PM

Selecting the filter (funnel) button will display the common Azure subscription selection/filter list. Changes will be immediately reflected in the Cloud Explorer.


src/extension.ts
```diff
@@ -2,31 +2,50 @@ import * as vscode from 'vscode';
 import * as k8s from 'vscode-kubernetes-tools-api';
 import * as azcs from 'azure-arm-containerservice';  // deprecated, but @azure/arm-containerservice doesn't play nicely with AzureAccount, so...
 
-import { AKSTreeProvider, AKSClusterTreeNode } from './aks-tree';
```
daniv-msft: >./aks-tree [](start = 53, length = 10)

Do we still need this aks-tree.ts file, or could we remove it as well?

src/tree/AksClusterTreeItem.ts
```diff
@@ -0,0 +1,23 @@
+import { AzExtParentTreeItem, AzureTreeItem } from "vscode-azureextensionui";
+import { Resource } from "azure-arm-storage/lib/models";
+import { CloudExplorerV1 } from "vscode-kubernetes-tools-api";
+
+export default class AkClusterTreeItem extends AzureTreeItem {
```
daniv-msft: >AkClusterTreeItem [](start = 21, length = 17)

It seems that moving from AKSClusterTreeNode to this new class creates a breaking change for extensions having a dependency on vscode-aks-tools.
For example, in our case we are reading the "session" object from the commandTarget, that is now being replaced by "root".
Our approach is very similar to the one documented here: [https://github.com/Azure/vscode-kubernetes-tools/blob/master/docs/extending/cloudexplorer.md#implementing-your-own-commands-for-cloud-resources](https://github.com/Azure/vscode-kubernetes-tools/blob/master/docs/extending/cloudexplorer.md#implementing-your-own-commands-for-cloud-resources)

This might be fine if no other extensions are taking a dependency on this extension (and we have some time to push a fix for our extension :) ), but I don't know whether it is the case.

src/extension.ts
philliphoff: Yes, I had intended to do that and forgotten; thanks for catching that!

src/tree/AksClusterTreeItem.ts
philliphoff: That's a good point; at the time, I didn't realize other extensions were building atop `vscode-aks-tools`.  (I don't see any other references in GitHub, but Dev Spaces isn't currently OSS and I haven't looked at its codebase in several years, so I would have missed it in any case.)

I could try to make it backward compatible with the existing API.  You could also make future updates to the Dev Spaces extension pinned to the new version of `vscode-aks-tools`, to ensure it always works.  That doesn't help older versions of Dev Spaces extension which end up with the new version of the AKS extension, obviously.

philliphoff: I've made updates that map (most of) the properties defined in the old interfaces to their respective newer properties.  This appears to resolve the issue with the existing Dev Spaces extension.  The properties that are not mapped did not appear to be exposed directly by the subscription/resource nodes, but also appear to be the more esoteric ones related to the subscription itself that I hope are low risk for breaking back-compat.

Let me know if this seems more reasonable.

daniv-msft: That looks good to me indeed. I also validated that, with the new version of your changes, our extension works as expected. Thanks for fixing this quickly!

package.json
```diff
@@ -36,20 +63,21 @@
...
     "devDependencies": {
-        "typescript": "^2.6.1",
-        "vscode": "^1.1.6",
-        "tslint": "^5.8.0",
+        "@types/js-yaml": "^3.12.1",
+        "@types/mocha": "^2.2.42",
         "@types/node": "^7.0.43",
-        "@types/mocha": "^2.2.42"
+        "tslint": "^5.8.0",
+        "typescript": "^2.6.1",
```
itowlson: Suggest we upgrade to pin a more recent version of TypeScript - with the new dependency, I'm getting errors `Cannot find name 'unknown'` in `vscode-azureextensionui` - we should switch to TypeScript 3.x or above.  (Which I should have done anyway in the first place!)

src/tree/AksClusterTreeItem.ts
```diff
@@ -0,0 +1,41 @@
+import { AzExtParentTreeItem, AzureTreeItem, ISubscriptionContext } from "vscode-azureextensionui";
+import { Resource } from "azure-arm-storage/lib/models";
+import { CloudExplorerV1 } from "vscode-kubernetes-tools-api";
+import { SubscriptionClient } from 'azure-arm-resource';
+import { toSubscription } from "../azure-api-utils";
+
+export default class AkClusterTreeItem extends AzureTreeItem {
```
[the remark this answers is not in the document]
itowlson: Actually... I would prefer `Node` to `Item`, because `vscode.TreeItem` refers specifically to the visual rendering of a data object in the tree, and if the data object calls itself a `TreeItem` then I find it harder to be clear whether the code is talking about the data object or the rendering.  But the standard library uses the `Item` suffix for the base type, so if that is the standard then I'm happy to go along with it!

src/tree/AksClusterTreeItem.ts
```diff
@@ -0,0 +1,41 @@
...
+            this.id = this.resource.id;
+        }
+
+    public readonly contextValue: string = `aks.cluster ${CloudExplorerV1.SHOW_KUBECONFIG_COMMANDS_CONTEXT}`;
+
+    public get label(): string {
+        return this.name;
+    }
+
+    public get armId(): string {
```
itowlson: Is it worth documenting that these properties exist for compatibility?  I wasn't expecting anyone to build on this so didn't define an API for it, so when DevSpaces did their work they had to use the internal structure at the time as a de facto API.  Now it's no longer obvious what purpose these properties serve, since they just rename base class properties.  A comment explaining about the de facto API might help avoid accidents with downstream consumers in future refactorings/tidyings!

src/tree/SubscriptionTreeItem.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
+        const client = new ResourceManagementClient(this.root.credentials, this.root.subscriptionId);
+        const aksClusterResources = await listAll(client.resources, client.resources.list({ filter: "resourceType eq 'Microsoft.ContainerService/managedClusters'" }));
+
+        return aksClusterResources.map((aksClusterResource) => new AksClusterTreeItem(this, aksClusterResource));
+    }
+
+    public get name(): string {
```
itowlson: Same comment about documenting these as being part of a de facto backward compatibility API.

src/extension.ts
```diff
@@ -2,31 +2,50 @@ import * as vscode from 'vscode';
...
 import { parseResource } from './azure-api-utils';
-
-const explorer = new AKSTreeProvider();
+import AksClusterTreeItem from './tree/AksClusterTreeItem';
+import AzureAccountTreeItem from './tree/AzureAccountTreeItem';
+import { createTelemetryReporter, registerUIExtensionVariables, AzExtTreeDataProvider, AzureUserInput, registerCommand } from 'vscode-azureextensionui';
+import selectSubscriptions from './commands/selectSubscriptions';
 
 export async function activate(context: vscode.ExtensionContext) {
+    const ext = {
```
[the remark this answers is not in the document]
philliphoff: It's boilerplate configuration for the Azure UI extension which handles the subscription UX.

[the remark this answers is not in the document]
philliphoff: Updated.

src/tree/AksClusterTreeItem.ts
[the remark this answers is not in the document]
philliphoff: While the VS Code API does keep a strict separation between the model and the rendered tree item that represents it, the Azure UI extension does not (I assume for the sake of simplicity).  So, I'd tend to its `TreeItem` naming convention.

Good catch on the missing `s` in the class name, though.  (I tend to follow .NET naming guidelines, hence the `AksCluster` rather than `AKSCluster`; let me know if you'd really prefer the latter.)

src/tree/AzureAccountTreeItem.ts
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    public refreshImpl?(): Promise<void> {
+        // NOTE: Updates to the subscription filter would normally refresh this node. However,
+        //       the Cloud Explorer wraps this node with its own and doesn't listen for change
+        //       events. Hence, we must force Cloud Explorer to refresh, which will then re-
+        //       enumerate this node's children.
+        vscode.commands.executeCommand('extension.vsKubernetesRefreshCloudExplorer');
+
+        return Promise.resolve();
```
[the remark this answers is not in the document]
philliphoff: It was mostly to be explicit; I don't have a huge preference either way.  The issue became moot after switching to `CloudExplorerV1.refresh()` (comments above), as getting to that API requires asynchrony.

src/tree/AzureAccountTreeItem.ts
```diff
@@ -0,0 +1,19 @@
...
+    public createSubscriptionTreeItem(root: ISubscriptionContext): SubscriptionTreeItemBase {
+        return new SubscriptionTreeItem(this, root);
+    }
+
+    public refreshImpl?(): Promise<void> {
+        // NOTE: Updates to the subscription filter would normally refresh this node. However,
+        //       the Cloud Explorer wraps this node with its own and doesn't listen for change
+        //       events. Hence, we must force Cloud Explorer to refresh, which will then re-
+        //       enumerate this node's children.
+        vscode.commands.executeCommand('extension.vsKubernetesRefreshCloudExplorer');
```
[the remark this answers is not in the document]
philliphoff: Seems to work; updated.

src/extension.ts
```diff
@@ -2,31 +2,50 @@ import * as vscode from 'vscode';
...
+        context,
+        ignoreBundle: !/^(false|0)?$/i.test(process.env.AZCODE_DOCKER_IGNORE_BUNDLE || ''),
+        outputChannel: vscode.window.createOutputChannel('Azure Identity'),
+        reporter: createTelemetryReporter(context),
+        ui: new AzureUserInput(context.globalState)
+    };
+
+    context.subscriptions.push(ext.outputChannel);
```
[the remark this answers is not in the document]
philliphoff: Seems reasonable; updated.

src/tree/AksClusterTreeItem.ts
philliphoff: Added back the interfaces used in the previous implementation, with a note that it represents a de facto API.

src/tree/SubscriptionTreeItem.ts
philliphoff: Added back the interfaces used in the previous implementation, with a note that it represents a de facto API.

src/tree/AksClusterTreeItem.ts
```diff
@@ -0,0 +1,41 @@
+import { AzExtParentTreeItem, AzureTreeItem, ISubscriptionContext } from "vscode-azureextensionui";
```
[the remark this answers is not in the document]
philliphoff: No strong preference.  I've worked in extensions with either convention (and sometimes both at the same time).

package.json
philliphoff: Updated.

itowlson:
Great stuff - really appreciate this.

PR #12, opened 2019-12-05
Resolve issue with Azure subscription icon.

### LINKED ISSUE #11: Subscription icon does not show in VSIXed extension
With the new login/display experience, subscriptions are meant to now have a 'key' icon.  When running the extension from source, this appears correctly.  But when packaged as a VSIX (or downloaded from the marketplace), the key does _not_ appear and there is a blank gap.  This is cosmetic but would be nice to figure out.

Tentatively assigning this to @philliphoff but if you don't have bandwidth then please assign it back to me!

philliphoff:
Resolves missing Azure subscription icon in the tree view by ensuring the underlying Azure UI extension logic always ignores "bundling" (which this extension does not yet do), rather than ignoring it only when debugging locally.

Resolves #11.


daniv-msft:
Looks good to me! Thank you for investigating and finding the root cause.

PR #22, opened 2020-04-20
Add aks cluster run diagnostics in cloud explorer

Tatsinnit:
This PR is to get thoughts\ideas\review for work and improve the extension functionality to add `Run AKS Diagnostics` feature.

The outcome of this work is displayed from in the 2 screesnshots below:

* `package.config` - Consist setting of command part.
* `detetorDiagnostics.ts` - Consist command logic.
   *  key functs: `detectorDiagnostics` ==> `getAppLensDetectorData` ==> `CreateDetectorWebView`
   * `model\AppLensARMresponse` - Tried to create class out so that I can have come kind of known data sctructure\contract for the jsonData returned.
* `*.css` - This file needs whole lot of work to trim down unwanted css which is copied from the element dom of azure portal.
* `webview html` - This needs cleaning (cleaner html - currently to get the same look and feel it is copy\pasted code form the portal DOM).

* AppLens ARM API is still inprogress towards fixing for `clusterinsights`

## Edit 27th April

3 key things beside small refactors.

* Add `ResourceMgmtClient` implementation and `shelljs` is removed.
    * `Get` call is via `client.resource` and **not** cli.
* Use of `Errorable` type.
    * `Utils` --> `Errorable` type exist.
    * `Interface` --> `AppLensARMresponse` is added.
* Dynamic `html` is added and no longer reside inside `detectorDiagnostics.ts`.

## Edit 18th May

* The extension now uses `network connectivity detector` API, Applens clusterInsight API which we envisioned to use was not designed for consumption from extension, for design reasons which Applens team is aware.

* New and clean HTML and CSS.

* HTML templates with `Handlebar`, along with HTML helper to handle the data which current detector API returns. 

* Other recommended changes and renames.

Thanks heaps in advance.

image

image


Tatsinnit:
:bulb: Thanks /assign: @itowlson

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+            const clusterAppLensData = await getAppLensDetectorData(armId);
+            // Create webview
+            await CreateDetectorWebView(target.value.name, clusterAppLensData);
+        } else {
+          vscode.window.showInformationMessage('Please select kubernetes cluster.');
+        }
+    }
+}
+
+async function CreateDetectorWebView(
```
Tatsinnit, line 32: :bulb: casing: `CreateDetectorWebView` to `createDetectorWebView`

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+    panel.webview.html = getWebviewContent(clusterAppLensData);
+    vscode.window.showInformationMessage("AKS Diagnostics Completed.");
+  } else {
+    vscode.window.showInformationMessage('Selected Cluster has no data returned.');
+  }
+}
+
+async function getAppLensDetectorData(clusterARMId: string): Promise<AppLensARMresponse | void> {
+  // ARM api call.
+  const appLensARMAPI = `az rest -m GET -u "${clusterARMId}/detectors/mcrEndpointUpdate?api-version=2019-08-01&executeChildren=true"`;
```
Tatsinnit, line 59: 🐛 AppLens ARM API is still broken hence using this place holder `mcrEndpointUpdate` detector call for data.

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (vscodeExtensionPath === undefined) {
+      vscode.window.showErrorMessage("No Extension path");
+      // TODO: context.extensionPath is missing at command level.
+      return "CSS file path missing";
+    }
+
+    const stylePathOnDisk = vscode.Uri.file(path.join(vscodeExtensionPath, "src/commands/style/detector.css"));
+    const styleUri = stylePathOnDisk.with({ scheme: 'vscode-resource' });
+
+    return `<!DOCTYPE html>
```
Tatsinnit, line 89: 💡 Is there a way to pass dynamic object with HTML so that all this can be extracted into separate html file? 

* (seems not doable) https://github.com/Microsoft/vscode/issues/62840

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    return appLensARMresponse;
+  }catch (err) {
+    vscode.window.showInformationMessage('Error: ' + err);
+  }
+}
+
+function getWebviewContent(clusterdata: AppLensARMresponse) {
+    const webviewClusterData = clusterdata?.properties;
+    const vscodeExtensionPath = process.env.VSCODE_CWD;
```
Tatsinnit, line 78: 💡 `context.extensionPath` is only avalable level above, hence `process.env.VSCODE_CWD` is used, but I wonder if there is any other way?

src/commands/style/detector.css
```diff
@@ -0,0 +1,386 @@
+html {
```
Tatsinnit, line 1: 💡 self note: this whole `css` file need triming to much smaller minimilistic version.

src/extension.ts
```diff
@@ -7,9 +7,11 @@ import AksClusterTreeItem from './tree/aksClusterTreeItem';
 import AzureAccountTreeItem from './tree/azureAccountTreeItem';
 import { createTelemetryReporter, registerUIExtensionVariables, AzExtTreeDataProvider, AzureUserInput, registerCommand } from 'vscode-azureextensionui';
 import selectSubscriptions from './commands/selectSubscriptions';
+import detectorDiagnostics from './commands/detectorDiagnostics';
 
 export async function activate(context: vscode.ExtensionContext) {
     const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+
```
Tatsinnit, line 14: 💡 remove extra line.

src/commands/detectorDiagnostics.ts
itowlson: We still need to call this placeholder endpoint with the Azure JS client using the logged in account credentials, not through the Azure CLI.

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from "vscode-azureextensionui";
+import { AppLensARMresponse } from './models/detectorAppLens';
+import * as path from 'path';
+
+const util = require('util');
+const exec = util.promisify(require('child_process').exec);
```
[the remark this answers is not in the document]
itowlson: (Though it should be moot anyway once you go to the Azure JS client.)

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+  return clusterAppLensData;
+}
+
+async function runAksShellCommand(azcomand: string): Promise<AppLensARMresponse | void> {
+  try {
+    const { stdout } = await exec(azcomand);
+    const appLensARMresponse = new AppLensARMresponse(JSON.parse(stdout));
+
+    return appLensARMresponse;
+  }catch (err) {
```
itowlson, line 71: Brace spacing (consider installing the tslint extension... it should catch this stuff).

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    // Build webview from data.
+    panel.webview.html = getWebviewContent(clusterAppLensData);
+    vscode.window.showInformationMessage("AKS Diagnostics Completed.");
+  } else {
+    vscode.window.showInformationMessage('Selected Cluster has no data returned.');
+  }
+}
+
+async function getAppLensDetectorData(clusterARMId: string): Promise<AppLensARMresponse | void> {
```
itowlson, line 57: Prefer `null` or `undefined` to `void` here.

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+  // ARM api call.
+  const appLensARMAPI = `az rest -m GET -u "${clusterARMId}/detectors/mcrEndpointUpdate?api-version=2019-08-01&executeChildren=true"`;
+  const clusterAppLensData = await runAksShellCommand(appLensARMAPI);
+
+  return clusterAppLensData;
+}
+
+async function runAksShellCommand(azcomand: string): Promise<AppLensARMresponse | void> {
+  try {
+    const { stdout } = await exec(azcomand);
```
itowlson, line 67: Are you skipping error checking here because you expect to switch over to the Azure JS client?

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+}
+
+function getWebviewContent(clusterdata: AppLensARMresponse) {
+    const webviewClusterData = clusterdata?.properties;
+    const vscodeExtensionPath = process.env.VSCODE_CWD;
+
+    if (vscodeExtensionPath === undefined) {
+      vscode.window.showErrorMessage("No Extension path");
+      // TODO: context.extensionPath is missing at command level.
```
itowlson, line 82: Pass in the ExtensionContext.

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+  }catch (err) {
+    vscode.window.showInformationMessage('Error: ' + err);
+  }
+}
+
+function getWebviewContent(clusterdata: AppLensARMresponse) {
+    const webviewClusterData = clusterdata?.properties;
+    const vscodeExtensionPath = process.env.VSCODE_CWD;
+
+    if (vscodeExtensionPath === undefined) {
```
itowlson, line 80: Rather than checking `=== undefined`, check `!vscodeExtensionPath`.

src/commands/detectorDiagnostics.ts
itowlson: Pass in the `ExtensionContext`.  I'm not sure `VSCODE_CWD` is guaranteed stable - it's certainly not something I've seen before.

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+      vscode.window.showErrorMessage("No Extension path");
+      // TODO: context.extensionPath is missing at command level.
+      return "CSS file path missing";
```
itowlson, line 83: You can't return a string that might be a message or file path with no indicator as to which it is!  Use a DU such as `Errorable`, or just throw here because it's totally unexpected.

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    const stylePathOnDisk = vscode.Uri.file(path.join(vscodeExtensionPath, "src/commands/style/detector.css"));
+    const styleUri = stylePathOnDisk.with({ scheme: 'vscode-resource' });
```
itowlson, line 87: Looks like the recommended way of doing this is:

```
    const onDiskPath = vscode.Uri.file(
        path.join(context.extensionPath, 'src, 'commands', 'style', 'detector.css')
      );
      const cssSrc = panel.webview.asWebviewUri(onDiskPath);
```

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+    const webviewClusterData = clusterdata?.properties;
+    const vscodeExtensionPath = process.env.VSCODE_CWD;
+
+    if (vscodeExtensionPath === undefined) {
+      vscode.window.showErrorMessage("No Extension path");
+      // TODO: context.extensionPath is missing at command level.
+      return "CSS file path missing";
+    }
+
+    const stylePathOnDisk = vscode.Uri.file(path.join(vscodeExtensionPath, "src/commands/style/detector.css"));
```
itowlson, line 86: Not sure we should be organising CSS files under the `src` directory but probably not an issue for now.

src/commands/detectorDiagnostics.ts
itowlson: If your HTML is going to be this big and this complex, then we should certainly investigate options for pulling it out into a `.html` file, even if we then need to dynamically populate some of it via JS.

src/commands/models/detectorAppLens.ts
```diff
@@ -0,0 +1,17 @@
+export class AppLensARMresponse {
```
itowlson, line 1: You could probably make this an interface.

src/commands/models/detectorAppLens.ts
```diff
@@ -0,0 +1,17 @@
...
+    id: String;
+    name: String;
+    location: String;
+    resourceGroup: String;
+    properties: any;
+    type: string;
```
itowlson, line 7: Nitpick: inconsistent `String` vs `string`.  (I usually use the lower case version.)

src/commands/models/detectorAppLens.ts
```diff
@@ -0,0 +1,17 @@
+export class AppLensARMresponse {
+    id: String;
```
itowlson, line 2: These should all be readonly.

src/commands/detectorDiagnostics.ts
Tatsinnit: 💡 note to self: https://github.com/Azure/vscode-aks-tools/blob/master/src/tree/subscriptionTreeItem.ts#L30

src/commands/detectorDiagnostics.ts
Tatsinnit: 💡 Cool, I found a way, seems like passing `extensioncontext` is not an option, I think the register command takes `IActionContext` **but** https://code.visualstudio.com/api/references/vscode-api#extensions.getExtension allows to get the extension context. `publisher`.`name` is accessible from `extension manifest`

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,152 @@
...
+        enableScripts: true,
+        enableCommandUris: true
+      }
+    );
+
+    // Build webview from data.
+    panel.webview.html = getWebviewContent(clusterAppLensData);
+    vscode.window.showInformationMessage("AKS Diagnostics Completed.");
+  } else {
+    vscode.window.showInformationMessage('Selected Cluster has no data returned.');
```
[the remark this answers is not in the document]
Tatsinnit: 💡 I like this idea, that way along with refresh button there is no need for overly use of `showinformationmessage` window pop-up. 

new flow will be like you mentioned:

* ` if clusterdata ==> createview else ==> simple message on the webview i.e. no data for given cluster.`

src/commands/detectorDiagnostics.ts
Tatsinnit: :bulb: Cool, found a much elegant way via handlebars : https://www.npmjs.com/package/handlebars

Tatsinnit:
:bulb: This break seems unrelated, probably caused by this:  

image

src/commands/detectorDiagnostics.ts
Tatsinnit: 💡 Thanks for suggestions, shell js cli call is removed. The new way uses the `resourcemgmtclient` something like:

```
      const client = new ResourceManagementClient(target.cloudResource.root.credentials, target.cloudResource.root.subscriptionId);
      const clusterInsightInfo = await client.resources.get(...);
```

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,130 @@
...
+              if (extensionPath) {
+                loadClusterInsights(cloudTarget, extensionPath);
+              }
+        } else {
+          vscode.window.showInformationMessage('This command only applies to AKS clusters.');
+        }
+    }
+}
+
+function getExtensionPath(): string | undefined {
```
itowlson, line 34: Honestly I still feel like you are better getting the `ExtensionContext` in here - but given that this works maybe it's not worth the effort!

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,130 @@
...
+  cloudTarget: k8s.CloudExplorerV1.CloudExplorerResourceNode,
+  extensionPath: string) {
+  const clsutername = cloudTarget.cloudResource.name;
+
+  vscode.window.withProgress({
+    location: vscode.ProgressLocation.Notification,
+    title: `Loading ${clsutername} diagnostics.`,
+    cancellable: false
+  }, () => {
+    return new Promise(async (resolve) => {
```
itowlson, line 55: Instead of `() => new Promise(...)` couldn't you just do `async () => { ... }`?  Not sure what the constraints are here.  (I have a wrapper called `longRunning` which allows me to write `longRunning("Doing stuff", () => someAsyncFunction())` - I think there's a version in `porter-vscode/src/utils/host.ts` otherwise hit me up for a pointer if it would be useful.)

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,130 @@
...
+  return;
+}
+
+async function getClusterInsightInfo(
+  target: k8s.CloudExplorerV1.CloudExplorerResourceNode
+  ): Promise<Errorable<AppLensAPIResult>> {
+  try {
+      const client = new ResourceManagementClient(target.cloudResource.root.credentials, target.cloudResource.root.subscriptionId);
+      const clusterInsightInfo = await client.resources.get(
+        target.cloudResource.armId.split("/")[4], target.cloudResource.resource.type,
```
itowlson, line 102: Maybe pull this out into a named method because otherwise the `[4]` is a bit mystical.

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,130 @@
...
+        target.cloudResource.resource.name, "detectors", "mcrEndpointUpdate", "2019-08-01");
+
+      return { succeeded: true, result: {apiresult: <AppLensARMresponse> clusterInsightInfo}};
```
itowlson, line 105: Not sure you need the additional layer of result object here.  Could you just write `{ succeeded: true, result: <AppLensARMresponse>clusterInsightInfo }`?

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+function getWebviewContent(
+  clusterdata: AppLensARMresponse,
+  vscodeExtensionPath: string) {
+    const webviewClusterData = clusterdata?.properties;
+
+    const stylePathOnDisk = vscode.Uri.file(path.join(vscodeExtensionPath, 'src', 'commands', 'style', 'detector.css'));
+    const pathToHtmlOnDisk = vscode.Uri.file(path.join(vscodeExtensionPath, 'src', 'commands', 'style', 'detector.html'));
```
itowlson, line 117: Nitpick: consistency: `stylePathOnDisk`/`pathToHtmlOnDisk`

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,130 @@
...
+    const styleUri = stylePathOnDisk.with({ scheme: 'vscode-resource' });
+    const pathUri = pathToHtmlOnDisk.with({scheme: 'vscode-resource'});
+
+    const htmldata = fs.readFileSync(pathUri.fsPath, 'utf8').toString();
+    const template = htmlhandlers.compile(htmldata);
+    const data = { "cssuri": `${styleUri}`,
+                   "clustername": `${webviewClusterData.metadata.name}`,
+                   "rowdata": `${webviewClusterData.dataset[0].table.rows[1].toString().split(',')[3]}`
```
itowlson, line 125: Definitely pull this out into a named helper because I have _no idea_ what the significance of the 0s, 1s and 3s is!

Tatsinnit: +1 and this is part which will be clear soon, hopefully once `AppLens` API is fixed.

Tatsinnit: 💡 `AppLens API` update just as a note: AppLens folks are still sixing their end of API and data contract et. al. (hopefully stays same. 📓 Which means once very little from vscod eside of things are left and once we have clear API we can move forward with the end-goal. Thanks.

src/commands/detectorDiagnostics.ts
Tatsinnit: 💡  once api is fixed, this is on my list :) further detail here: https://github.com/Azure/vscode-aks-tools/pull/22#discussion_r416194960

src/commands/detectorDiagnostics.ts
Tatsinnit: 💡 **keeping this up-to-date**:  `AppLens` API is not consumable for this extension purpose hence we are figuring out what detector api to expose for this experience.

src/commands/detectorDiagnostics.ts
Tatsinnit: 💡 API discussion is over: we will be using `detectors` - `networkconnectivity` arm api.

src/commands/detectorDiagnostics.ts
Tatsinnit: 💡 API discussion is over: we will be using `detectors` - `networkconnectivity` arm api.

src/commands/helpers/networkconnectivityhtmlhelper.ts
```diff
@@ -0,0 +1,41 @@
...
+          const htmlRawDataArray = webviewClusterData.dataset[datasetIndex].table.rows;
+
+          const htmlJsonDataSet = {
+                subnet: {
+                  subnetClass: htmlRawDataArray[0][0].toString().toLowerCase(),
+                  subnetDataset: htmlRawDataArray
+                }
+              };
+
+          return JSON.stringify(htmlJsonDataSet);
```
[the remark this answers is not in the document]
Tatsinnit: 💡 I think I know what happened here, I refactored this and wrote the functionalists with signature first: i.e. `funct foo(a :type1, b :type2) -> str`  and worked on the premise of type rather then actually checking that I can just return the `htmlJsonDataSet` with `any` type. 

Sorted now, might show you tomorrow on our catch up or push it if I cover most of the key points. :)

src/commands/helpers/networkconnectivityhtmlhelper.ts
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+      static htmlHandlerRegisterHelper() {
+        htmlhandlers.registerHelper("generatehtml", this.generateHyperlinkHelper);
+      }
+
+      static generateHyperlinkHelper(htmltext: string) {
```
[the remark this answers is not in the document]
Tatsinnit: 📓 Thank you, sounds good, how about simple as: `markdownHelper`. Yes, so far all the data which API returns is html or the git style markdown hyperlinks.

src/commands/helpers/networkconnectivityhtmlhelper.ts
```diff
@@ -0,0 +1,41 @@
+import { AppLensARMresponse } from '../models/applensarmresponse';
+import * as htmlhandlers from "handlebars";
+
+export default class NetworkConnectivityHtmlHelper {
+
+    static convertHtmlJsonConfiguration(
+        webviewClusterData: AppLensARMresponse["properties"],
+        datasetIndex: number
+        ): string {
+          const htmlRawDataArray = webviewClusterData.dataset[datasetIndex].table.rows;
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Yes Indeed, an ARM Json response with HTML embedded.

A sample snippet from the ARM call with some stuff redacted:

```
"dataset": [
      {
        "renderingProperties": {
          "description": null,
          "enableEmailButtons": false,
          "isContainerNeeded": true,
          "title": "Network Configuration",
          "type": 9
        },
        "table": {
          "columns": [
            {
              "columnName": "Markdown",
              "columnType": null,
              "dataType": "String"
            }
          ],
          "rows": [
            [
              "<table>\r\n<tr>\r\n<th>Property</th><th>Value</th></tr>\r\n<tr>\r\n <td> Region </td>  <td> some_region </td>\r\n</tr>\r\n<tr>\r\n <td> FQDN </td>  <td> XXXXX.XXX.XXX.azmk8s.io </td>\r\n</tr>\r\n<tr>\r\n <td> Network Plugin </td>  <td> Kubenet </td>\r\n</tr>\r\n<tr>\r\n <td> Cluster Subnet Address Range </td>  <td> XX.XX.XX.XX/89 </td>\r\n</tr>\r\n<tr>\r\n <td> VNet Name </td>  <td> aks-vnet-XXXXX </td>\r\n</tr>\r\n<tr>\r\n <td> VNet Address Ranges </td>  <td> x.x.x.x/X </td>\r\n</tr>\r\n<tr>\r\n <td> NSG Name </td>  <td> aks-agentpool-XXXXXXXXXXXX-nsg </td>\r\n</tr>\r\n<tr>\r\n <td> Route Table Name </td>  <td> aks-agentpool-XXXXXXXX-routetable </td>\r\n</tr>\r\n<tr>\r\n <td> Allocated Outbound Ports </td>  <td> 0 (default) </td>\r\n</tr>\r\n<tr>\r\n <td> Is Private Cluster </td>  <td> XXX </td>\r\n</tr>\r\n</table>\r\n"
            ]
          ],
          "tableName": ""
        }
      }
```

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,144 @@
...
+function getWebviewContent(
+  clusterdata: AppLensARMresponse,
+  vscodeExtensionPath: string) {
+    const webviewClusterData = clusterdata?.properties;
+
+    const stylePathOnDisk = vscode.Uri.file(path.join(vscodeExtensionPath, 'src', 'commands', 'style', "networkconnectivity", 'networkConnectivity.css'));
+    const htmlPathOnDisk = vscode.Uri.file(path.join(vscodeExtensionPath, 'src', 'commands', 'style', "networkconnectivity", 'networkConnectivity.html'));
+    const styleUri = stylePathOnDisk.with({ scheme: 'vscode-resource' });
+    const pathUri = htmlPathOnDisk.with({scheme: 'vscode-resource'});
+    const portalUrl = `https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/${clusterdata.id.split('detectors')[0]}aksDiagnostics`;
```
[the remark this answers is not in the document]
Tatsinnit: 💡 found the answer to this from one of the portal feedback links: so its as simple as this --  just a sample but this will does the trick `https://portal.azure.com/#resource/subscriptions/sub_id/resourceGroups/res_grp/overview`

src/commands/detectorDiagnostics.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    const htmldata = fs.readFileSync(pathUri.fsPath, 'utf8').toString();
+
+    NetworkConnectivityHtmlHelper.htmlHandlerRegisterHelper();
+    const template = htmlhandlers.compile(htmldata);
+    const data = { cssuri: styleUri,
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Fixed it following the convention mentioned here https://handlebarsjs.com/guide/#installation

src/commands/detectorDiagnostics.ts
Tatsinnit: :bulb: Discussed and we will get on with the `getExtension` vscode function.

itowlson:
This looks good - thanks for your patience in landing it!

Issue #14, opened 2019-12-11
deprecation: package needs updating


squillace:
![image](https://user-images.githubusercontent.com/1422718/70632922-aceb5080-1bfd-11ea-8678-8cd0882e0379.png)

itowlson:
Last time I looked at this, `@azure/arm-containerservice` didn't actually work for us - I think it didn't play nicely with the credentials that the Azure Account extension gave us.  I think I raised an issue for that but I can't remember where or find it now... anyway @Tatsinnit has bravely volunteered to pick this one up so hopefully he will be able to get it sorted!

Tatsinnit:
👍  Thanks: @itowlson - I might have something to share\show\thought-share tomorrow as a fix for this, at least at this repo level. 🙈

Tatsinnit:
💡 I am re-opening this, because I can see with the changes type fix we end up in a internal error for `MS-Rest-JS` in --> `ServiceClient.ts` ==> `sendRequest` which cause extension level type error. Thanks heaps.

Tatsinnit:
Hiya guys @philliphoff & @itowlson 🛡️ 🤖 

Below is my 2 cent finding of how this update will break existing `merge config` and `save kubeconfig` changes. 

# Weird `ms-rest-js ==> serviceClient.ts` failure led to more deep dependency issue.

I am sure new package is bigger and better, **but**I really think that this whole `@azure/containerservices or arm-resource` et. al. packages change was not thought through and they contain breaking changes from the original packages, and mere: https://github.com/Azure/azure-sdk-for-js/blob/master/documentation/Migration.md is not full thought through for various other dependencies 🐉 

Here is the **weird** behaviour I see so I am planning to revert my changes and keep using old packages until we have clear plan to tackle this. Please advice.

* **Behaviour after Package update**: When I select things like `save kubeconfig` I see random error thrown which is like this: `TypeError: Cannot read property 'then' of undefined` which gets thrown from this line: https://github.com/Azure/vscode-aks-tools/blob/master/src/extension.ts#L52  

# **Lets Dig Deeper**

[...]

PR #23, opened 2020-05-28
Fix arm-containerservice deprecated pacakages warning.

Tatsinnit:
Following PR fixes the deprecated warning, screenshot below. in an attempt to fix this: https://github.com/Azure/vscode-aks-tools/issues/14 

* The `DOM` inclusion in `tsconfig` as build solution is detailed here: https://github.com/Azure/ms-rest-js/issues/367 

* I updated the `azure/arm-containerservice` to `@azure/arm-containerservice`  https://www.npmjs.com/package/@azure/arm-containerservice  more details here: https://github.com/Azure/azure-sdk-for-js/blob/master/documentation/Migration.md 

* But the key challenge I had to fix is this: (Which will explain my `<unknown>` type conversion.
 so after change, I got the following error on `azcs.ContainerServiceClient(credentials, target.root.subscriptionId);`

Which probably has to do with the type which get passed as `target` so if I do any `unknown` with `msRestJs.ServiceClientCredentials` type conversion of target.root.credentials **that** fixes the whole the issue.

```
Argument of type 'import(".../ms-rest/index").ServiceClientCredentials' is not assignable to parameter of type 'import("...@azure/ms-rest-js/es/lib/credentials/serviceClientCredentials").ServiceClientCredentials'.

```

With this **fix**, new `nom install` gives this

Screen Shot 2020-05-28 at 9 15 23 PM

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -48,8 +48,8 @@ async function getClusterKubeconfig(target: AksClusterTreeItem): Promise<string 
         vscode.window.showErrorMessage(`Invalid ARM id ${target.id}`);
         return;
     }
-    const credentials = <msRestJs.ServiceClientCredentials> <unknown>target.root.credentials;
-    const client = new azcs.ContainerServiceClient(credentials, target.root.subscriptionId);  // TODO: safely
+
+    const client = new azcs.ContainerServiceClient(restJSCredentialsFrom(target), target.root.subscriptionId);  // TODO: safely
     try {
         const accessProfile = await client.managedClusters.getAccessProfile(resourceGroupName, name, 'clusterUser');
         const kubeconfig = accessProfile.kubeConfig!.toString();  // TODO: safely
@@ -58,4 +58,8 @@ async function getClusterKubeconfig(target: AksClusterTreeItem): Promise<string 
         vscode.window.showErrorMessage(`Can't get kubeconfig: ${e}`);
         return undefined;
     }
+}
+
+function restJSCredentialsFrom(target: AksClusterTreeItem) {
+    return <msRestJs.ServiceClientCredentials><unknown>target.root.credentials;
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,6 +1,7 @@
 import * as vscode from 'vscode';
 import * as k8s from 'vscode-kubernetes-tools-api';
-import * as azcs from 'azure-arm-containerservice';  // deprecated, but @azure/arm-containerservice doesn't play nicely with AzureAccount, so...
+import * as azcs from '@azure/arm-containerservice';
+import * as msRestJs from "@azure/ms-rest-js";
 
 import { parseResource } from './azure-api-utils';
 import AksClusterTreeItem from './tree/aksClusterTreeItem';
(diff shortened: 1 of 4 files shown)


Tatsinnit:
The build error you see here is directly associated with : https://github.com/Azure/ms-rest-js/issues/367  I will try with version below `.2.0.3` for `"@azure/ms-rest-js": "2.0.2"`

tsconfig.json
```diff
@@ -4,7 +4,8 @@
         "target": "es6",
         "outDir": "out",
         "lib": [
-            "es6"
+            "es6",
+            "DOM"
```
itowlson, line 8: This seems strange and I don't get why we would need `DOM` here (I mean I get that we need it here to make things work but I don't know _why don't they work without it_ and _is this the right fix_).  Is there documentation that prescribes this?  If not, can we check in with the team that this is what they expect?

src/extension.ts
```diff
@@ -47,7 +48,8 @@ async function getClusterKubeconfig(target: AksClusterTreeItem): Promise<string
         vscode.window.showErrorMessage(`Invalid ARM id ${target.id}`);
         return;
     }
-    const client = new azcs.ContainerServiceClient(target.root.credentials, target.root.subscriptionId);  // TODO: safely
+    const credentials = <msRestJs.ServiceClientCredentials> <unknown>target.root.credentials;
```
itowlson, line 51: Very minor style suggestion: tuck away the ugly type assertions in a named helper e.g. `restJSCredentialsFrom(...)`

itowlson:
@philliphoff Would you mind taking a very quick look at this PR please?  I think you understand the credentials and Azure Account stuff better than the rest of us do!  Thanks!

tsconfig.json
itowlson: Oh never mind I realised you had already posted the link and oh dear yeah that's not good.

Tatsinnit: 😊 Yes, that is almost year and above when folks found the issue in “ms-rest-js” , so essentially deprecation help page should link the existing huge for folks who are fixing this deprecation to know what trick will actually fix their typing issues. 😎🙏

Tatsinnit:
> This looks fine to me though I would like to understand the DOM thing better. Would you be able to spare a few minutes (no longer) to look into that/explain it to me please? Thanks!

:notebook: I will add few chain of thought as continuity from what I understand by reading the ms-rest-js issue.

`dom` lib is a set of JavaScript Web API interfaces, including DOM, DOM Events,... full source code here: https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts 

For backend development, it is optional to include this. However, there is some Node.js package that is used for both frontend and backend. And you will meet type missing error when compiling without including it.

Hence this issue for us of missing types through ms-rest-js , and like one of the discussion here point: https://github.com/Azure/ms-rest-js/issues/367 

> The root cause IIUC is that the `lib` entry in tsconfig is applied for an entire program and a library's tsconfig and your dependencies' tsconfigs are ignored. So any lib entries your library needs to build will have to be present in consumers' tsconfig else they will get errors.

philliphoff:
Assuming that the credentials returned by `vscode-azure-account` match those requested by the  new ARM library, beyond the superficial typed definition requiring the ugly cast, then it's probably fine for now.

You're not the only extension dealing with this issue (see [vscode-azure-account#140](https://github.com/microsoft/vscode-azure-account/issues/140)).  If and when that library is *finally* updated, it could break this workaround.

PR #26, opened 2020-06-09
Detector feature: Safe handling clunky embedded html responce from app lens api.

Tatsinnit:
I noticed a behaviour that for a cluster once that app lens API was returning data but it was empty for a dataset `(I cannot replicate it, because it was one-off data glitch)`  for some fields and it started working on the second go, so this handle will help in guaranteeing that at least the HTML handling display.

* Added `.length` check to make sure actually the data exist for that index before, read any existing value.
* Added `handlebar` if statements to handle the `html` display` 

Adding them as pro-active checks.

Thanks,

### DIFF
--- a/src/commands/detectorDiagnostics/helpers/networkconnectivityhtmlhelper.ts
+++ b/src/commands/detectorDiagnostics/helpers/networkconnectivityhtmlhelper.ts
@@ -18,7 +18,7 @@ export function convertHtmlJsonConfiguration(
     return htmlJsonDataSet;
   }
 
-  return;
+  return undefined;
 }
 
 export function htmlHandlerRegisterHelper() {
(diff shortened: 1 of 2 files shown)


src/commands/detectorDiagnostics/helpers/networkconnectivityhtmlhelper.ts
```diff
@@ -5,17 +5,20 @@ export function convertHtmlJsonConfiguration(
...
+      subnet: {
+        subnetClass: htmlRawDataArray[0][0].toString().toLowerCase(),
+        subnetDataset: htmlRawDataArray
+      }
+    };
+    return htmlJsonDataSet;
+  }
 
-  return htmlJsonDataSet;
+  return;
```
itowlson, line 21: This will return `undefined` - prefer to be explicit if that's what you intend.

Issue #4, opened 2019-05-24
Cloud "Save Kubeconfig" doesn't work with RBAC enabled cluster


timja:
```
'Tim.Jacomb@HMCTS.NET' with object id '****' does not have authorization to perform action 'Microsoft.ContainerService/managedClusters/accessProfiles/listCredential/action' over scope '/subscriptions/***/resourceGroups/demo-00-rg/providers/Microsoft.ContainerService/managedClusters/aks-demo-cluster-00/accessProfiles/clusterUser'.
```

https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#azure-kubernetes-service-cluster-user-role

The built in role gives:
"Microsoft.ContainerService/managedClusters/listClusterUserCredential/action"

Whereas this vscode extension is trying to use:
"Microsoft.ContainerService/managedClusters/accessProfiles/listCredential/action"

itowlson:
@timja Sorry for dropping the ball on this one and for so long, and thank you for pointing me at the API I should be using.  Can you clarify how to reproduce this issue?  I created a cluster with RBAC but Save Kubeconfig seems to be working okay for me.  Does this relate to the permissions of the logged-in Azure user (i.e. AAD RBAC on the Azure object) rather than RBAC being enabled on the cluster itself (i.e. Kubernetes RBAC on the Kubernetes resources)?  Thanks!

timja:
> @timja Sorry for dropping the ball on this one and for so long, and thank you for pointing me at the API I should be using. Can you clarify how to reproduce this issue? I created a cluster with RBAC but Save Kubeconfig seems to be working okay for me. Does this relate to the permissions of the logged-in Azure user (i.e. AAD RBAC on the Azure object) rather than RBAC being enabled on the cluster itself (i.e. Kubernetes RBAC on the Kubernetes resources)? Thanks!

Yes it relates to your azure rbac role, if you only have cluster user role it doesn’t work

itowlson:
@tatsinnit can you take a look at this please?

Tatsinnit:
💡Hiya @timja ,  I tried to repro this, but it **works** for me, I wonder if this could be double check or specific steps to repro can be shared:

Here is what I did, for the successful scenario:

* Created `aks cluster` via portal default version of cluster I used was `1.15`.
* Once resource is all done: I went to `Access Control` here: 

[...]

PR #32, opened 2020-06-17
Fix for rbac save-kubeconfig by using ListClusterUserCredentials.

Tatsinnit:
This PR is to fix `rbac related save-kubeconfig` issue: https://github.com/Azure/vscode-aks-tools/issues/4

Also seems like it is recommended to use `listClusterUserCredentials` instead of `getaccessprofiler` here https://docs.microsoft.com/en-us/rest/api/aks/managedclusters/getaccessprofile 

Opening this as a draft for Ivan to take a quick look. Thanks.

cc: @itowlson

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -49,8 +49,9 @@ async function getClusterKubeconfig(target: AksClusterTreeItem): Promise<string 
     }
     const client = new azcs.ContainerServiceClient(target.root.credentials, target.root.subscriptionId);  // TODO: safely
     try {
-        const accessProfile = await client.managedClusters.getAccessProfile(resourceGroupName, name, 'clusterUser');
-        const kubeconfig = accessProfile.kubeConfig!.toString();  // TODO: safely
+        const clusterUserCredentials = await client.managedClusters.listClusterUserCredentials(resourceGroupName, name);
+        const kubeconfigList = clusterUserCredentials.kubeconfigs!.filter((kubeInfo) => kubeInfo.name === "clusterUser");  // TODO: safely
+        const kubeconfig = kubeconfigList[0].value?.toString();
         return kubeconfig;
     } catch (e) {
         vscode.window.showErrorMessage(`Can't get kubeconfig: ${e}`);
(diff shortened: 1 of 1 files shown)


itowlson:
That's an interesting issue re Get Access Profile.  The old API provided admin credentials, so we probably can't blanket move to user credentials.  (And the main use for the extension is with dev clusters where the user will typically want to be admin if they can.)  I wonder if we could display a prompt and have the user specify whether they want admin or user credentials?

itowlson:
Hmm, just looked at the old code and maybe we did get user rather than admin creds by default.  If so, yes, definitely switch over to the new API for all cases.  Might be good to compare the old credential YAML with the new credential YAML to be sure.

src/extension.ts
```diff
@@ -49,8 +49,9 @@ async function getClusterKubeconfig(target: AksClusterTreeItem): Promise<string
     }
     const client = new azcs.ContainerServiceClient(target.root.credentials, target.root.subscriptionId);  // TODO: safely
     try {
-        const accessProfile = await client.managedClusters.getAccessProfile(resourceGroupName, name, 'clusterUser');
-        const kubeconfig = accessProfile.kubeConfig!.toString();  // TODO: safely
+        const clusterUserCredentials = await client.managedClusters.listClusterUserCredentials(resourceGroupName, name);
+        const kubeconfigList = clusterUserCredentials.kubeconfigs!.filter((kubeInfo) => kubeInfo.name === "clusterUser");  // TODO: safely
```
itowlson, line 53: I don't _think_ you need the `!` here (which is what the 'TODO: safely' refers to).  The docs imply you always get the array - though obviously if the TS bindings type it as nullable then we have to work to those.

itowlson: Consider using `kubeconfigs.find(...)` instead of `kubeconfigs.filter(...)[0]` (and then check for `undefined`).

src/extension.ts
```diff
@@ -49,8 +49,9 @@ async function getClusterKubeconfig(target: AksClusterTreeItem): Promise<string
...
+        const kubeconfig = kubeconfigList[0].value?.toString();
```
itowlson, line 54: `value` is nullable and not typed as string?  Fie!

Tatsinnit: :) 
Screen Shot 2020-06-21 at 7 18 07 PM

Tatsinnit:
💡 Note: I compared the old credential YAML with the new credential YAML to be sure, and they both are identical.

itowlson:
Thanks for the double check on the YAML - that's awesome.

PR #33, opened 2020-07-15
AKS-Periscope integration in cloudExplorer for aks cluster.

Tatsinnit:
This PR enables the `AKS Periscope` tool for `vscode` - `aks extension ==> cloudExplorer`

### What this PR achieve

This PR enables users for the 2 scenarios for `aks-periscope` tool to run without any user input via `VsCode` there are 3 screenshots below for the Webviews UI as well.

**Scenario 1**: (only one storage account in aka-diagnostic setting for cluster)

* User have attached a Storage account to their cluster diagnostic setting, and now when user selected `aks-cluster` and and select `Run AKS Periscope` the extension will enable the backend to run aks-periscope on the cluster and display user with the output UI mentioned below.

**Scenario 2**: (more then one storage account in aka-diagnostic setting for cluster)

* User have attached couple of Storage account to their cluster diagnostic setting, and now when user selected `aks-cluster` and and select `Run AKS Periscope`, user will be served with `quickPick` for selection of one of those 2 or more storage account and from the selected storage account, the extension will enable the backend to run aks-periscope on the cluster and display user with the output UI mentioned below.

**Scenario 3**: (No storage account in diagnostic setting)

* User have no-attached Storage account to their cluster diagnostic setting, and now when user selected `aks-cluster` and and select `Run AKS Periscope` the extension will check and present user with helpful message and link to the new doc which is under writing for how to add the storage account to the diagnostic setting for cluster.

Key challenges we over come are:

* Integrating of aka-cluster information with diagnostic setting information and leveraging the existing functionality of this extension to bridge the 2 together.
* Once we have storage account - We go and generate SAS and run AKS-Periscope tool under the hood, formulate more user friendly output for Webviews for downloadable link lasting for 5 mins and shareable link wit restricted permissions for 7 days.

**Tool level challenges.**

Few design things which periscope needs to re-think or I will re-think and send it to the periscope guys to review: 

* Specific challenge was to carve all this from the existing design issues with aks-periscope like for example: https://github.com/Azure/aks-periscope/issues/30 
* Also the tool does-not return anything which we can utilise, also it has false positive failures like this: https://github.com/Azure/aks-periscope/issues/15 or https://github.com/Azure/aks-periscope/issues/29 or https://github.com/Azure/aks-periscope/issues/25 or 

In the end user will get served with generate downloadable link button in Webviews along with generate buttons and user can generate the downloadable and shareable link for node logs `*.zip` files.

technical challenges which we overcome so far
So far I have over come 6 key challenges with regards to this work for extension.:

* In Memory KubeConfig File management.
* Arm Monitoring for getting – User Diagnostic setting.
* Arm Storage for getting user slected or Auto selected Sas URL generation.
* In-Memory file Management for the periscope yaml file for deployment.
* Run the file from Temp storage with kubeConfig context but not effecting user level context in extension.
* Storage link generation was tricky given `Aks-periscope` tools gives output while data is loading in storage but I have found a way.
 
Screen Shot 2020-08-09 at 9 07 11 PM

image

image

Thanks


src/commands/periscope/models/storage.ts
```diff
@@ -0,0 +1,5 @@
+export interface PeriscopeStorage {
+    storageName: string;
+    storageKey: string;
+    storageDeploymentSas: string;
+  }
```
itowlson, line 5: Alignment looks wonky - do you have tslint or eslint installed?

itowlson: Or editorconfig

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,248 @@
...
+import * as path from 'path';
+import * as buffer from 'buffer';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+import { parseResource } from '../../azure-api-utils';
+import * as tmpfile from '../utils/tempfile';
+import { getClusterKubeconfig } from '../../extension';
+import { getExtensionPath, longRunning } from '../utils/host';
+import {
+  getClusterDiagnosticSettings,
```
itowlson, line 12: You're indenting by 4 elsewhere.  Is this a different indent standard or are you using tabs instead of spaces?  Use editorconfig or something to keep things consistent.

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,248 @@
...
+            2. Generate SAS for the storage account selected\provided or default supplied.
+            3. Run the clean-up and then aks-persicope on the cluster with current cluster kubeconfig.
+            4. Generate Dowloadable link once action is performed at the webView.
+        **/
+        await longRunning(`Checking storage associated cluster.`,
+          async () => {
+            // Check if Diagnostic setting exist.
+            const diagnosticSettings = await getClusterDiagnosticSettings(cloudResource);
+
+            if (diagnosticSettings?.value?.length) {
```
itowlson, line 48: I don't think this lot should be under the "Checking storage..." progress indicator.  It looks like the idea is to check the storage (long running action 1), then optionally do some UI interaction, then install Periscope (long running action 2).

Remember that `longRunning` (at least if you copied my code) returns the value of the long-running task.  So you can do something like:

```
const diagnosticSettings = await longRunning("...", () =>
    getClusterDiagnosticSettings(cloudResource)
);
```

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,248 @@
...
+              // If cluster diagnosticSetting exist pass it on for UI workflow and aks-preiscope deployment.
+              await selectStorageAccountAndInstallAKSPeriscope(diagnosticSettings, clusterKubeConfig, cloudResource);
```
itowlson, line 50: I'm a bit suspicious of these names.  It looks like what you want to do is something like:

```
const selectedAccount = (branching to select in different ways);
installAKSPeriscope(selectedAccount);
```

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,248 @@
...
+        storageAccountNameToStorageIdMap.set(name!, item.storageAccountId!);
+      }
+    });
+
+    const quickPick = vscode.window.createQuickPick();
+    quickPick.placeholder = "Select storage account for periscope deployment:";
+    quickPick.title = `Found more than 1 storage account associated with cluster ${clusterName}. Please select from below.`;
+    quickPick.ignoreFocusOut = true;
+    quickPick.items = Array.from(storageAccountNameToStorageIdMap.keys()).map((label) => ({ label }));
+    quickPick.onDidChangeSelection(async ([{ label }]) => {
```
itowlson, line 97: Not sure what this is about... I think you might have found the low-level quick pick API but the normal one is much simpler to use...

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,248 @@
...
+      quickPick.hide();
+
+      await longRunning(`Generate SAS for ${clusterName} cluster.`,
```
itowlson, line 100: I don't get why you set up the quick pick, then do this long-running thing, _then_ show the quick pick UI.

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,248 @@
...
+  clusterName: string) {
+
+  await longRunning(`Installing aks periscope for ${clusterName} cluster.`,
+    async () => {
+      const kubectl = await k8s.extension.kubectl.v1;
+
+      if (kubectl.available && periscopeStorageInfo) {
+        await inMemoryAKSPeriscopeDeplomentFileHandle(clusterKubeConfig, kubectl, periscopeStorageInfo, clusterName);
+      } else {
+        vscode.window.showInformationMessage('Associated storage to the cluster have issues.');
```
itowlson, line 159: "issues"?!

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+async function inMemoryAKSPeriscopeDeplomentFileHandle(
+  clusterKubeConfig: string,
+  kubectl: any,
+  periscopeStorageInfo: PeriscopeStorage,
+  clusterName: string
+): Promise<string | undefined> {
+  const https = require("https");
+  const aksPeriscopeFile = fs.createWriteStream(path.resolve(__dirname, 'aks-periscope.yaml'));
```
itowlson, line 172: I think you should create this in the temporary directory rather than the program directory.

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+async function createPeriscopeWebView(
+  clusterName: string,
+  outputResult: any,
+  periscopeStorageInfo: PeriscopeStorage
+) {
+  const panel = vscode.window.createWebviewPanel(
+    'AKS Periscope',
+    'AKS periscope view for : ' + clusterName,
```
itowlson, line 216: Keep titles short e.g. `AKS Periscope: ${clusterName}`

src/commands/utils/host.ts
```diff
@@ -3,20 +3,63 @@ import * as vscode from 'vscode';
...
-    }
-    return vscodeExtensionPath;
+  if (!vscodeExtensionPath) {
+    vscode.window.showInformationMessage('No Extension path found.');
+    return;
+  }
+  return vscodeExtensionPath;
 }
+
+export function getSASKey(
```
itowlson, line 24: This doesn't belong in `host.ts` which is for functions relating to the hosting environment (VS Code).  Consider an `azure-storage` module instead.

src/commands/utils/host.ts
```diff
@@ -3,20 +3,63 @@ import * as vscode from 'vscode';
...
+  storageAccount: string,
+  storageKey: string,
+  generateSevenDaySas = false
+) {
+  const storage = require("azure-storage");
+  const startDate = new Date();
+  const expiryDate = new Date();
+
+  startDate.setTime(startDate.getTime() - 5 * 60);
```
itowlson, line 33: What is this doing?  Setting the time back 5 minutes?  Give it an intention-revealing name.

itowlson: Hmm, isn't this in milliseconds?

src/commands/utils/host.ts
```diff
@@ -3,20 +3,63 @@ import * as vscode from 'vscode';
...
+  let defaultExpirationTime = expiryDate.getTime() + 2 * 24 * 60 * 60 * 1000;
```
itowlson, line 34: Because you overwrite this in the 7-day case, you are using the variable as the _actual_ expiration time not just a default; so just call it `expirationTime`.

src/commands/utils/host.ts
```diff
@@ -3,20 +3,63 @@ import * as vscode from 'vscode';
...
+    vscode.window.showInformationMessage('No Extension path found.');
+    return;
+  }
+  return vscodeExtensionPath;
 }
+
+export function getSASKey(
+  storageAccount: string,
+  storageKey: string,
+  generateSevenDaySas = false
```
itowlson, line 27: Consider making this something like `expiryDays` - then the caller is in control of what options to offer and this low-level function is not bound to the Periscope UI.

src/commands/utils/host.ts
```diff
@@ -3,20 +3,63 @@ import * as vscode from 'vscode';
...
+  const sharedAccessPolicy = {
+    AccessPolicy: {
+      Services: AccountSasConstants.Services.BLOB,
+      ResourceTypes: AccountSasConstants.Resources.SERVICE +
+        AccountSasConstants.Resources.CONTAINER +
+        AccountSasConstants.Resources.OBJECT,
+      Permissions: AccountSasConstants.Permissions.READ +
+        AccountSasConstants.Permissions.ADD +
+        AccountSasConstants.Permissions.CREATE +
+        AccountSasConstants.Permissions.WRITE +
```
itowlson, line 51: Does the shareable link need write permissions?  Something to discuss with the PMs who will know the use cases.

src/extension.ts
```diff
@@ -41,7 +45,7 @@ export async function activate(context: vscode.ExtensionContext) {
     }
 }
 
-async function getClusterKubeconfig(target: AksClusterTreeItem): Promise<string | undefined> {
+export async function getClusterKubeconfig(target: AksClusterTreeItem): Promise<string | undefined> {
```
itowlson, line 48: You don't actually need to do this I think - I think it is already exposed on the tree cluster object as `getKubeconfigYaml` or something like that.

Tatsinnit:
> I'm afraid I struggled to follow a lot of the logic in some of the more complicated functions. It would be very useful to break bits out into well-named functions, and maybe split some of the utility methods out into more general-purpose files (e.g. a storage utilities file). It would also be really useful if you could sit down with Andy and/or the PMs and pin down a clearer user interface because I'm afraid I got lost in all the "output is this, link is that, generate some other link..." options. I know this is hard to express in a PR because HTML is so verbose and fiddly but again I was struggling with it I'm afraid...


+1 for UI part and I am with you, sounds good, let me block some meeting there is no mock UI I have received so far so the UI part is all fuzzy. Thank you for eye balling this 🙏😊

Via this code part of things we achieved is something which the tool as periscope should be doing vs what we handle via vscode but that is for the time when periscope tool has those Northstar enhancements moving forward. 

Essentially key thing I have achieved is the auto download link generation which I wanted your eyes on and initial comments will be good to make this much better.

package.json
```diff
@@ -39,6 +40,10 @@
             {
                 "command": "aks.detectorDiagnostics",
                 "title": "Run AKS Diagnostics"
+            },
+            {
+                "command": "aks.periscope",
+                "title": "Install AKS Periscope"
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Agree, I have replaced this with `Run` because actually this is `tool which runs` at the background.

resources/webviews/periscope/periscope.html
```diff
@@ -0,0 +1,110 @@
...
+            vscode.postMessage({ command: "7days" })
+        }
+    </script>
+</head>
+
+<body>
+
+    <div class="heading">
+        <h2 class="content-header">Periscope ran for cluster: {{name}}</h2>
+        <br\> Know more about periscope <a href="https://aka.ms/periscope">https://aka.ms/periscope</a>
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Note, thanks for this, fyi --> Whole new `UI` is coming where I spent more than a week with PM for nicer outcome. I will push those changes soon, and update this PR.

src/extension.ts
Tatsinnit: 💡 Cool,  (Setting this note for myself as a context) I have tried but something seems disconnect, so I might show it to you once I have pushed my local changes. It is to do with `getKubeConfig` available at run time at `cloudExplorer.api.explorer.provider[0].getKubeConfig` but I am not sure how can I translate this to what we have.

resources/webviews/periscope/periscope.html
```diff
@@ -0,0 +1,110 @@
...
+        {{#unless outputCode}}
+            <div class="panel-body data-container-body">
+                Your logs are being uplaoded to the storage account: {{storageAccName}}.
+            </div>
+
+            <div class="panel-body data-container-body">
+                <h5 class="panel-title">AKS periscope the logs are uploaded successfully, <br />
+                    (Please note: Due to the nature how Periscope tool works it could be a possibility that log files of
+                    some nodes are still writing to the storage hence
+                    it is recommended to re-click generate button or wait for a minutebefore generating the downloable link.
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Yeah, so essentially this is `by design` for AKS for now, but this will change: https://github.com/Azure/aks-periscope/issues/30  I have bunch of other enhancement mentioned against this repository and planning to latter possibly float a design doc for the `aks-periscope` what happens is that `aks-periscope` tool runs, and whole storage is still loading the logs, it goes greedy and reports, I am done, (essentially not actually done, runs the process at the background, i.e. no `go- channel` under the hood.

src/commands/periscope/helpers/storageandhtmlhelper.ts
```diff
@@ -0,0 +1,170 @@
...
+    } catch (e) {
+        vscode.window.showErrorMessage(`Storage associated with cluster had following error: ${e}`);
+        return undefined;
+    }
+}
+
+export async function generateDownloadableLinks(
+    periscopeStorage: PeriscopeStorage,
+    output: string,
+    generateSevenDaySas = false) {
```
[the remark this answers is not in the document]
Tatsinnit: 💡 This will be changes to `expiryDate` soon, following one of the idea you mentioned, that way this functionalists is more usable by anyone passing the expiry date and getting the `sas` back.

resources/webviews/periscope/periscope.html
```diff
@@ -0,0 +1,110 @@
...
+                {{#each downloadableLinks}}
+                <ul>
+                    <li><a href="{{this}}">Node log {{@index}}.</a></li>
+                </ul>
+                {{/each}}
+                </div>
+                {{/if}}
+
+                <div class="panel-body data-container-body">
+                    <h5 class="panel-title">Generate 7-day expiration link for cluster node logs.</h5>
```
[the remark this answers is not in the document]
Tatsinnit: 👍  thanks for this, pretty. much whole html stuff is re-new-ed now and all this will be gone. Will update this PR EOD with what we have now.

src/commands/periscope/periscope.ts
Tatsinnit: 💡 I am easy with any simpler mechanism to keep user updated, main intent is to inform user that -- since you have selected your account in case of more than 1 account, code will go and generate the SAS. along with few other long running tasks.

Tatsinnit:
> > I'm afraid I struggled to follow a lot of the logic in some of the more complicated functions. It would be very useful to break bits out into well-named functions, and maybe split some of the utility methods out into more general-purpose files (e.g. a storage utilities file). It would also be really useful if you could sit down with Andy and/or the PMs and pin down a clearer user interface because I'm afraid I got lost in all the "output is this, link is that, generate some other link..." options. I know this is hard to express in a PR because HTML is so verbose and fiddly but again I was struggling with it I'm afraid...
> 
> +1 for UI part and I am with you, sounds good, let me block some meeting with PMs, there is no mock UI I have received so far so the UI part is all fuzzy + I feel like dropping the manual UI steps for a scenario where there is no storage attached to diagnosticSettings (PMs can guide me more on that). Thank you for eye balling this 🙏😊
> 
> Via this code part of things we achieved is something which the tool as periscope should be doing vs what we handle via vscode but that is for the time when periscope tool has those Northstar enhancements moving forward.
> 
> Essentially key thing I have achieved is the auto download link generation which I wanted your eyes on and initial comments will be good to make this much better.

Thank you for starting this discussion, as I was really looking forward it to be part of this Draft PR.

* Fyi: UI is redone, detail of 3 goal scenario for this work is explain on the top description part as well and **PM team** was kind enough to do the screen flow which I reciprocated in the Webviews for `VsCode` along with functionality.🙏😊

src/commands/periscope/periscope.ts
Tatsinnit: 💡 Yeah will be good to know if there is any simpler alternative, but I think ask is to have some kind of `table, title, placeholder, along with focusOut disabled` I am currently using this: https://code.visualstudio.com/api/references/vscode-api#window.createQuickPick Which helped in the `onDidChangeSelection` to take action for the scenario where there are more than 1 storage account. :)

src/extension.ts
Tatsinnit: 💡 Thank you for the discussion, it's implemented as external help function now. 🙏😊

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,258 @@
...
+    } catch (e) {
+        vscode.window.showErrorMessage(`Storage associated with cluster had following error: ${e}`);
+        return undefined;
+    }
+}
+
+export function writeTempAKSDeploymentFile(file: any) {
+    const https = require("https");
+
+    https.get('https://raw.githubusercontent.com/Azure/aks-periscope/master/deployment/aks-periscope.yaml', (res: any) => {
```
itowlson, line 114: If you're doing a callback rather than an `await` then how does your caller know the write has finished?

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,258 @@
...
+    // persicope file is written now, replace the acc name and keys.
+    const replace = require("replace");
+
+    const base64Name = Buffer.from(periscopeStorageInfo.storageName).toString('base64');
+    const base64Sas = Buffer.from(periscopeStorageInfo.storageDeploymentSas).toString('base64');
+    replace({
+        regex: "# <accountName, base64 encoded>",
+        replacement: `"${base64Name}"`,
+        paths: [aksPeriscopeFilePath],
+        recursive: true,
```
itowlson, line 135: Are you sure you want these to be recursive?

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,258 @@
...
+                listCurrentUploadedFolders.push(item.name);
+            }
+        }
+
+        // Sort and get the latest uploaded folder under the container used by periscope.
+        const periscopeHtmlinterfaceList = [];
+        for await (const blob of containerClient.listBlobsFlat()) {
+            // Get the latest uploaded folder, then Identify the Zip files in the latest uploaded logs within that folder
+            // and extract *.zip files which are individual node logs.
+            const periscopeHTMLInterface = <PeriscopeHTMLInterface>{};
```
itowlson, line 189: Again, avoid this idiom; you're misleading the type system.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,258 @@
...
+            if (blob.name.indexOf(listCurrentUploadedFolders.sort().reverse()[0]) !== -1 && blob.name.indexOf('.zip') !== -1) {
```
itowlson, line 190: It's not obvious what the intent is here: could we pull this out into something with an intention-revealing name?  (Whether a function or a local variable.)

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,258 @@
...
+        // List all current blob.
+        const listCurrentUploadedFolders = [];
+        for await (const item of containerClient.listBlobsByHierarchy("/")) {
+            if (item.kind === "prefix") {
+                listCurrentUploadedFolders.push(item.name);
+            }
+        }
+
+        // Sort and get the latest uploaded folder under the container used by periscope.
+        const periscopeHtmlinterfaceList = [];
```
itowlson, line 185:
```suggestion
        const periscopeHtmlInterfaceList = [];
```

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,258 @@
...
+        const diagnosticSettings = await diagnosticMonitor.diagnosticSettingsOperations.list(cloudResource.id);
+
+        return diagnosticSettings;
+    } catch (e) {
+        vscode.window.showErrorMessage(`Error fetching cluster diagnostic monitor: ${e}`);
+        return undefined;
+    }
+}
+
+export async function selectStorageAccountAndInstallAKSPeriscope<T>(
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Thanks for noticing this, I think this is rickashay from the last refactor to align all the function properly, I will `better name + remove` some stuff necessary for this funct.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,258 @@
...
+        clusterStorageInfo.sevenDaysSasyKey = getSASKey(accountName, clusterStorageInfo.storageKey!, SASExpiryTime.SevenDays);
+
+        return clusterStorageInfo;
+    } catch (e) {
+        vscode.window.showErrorMessage(`Storage associated with cluster had following error: ${e}`);
+        return undefined;
+    }
+}
+
+export function writeTempAKSDeploymentFile(file: any) {
```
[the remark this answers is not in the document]
Tatsinnit: 💡 This is actually is associated with another awesome idea (from what:  I think there is one associated comment above), so essentially, I will make this function much simpler and all the `finish` capture stuff moved here, so that once the chunks are done , then the `replacement` functionalists will be called.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,258 @@
...
+    BlobServiceClient,
+    StorageSharedKeyCredential
+} = require("@azure/storage-blob");
+
+export async function getClusterDiagnosticSettings(
+    target: k8s.CloudExplorerV1.CloudExplorerResourceNode
+): Promise<amon.MonitorManagementModels.DiagnosticSettingsCategoryResourceCollection | undefined> {
+    try {
+        // Get daignostic setting via diagnostic monitor
+        const cloudResource = target.cloudResource;
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Yeah I think, one key purpose is that if cluster type if `cloudResource` then it has type of `any` rather then the its parent which is well typed. Please let me know what do you think? (thoughts, idea et. al.)

image

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,258 @@
...
+
+        return clusterStorageInfo;
+    } catch (e) {
+        vscode.window.showErrorMessage(`Storage associated with cluster had following error: ${e}`);
+        return undefined;
+    }
+}
+
+export function writeTempAKSDeploymentFile(file: any) {
+    const https = require("https");
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Keeping this as note to self, for the case when I might quickly discuss couple of stuff: 

so what I noticed is that : use of `import` cause this to `get` funct unrecognisable: I think its explained here, possibly the cause: https://stackoverflow.com/questions/35706164/typescript-import-as-vs-import-require

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,193 @@
+import * as vscode from 'vscode';
+import * as fs from 'fs';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+import { getClusterKubeconfig } from '../../extension';
+import * as tmpfile from '../utils/tempfile';
+import { getExtensionPath, longRunning } from '../utils/host';
+import {
+  getClusterDiagnosticSettings,
```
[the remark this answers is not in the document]
Tatsinnit: 💡 fyi: I have use `editorconfig` for formatting of these files.

itowlson:
The k8s extension API types it as `any` because _it’s the resource that your extension defined_.  The k8s extension doesn’t know that type.  But you do.  So when you retrieve your cloud resource object from the ignorant k8s API, you can and should use a cast/assertion to ‘remember’ what it is.

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,177 @@
...
+async function getDiagnosticSettingsStorageAccount(
+  cloudTarget: k8s.CloudExplorerV1.CloudExplorerResourceNode,
+): Promise<string | undefined> {
+  const clusterDiagnosticSettings = await getClusterDiagnosticSettings(cloudTarget);
+
+  if (clusterDiagnosticSettings?.value?.length) {
+    const storageAccountId = await selectStorageAccountAndInstallAKSPeriscope(clusterDiagnosticSettings);
+    return storageAccountId;
+  } else {
+    // If there is no storage account attached to diagnostic setting, don't move forward and at this point we will render webview with helpful content.
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Thank you, yeah I think with the new flow, I have moved all the `createwebview` to the top level function.

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,177 @@
...
+    // For the case of successful run of the tool we render webview with the output information.
+    panel.webview.html = getWebviewContent(clusterName, extensionPath, outputResult, periscopeStorageInfo, []);
+
+    let downloadableAndShareableNodeLogsList: PeriscopeHTMLInterface[] | undefined;
+
+    panel.webview.onDidReceiveMessage(
+      async (message) => {
+        await longRunning(`Generating storage downloadable link.`,
+          async () => {
+            if (message.command === "generateDownloadLink") {
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Cool,  I might get some thoughts shred's with you  tomorrow or early next week (if you have time) week once you have had chance to look at the current state incorporating all the changes, and see if this is in need now.

I kind of agree as well, but other chain of that is that since it is just used at this one place.I guess on the other hand it will make more sense if we can put some extension level constant.

Tatsinnit:
> The k8s extension API types it as `any` because _it’s the resource that your extension defined_. The k8s extension doesn’t know that type. But you do. So when you retrieve your cloud resource object from the ignorant k8s API, you can and should use a cast/assertion to ‘remember’ what it is. From: Tatsinnit <notifications@github.com> Sent: Tuesday, 11 August 2020 22:42 To: Azure/vscode-aks-tools <vscode-aks-tools@noreply.github.com> Cc: Ivan Towlson <Ivan.Towlson@microsoft.com>; Review requested <review_requested@noreply.github.com> Subject: Re: [Azure/vscode-aks-tools] AKS-Periscope integration in cloudExplorer for aks cluster. (#33) @Tatsinnit commented on this pull request.
> ______________

💡 Thank you for this note and just a loop back for this PR, we eventually found the right type for cluster resource in this instance as `AksClusterTreeItem` 🙏 :)

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,265 @@
...
+        // Create quick pick for more than 1 storage account scenario.
+        const selectedQuickPick = await vscode.window.showQuickPick(
+            accountQuickPicks,
+            {
+                placeHolder: "Select storage account for Periscope deployment:",
+                ignoreFocusOut: true,
+                onDidSelectItem: (item: string) => { return item.toString(); }
+            });
+
+        if (selectedQuickPick?.label) {
```
itowlson, line 69: Can't this just be `if (selectedQuickPick)`?  There shouldn't be any picks without labels.

itowlson: Then you can drop the `?` on the next line because the compiler will know that `selectedQuickPick` is non-null.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,265 @@
...
+
+        return diagnosticSettings;
+    } catch (e) {
+        vscode.window.showErrorMessage(`Error fetching cluster diagnostic monitor: ${e}`);
+        return undefined;
+    }
+}
+
+export async function chooseStorageAccount(
+    diagnosticSettings: amon.MonitorManagementModels.DiagnosticSettingsResourceCollection | undefined,
```
itowlson, line 35: Do you actually need to allow `undefined` as input here?  Consider making the argument 

```suggestion
    diagnosticSettings: amon.MonitorManagementModels.DiagnosticSettingsResourceCollection,
```

and bailing early at the call site.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,265 @@
...
+): Promise<string | undefined> {
+    /*
+        Check the diagnostic setting is 1 or more than 1:
+          1. For the scenario of 1 storage account in diagnostic settings - Pick the storageId resource and get SAS.
+          2. For the scenario for more than 1 then show VsCode quickPick to select and get SAS of selected.
+    */
+
+    if (diagnosticSettings && diagnosticSettings.value!.length > 1) {
```
itowlson, line 43: Suggest using the early-exit strategy e.g.

```
if (!diagnosticSettings || !diagnosticSettings.value) {
  return undefined;
}

if (diagnosticSettings.value.length === 1) {
  // return the only one
}

// do the picker and return the result of that
```

This avoids the extra indentation and makes your `if` conditions less complex.  You dispose of the 'easy' cases early and then focus on the hard case (which, by the way, you could extract to a helper function to keep the logic here _even more_ focused on the which-case-are-we-in thing).

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,265 @@
...
+        // accounts is now an array of {id, name}
+        const accountQuickPicks = storageAccountNameToStorageIdArray;
+
+        // Create quick pick for more than 1 storage account scenario.
+        const selectedQuickPick = await vscode.window.showQuickPick(
+            accountQuickPicks,
+            {
+                placeHolder: "Select storage account for Periscope deployment:",
+                ignoreFocusOut: true,
+                onDidSelectItem: (item: string) => { return item.toString(); }
```
itowlson, line 66: I still don't understand what this is doing.  I've _never_ had to use `onDidSelectItem` in a quick pick.  Normally you just provide the array of entries and the placeholder/title, and then get the result from `await showQuickPick`.  That seems like it should suffice fine here; why do you need this in this case?

Also, if you _do_ need it... it appears to be calling `toString` on a string.  Is that really adding any value?

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    return undefined;
+}
+
+export async function getStorageInfo(
+    cluster: AksClusterTreeItem,
+    diagnosticStorageAccountId: string
+): Promise<PeriscopeStorage | undefined> {
+    try {
+        const { resourceGroupName, name: accountName } = parseResource(diagnosticStorageAccountId!);
```
itowlson, line 87: You shouldn't need the dammit operator - `diagnosticStorageAccountId` is already non-nullable.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,265 @@
...
+
+        if (!resourceGroupName || !accountName) {
+            vscode.window.showErrorMessage(`Invalid storage id ${diagnosticStorageAccountId} associated with the cluster`);
+            return;
```
itowlson, line 91: Nit:
```suggestion
            return undefined;
```

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -0,0 +1,265 @@
...
+        vscode.window.showErrorMessage(`Storage associated with cluster had following error: ${e}`);
+        return undefined;
+    }
+}
+
+export async function prepareAKSPeriscopeDeploymetFile(
+    clusterStorageInfo: PeriscopeStorage
+): Promise<string | undefined> {
+    const tempFile = tmp.fileSync({ prefix: "aks-periscope-", postfix: `.yaml` });
+    const axios = require("axios");
```
itowlson, line 117: Again, is there a reason why you don't `import * as axios from 'axios';` this at the top of the file?

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,163 @@
...
+        } else {
+            vscode.window.showInformationMessage('This command only applies to AKS clusters.');
+        }
+    }
+}
+
+async function runAKSPeriscope(
+    cluster: AksClusterTreeItem,
+    clusterKubeConfig: string
+): Promise<void | undefined> {
```
itowlson, line 48: I don't think you need `void | undefined` here - your consumer is not expecting a result.  Just go for `Promise<void>` and use plain `return;` statements (without return values).

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    if (!clusterDiagnosticSettings || !clusterDiagnosticSettings.value?.length) {
+        // If there is no storage account attached to diagnostic setting, don't move forward and at this point we will render webview with helpful content.
+        await createPeriscopeWebView(cluster.name, undefined, undefined, false);
+        return undefined;
+
+    }
+
+    const clusterStorageAccountId = await longRunning(`Identifying cluster associated storage account.`,
+        () => chooseStorageAccount(clusterDiagnosticSettings)
```
itowlson, line 64: This can display interactive UI so shouldn't have a `longRunning` around it.  I think that function is quick on the non-interactive paths so you can probably just remove the `longRunning` - if there is nothing to select, then you don't need the progress popup, and if there is, then the quick pick list shows the user that you are waiting for _them_ (rather than for a background process, which is what the progress popup is for).

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,163 @@
...
+    // Generate storage sas keys, manage aks persicope run.
+    if (!clusterStorageAccountId) return undefined;
+
+    const clusterStorageInfo = await longRunning(`Generating SAS for ${clusterName} cluster.`,
+        () => getStorageInfo(cluster, clusterStorageAccountId)
+    );
+
+    if (!clusterStorageInfo) return undefined;
+
+    const aksDeplymentFile = await longRunning(`Deploying AKS Periscope to ${clusterName}.`,
```
itowlson, line 76: Nit:

```suggestion
    const aksDeploymentFile = await longRunning(`Deploying AKS Periscope to ${clusterName}.`,
```

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,163 @@
...
+        () => prepareAKSPeriscopeDeploymetFile(clusterStorageInfo)
+    );
+
+    if (!aksDeplymentFile) return undefined;
+
+    const runCommandResult = await longRunning(`Running AKS Periscope on ${clusterName}.`,
+        () => runAssociatedAKSPeriscopeCommand(aksDeplymentFile, clusterKubeConfig)
+    );
+
+    await createPeriscopeWebView(clusterName, runCommandResult, clusterStorageInfo);
```
itowlson, line 86: I like this new flow a lot.  Very clear.  Good stuff!

src/commands/periscope/periscope.ts
```diff
@@ -0,0 +1,163 @@
...
+            (f) => kubectl.api.invokeCommand(`delete ns aks-periscope --kubeconfig="${f}"`));
+
+        // Deploy the aks-periscope.
+        const runCommandResult = await tmpfile.withOptionalTempFile<k8s.KubectlV1.ShellResult | undefined>(
+            clusterKubeConfig, "YAML",
+            (f) => kubectl.api.invokeCommand(`apply -f ${aksPeriscopeFile} --kubeconfig="${f}" && kubectl cluster-info --kubeconfig="${f}"`));
+
+        return runCommandResult;
+    } else {
+        vscode.window.showWarningMessage(`Kubectl is unavailable.`);
```
itowlson, line 109: Nit: again, this is the sort of case where I would consider early exit - dispose of the unlikely error case, and then proceed with the main logic without having an unfinished `if` dangling over the reader.

```
const kubectl = ...;

if (!kubectl.available) {
  // message and return
}

// now do all the stuff that uses kubectl unconditionally and without the extra indent!
```

itowlson (approved):
Some suggestions, some maybe worth cleaning up, others probably not worth it at this point.  But looks good and happy to merge in this state if you can't muster the energy to keep banging your head against it!

Issue #34, opened 2020-08-18
WorkItem: Generic wrapper for telemetry for enabling all feature for telemetry.


Tatsinnit:
Intent of this placeholder work item is to hook wrapper functionality to enable all the existing feature of this extension to enable telemetry.

* Possibly `regiterCommand` wrapper which add telemetry based on command used.
* Some kind of update document for the telemetry enablement for wider audience.

Tatsinnit:
New telemetry wrapper method for `registerCommand` is `registerCommandWithTelemetry` this calls a `Telemeterise` which wraps the `callback and with sendTelemetry` mechanism with editing register command. Thanks.

PR #36, opened 2020-08-25
Telemetry wrapper for register command.

Tatsinnit:
Draft PR to get some thoughts on generalising the `telemetry` with the `registerCommand`. 

This work creates a helper for `registerCommand` + `EnableTelemetry` ==> `registerCommandWithTelemetry` and everytime user will run any feature under this extension telemetry data will also be part of it.

I have tested the following the changes locally and they appear the `VSCodeExt` db for telemetry. 

* Associated work item: https://github.com/Azure/vscode-aks-tools/issues/34 

Thanks heaps, cc: @itowlson 🙏😊

### DIFF
--- a/src/commands/utils/host.ts
+++ b/src/commands/utils/host.ts
@@ -1,4 +1,6 @@
 import * as vscode from 'vscode';
+import { registerCommand } from 'vscode-azureextensionui';
+import { reporter } from '../utils/reporter';
 
 const meta = require('../../../package.json');
 
@@ -19,4 +21,13 @@ export function getExtensionPath(): string | undefined {
         return;
     }
     return vscodeExtensionPath;
+}
+
+export function registerCommandWithTelemetry(
+    commandId: string,
+    callback: any
+) {
+
+    reporter.sendTelemetryEvent("command", { command: commandId });
+    registerCommand(commandId, callback);
 }
--- a/src/commands/utils/host.ts
+++ b/src/commands/utils/host.ts
@@ -1,6 +1,4 @@
 import * as vscode from 'vscode';
-import { registerCommand } from 'vscode-azureextensionui';
-import { reporter } from '../utils/reporter';
 
 const meta = require('../../../package.json');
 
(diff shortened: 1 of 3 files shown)


src/commands/utils/host.ts
```diff
@@ -20,3 +22,12 @@ export function getExtensionPath(): string | undefined {
     }
     return vscodeExtensionPath;
 }
+
+export function registerCommandWithTelemetry(
+    commandId: string,
+    callback: any
+) {
+
+    reporter.sendTelemetryEvent("command", { command: commandId });
```
itowlson, line 31: No, this is wrong.  This will send an event for every command when the user activates the extension, and then do nothing when the user _invokes_ the command.

What you need is something more like:

```
function registerCommand(command: string, callback: (...args: any[]) => any): vscode.Disposable {
    const wrappedCallback = telemetrise(command, callback);
    return vscode.commands.registerCommand(command, wrappedCallback);
}

function telemetrise(command: string, callback: (...args: any[]) => any): (...args: any[]) => any {
    return (a) => {
        if (reporter) {
            reporter.sendTelemetryEvent("command", { command: command });
        }
        return callback(a);
    };
}
```

This is a simplified version of the `vscode-kubernetes-tools` wrapping code - you may need to make some changes to make it work with the `vscode-azureextensionui` stuff.  But the concept remains the same:

* The `telemetrise` function takes a command handler, and produces a new handler that does _exactly the same thing_ except it shoots off some telemetry first.
* The `registerCommand` function does NOT send telemetry events itself.  Rather, it asks `telemetrise` to give it an instrumented version of the handler.  Then it registers that instrumented handler instead of the vanilla handler.

Consequently, when the user issues a command registered in this way, VS Code hands control to the _instrumented_ handler (because that's the one we registered), and that handler queues a telemetry event, and then hands off to the "original" handler.

Note also that we do some typing to ensure that callers of `registerCommad` only pass legitimate handlers, and that we have to take care to forward args from the instrumented handler to the underlying original handler.

itowlson (changes requested):

src/commands/utils/host.ts
Tatsinnit: Ah, HOFs :)

itowlson (approved):
Looks good to me - thanks!

Issue #40, opened 2020-09-17
Refresh icon in Cloud Explorer.


Tatsinnit:
**Problem**:

* When I add new cluster under existing sub I donot have any way to refresh my sub for new cluster to appear I.e. Refresh icon in cloud explorer will resolve this issue.

**Steps to produce**:

* I added new cluster and had my explorer window open in extension.
* Once new AKS cluster is added, I tried to refresh the current sub in my extension.
* Result: Sub is not refreshed with new cluster.

**Screenshot below**: 

![image](https://user-images.githubusercontent.com/6233295/93402731-a51db280-f8d9-11ea-9ad1-3b1f3a37884b.png)

Thanks,

Tatsinnit:
💡 I fee like this is not really a `bug` but more kind of feature is in need, the `refresh` here is only responsible for the subs addition refresh not the treeItem refresh, jsut like `select subscription` we can add a `refresh subscription` and that way we can achieve this.

Tatsinnit:
☕️💡upon further rethink, this is not really a big of this extension essentially as explained in the associated closed PR, this api https://github.com/Azure/vscode-kubernetes-tools-api/blob/master/ts/cloudexplorer/v1.ts takes care of resolving the target.

Hence rather then hack I might open a tracking issue against the api respository. Please feel free to add more context and we can reopen it if there is any further info.

thanks,

PR #41, opened 2020-09-18
Add refresh subscription functionality.

Tatsinnit:
Following Draft PR adds the refresh `subscription` functionality for the `cloud explorer --> subscriptions` so that it take care of the missing gap detailed here: https://github.com/Azure/vscode-aks-tools/issues/40 cc: @viper4u - Please let me know if you have any thoughts?

Currently there is no way to refresh existing subs hence any new cluster added will not be shown under the existing subs in the cloud explorer.

Opening draft PR for wider thoughts \ ideas.

Thanks heaps.

The new way will look like this: 

image


package.json
```diff
@@ -37,6 +37,14 @@
                     "light": "resources/light/filter.svg"
                 }
             },
+            {
+                "command": "aks.refreshSubscription",
+                "title": "Refresh subscription...",
```
itowlson, line 42: Title case, and no ellipsis unless it needs additional data to complete (which seems unlikely).

src/commands/refreshSubscriptions.ts
```diff
@@ -0,0 +1,14 @@
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from "vscode-azureextensionui";
+
+export default async function refreshSubscriptions(context: IActionContext, target: any): Promise<void> {
```
itowlson, line 4: This says 'subscriptions' plural - I thought it was refreshing the cluster list for the selected subscription?

src/commands/refreshSubscriptions.ts
```diff
@@ -0,0 +1,14 @@
...
+    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+
+    if (cloudExplorer.available) {
+        const commandTarget = cloudExplorer.api.resolveCommandTarget(target);
+
+        if (commandTarget && commandTarget.nodeType === 'resource') {
+            target.provider.treeDataProvider.refresh();
```
itowlson, line 11: This will refresh the entire Azure section of the tree including any other subscriptions won't it?

itowlson, line 11: I'm afraid I'm deeply uncomfortable with the way you get the tree data provider here.  `target` is an opaque value provided by the k8s extension.  The **ONLY** thing we promise you can do with it is call `resolveCommandTarget` on it.  **ANYTHING** else might change at the drop of a hat, in a minor version, without warning. If you want your code to keep working, you can't poke around in the `target` object with the debugger and take a dependency on things that happen to be there on the day you did the poking - you have to stick to the contract.  And the contract (https://github.com/Azure/vscode-kubernetes-tools-api/blob/master/ts/cloudexplorer/v1.ts) says that resolving the target will give you back the resource you originally provided and _nothing else_ (well except the cloud name).

(Really we should treat `target` as type `unknown` rather than type `any` - I wonder if this works with passing it into `resolveCommandTarget`.)

Anyway the upshot is that you need to get the reference the tree provider some other way, either by stashing a copy in every tree node (because the contract promises to hand your tree node back to you in the `cloudResource` member), or by storing it a global, or closing over it, or something.  Sorry.

itowlson (changes requested):

PR #42, opened 2020-09-21
Adding cicd configure/browse pipeline options

kanika1894:
1. Adding the menu option for setting up ci/cd and browse pipeline
image

On selecting configure ci/cd option, the configure-cicd-pipeline exposed by 'Deploy-to-azure' extension gets executed. This enables the user to configure a ci-cd pipeline from VSCode. The implementation of this command can be found here : https://github.com/microsoft/vscode-deploy-azure. 
The parallel work in deploy to Azure extension, to support this functioning can be tracked here : https://github.com/microsoft/vscode-deploy-azure/pull/165

2. If the extension isn't installed, a warning message will appear and will open the window to install 'Deploy to Azure' extension
image

The second functionality is of 'Browse all pipelines'. Selecting this option takes the user to the deployment center page of the selected cluster where he/she can have a look at the configured pipelines of the cluster.


src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+async function getCluster(target: any): Promise<AksClusterTreeItem | undefined> {
+    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+    if (cloudExplorer.available) {
+        const clusterTarget = cloudExplorer.api.resolveCommandTarget(target);
+        if (clusterTarget && clusterTarget.cloudName === "Azure" && clusterTarget.nodeType === "resource" && clusterTarget.cloudResource.nodeType === "cluster") {
+            return clusterTarget.cloudResource as AksClusterTreeItem;
+        } else {
+            vscode.window.showInformationMessage('This command only applies to AKS clusters.');
```
bishal-pdMSFT, line 36: will we ever land here as in the manifest, you are specifying a condition of `viewItem =~ /aks\\.cluster/i`

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,64 @@
...
+    } else {
+        return;
+    }
+}
+
+async function isDeployToAzureExtensionInstalled(): Promise<boolean> {
+    const pipelinesExtension = vscode.extensions.getExtension(deployToAzureExtensionId);
+    if (!pipelinesExtension) {
+        try {
+            vscode.window.showWarningMessage('Please install/enable `Deploy to Azure` extension and then retry.');
```
bishal-pdMSFT, line 48: why is this wrapped in try-catch. Do you expect this to fail?

src/commands/cicdPipeline.ts
kanika1894: But this can come here if the cloudProvider isn't Azure.

src/commands/cicdPipeline.ts
kanika1894: No  failure expected
WIll remove the try catch

package.json
```diff
@@ -66,6 +82,16 @@
                 {
                     "command": "aks.periscope",
                     "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i"
+                },
+                {
+                    "command": "extension.BrowseCICDPipeline",
```
bishal-pdMSFT, line 87: should we use our extension name instead of `extension`? `azure-deploy.BrowseCICDPipeline`

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,60 @@
...
+        if (!!cluster) {
+            await executeDeployToAzureExtensionInstalled(configurePipelineCommand, cluster);
+        }
+    }
+}
+
+export async function browsePipeline(context: IActionContext, target: any): Promise<void> {
+    if (await isDeployToAzureExtensionInstalled()) {
+        const cluster = await getCluster(target);
+        if (!!cluster) {
```
Tatsinnit, line 23: 💡 Wondering why there is the use of `double bang` i.e. `!!`.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,60 @@
...
+import { IActionContext } from 'vscode-azureextensionui';
+
+const deployToAzureExtensionId = 'ms-vscode-deploy-azure.azure-deploy';
+const configurePipelineCommand = 'configure-cicd-pipeline';
+const browsePipelineCommand = 'browse-cicd-pipeline';
+
+export async function configurePipeline(context: IActionContext, target: any): Promise<void> {
+    if (await isDeployToAzureExtensionInstalled()) {
+        const cluster = await getCluster(target);
+        if (!!cluster) {
```
Tatsinnit, line 14: 💡 Wondering why there is use of `double bang- !!` because `getCluster` returns either the `clustertreeitem` or `undefined`.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,60 @@
...
+import AksClusterTreeItem from '../tree/aksClusterTreeItem';
+import { IActionContext } from 'vscode-azureextensionui';
+
+const deployToAzureExtensionId = 'ms-vscode-deploy-azure.azure-deploy';
+const configurePipelineCommand = 'configure-cicd-pipeline';
+const browsePipelineCommand = 'browse-cicd-pipeline';
+
+export async function configurePipeline(context: IActionContext, target: any): Promise<void> {
+    if (await isDeployToAzureExtensionInstalled()) {
+        const cluster = await getCluster(target);
```
Tatsinnit, line 13: 💡 📓 idea share: seems like this code is repetitive:

* we can easily move this to a private funct, and then call it from both `configurePipeline` and `browsePipeline` .



```
if (await isDeployToAzureExtensionInstalled()) {
        const cluster = await getCluster(target);
            if (!!cluster) {
await executeDeployToAzureExtensionInstalled(browsePipelineCommand, cluster);
        }
    }
```

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+    return true;
+}
+
+async function executeDeployToAzureExtensionInstalled(commandToRun: string, cluster: any): Promise<void> {
+    const listOfCommands = await vscode.commands.getCommands();
+    if (listOfCommands.find((commmand: string) => commmand === commandToRun)) {
+        return vscode.commands.executeCommand(commandToRun, cluster);
+    }
+    throw new Error(utils.format('Unable to find command %s. Make sure `Deploy to Azure` extension is installed and enabled.', commandToRun));
```
Tatsinnit, line 59: 💡 Idea, we could do this: `Unable to find command${commandToRun}. Make sure `Deploy to Azure` extension is installed and enabled.`, also, how is this `Error` gets presented to the user?

src/extension.ts
```diff
@@ -30,6 +31,8 @@ export async function activate(context: vscode.ExtensionContext) {
         registerCommandWithTelemetry('aks.selectSubscriptions', selectSubscriptions);
         registerCommandWithTelemetry('aks.detectorDiagnostics', detectorDiagnostics);
         registerCommandWithTelemetry('aks.periscope', periscope);
+        registerCommand('azure-deploy.ConfigureCICDPipeline', configurePipeline),
```
Tatsinnit, line 34: 💡 Idea: if you are keen to make these commands be part of telemetry, use `registerCommandWithTelemetry` Also, If I have time, I can help you with this work if you need hand, but don't entirely know the scope yet so you can find me with internal alias. Thanks heaps.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,60 @@
...
+        }
+    } else {
+        return;
+    }
+}
+
+async function isDeployToAzureExtensionInstalled(): Promise<boolean> {
+    const pipelinesExtension = vscode.extensions.getExtension(deployToAzureExtensionId);
+    if (!pipelinesExtension) {
+        vscode.window.showWarningMessage('Please install/enable `Deploy to Azure` extension and start again.');
```
Tatsinnit, line 47: 💡 regarding the flow of this info, here in line 47 we display `warning` and then we go ahead and `execute command` ? and if you see carefully we have `await` but we return `false` after execution, wondering why?

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (cloudExplorer.available) {
+        const clusterTarget = cloudExplorer.api.resolveCommandTarget(target);
+        if (clusterTarget && clusterTarget.cloudName === "Azure" && clusterTarget.nodeType === "resource" && clusterTarget.cloudResource.nodeType === "cluster") {
+            return clusterTarget.cloudResource as AksClusterTreeItem;
+        } else {
+            vscode.window.showInformationMessage('This command only applies to AKS clusters.');
+            return;
+        }
+    } else {
+        return;
```
Tatsinnit, line 40: 💡 Idea: We can return for the failure case early.

Tatsinnit (changes requested):
💡 Thank you for this, added 2 cents of ideas below, also I am happy to help-out as well (whenever is cool). I will let Ivan to comment more. Also please let me know if you have internal design docs, you can add us to understand the ideas / thought-works as build up idea. Kind Regards

src/commands/cicdPipeline.ts
kanika1894: This was to have the value converted to boolean and execute further commands only if the cluster existed.
Now, with the code refactoring, this is no longer required.

src/commands/cicdPipeline.ts
kanika1894: Agreed.
Have made changes accordingly.

src/commands/cicdPipeline.ts
kanika1894: The error windows appears to the user as follows :
image

src/commands/cicdPipeline.ts
kanika1894: Didn't understand ?

bishal-pdMSFT: If I understand correctly, the ask here is to return earlier in case of failure. So, you can do this at the top of function
```
if (!cloudExplorer.available) {
    return;
}
```

src/commands/cicdPipeline.ts
bishal-pdMSFT: @Tatsinnit the idea here is to open the VS Code extension marketplace page so that it is easy for user to go ahead and install the extension. That's what `extension.open` command is doing. And we are waiting for the markeplace page to open before we return from the current command.

bishal-pdMSFT: We are returning `false` as we can not proceed without user actually going and installing the extension. Hence we exit from current command and let user install the extension and re-trigger the command.

src/commands/cicdPipeline.ts
Tatsinnit: 💡Thought share: Plus a minor addition: same goes for the inner `empty return` can we please change it to `return undefined` just to keep consistent in the extension, I think anywhere we will explicitly do return undefined in anywhere already existing code as well. + It sits well with the functionalists `annotation` return type as well.

bishal-pdMSFT:
> 💡 Thank you for this, added 2 cents of ideas below, also I am happy to help-out as well (whenever is cool). I will let Ivan to comment more. Also please let me know if you have internal design docs, you can add us to understand the ideas / thought-works as build up idea. Kind Regards
> 
> I also have a feeling that the comments here [Azure/vscode-kubernetes-tools#818 (comment)](https://github.com/Azure/vscode-kubernetes-tools/pull/818#issuecomment-694805382) will stand true for this extension as well.
> 
> An initial design flow Doc (it exist) will help in understanding and contribute better for this change. Thanks

@Tatsinnit this PR is result of discussion with Ralph and Ivan. If you share your email id, I will forward the mail to you.
Also, I would suggest you to try out the "[Deploy to Azure](https://marketplace.visualstudio.com/items?itemName=ms-vscode-deploy-azure.azure-deploy)" extension and see how it helps in creating a GitHub workflow for deploying user's code to AKS namespace.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,53 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import * as utils from 'util';
+import { IActionContext } from 'vscode-azureextensionui';
+
+const deployToAzureExtensionId = 'ms-vscode-deploy-azure.azure-deploy';
+const configurePipelineCommand = 'configure-cicd-pipeline';
+const browsePipelineCommand = 'browse-cicd-pipeline';
+
+export async function configurePipeline(context: IActionContext, target: any): Promise<void> {
```
Tatsinnit, line 10: 💡 🔂 Repetitive duplicate function: both `configurePipeline` and `browsePipeline` are `duplicate` functions, i.e. it is repetitive code and only thing changed is the `const, we can merge the 2 functions with better name, and supply that via `register`. = )

Another question for you guys is: It might also be worth to share the thought process behind this change for both Kubernetes and aks services extension.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (await isDeployToAzureExtensionInstalled()) {
+        await getClusterAndExecute(target, configurePipelineCommand);
+    }
+}
+
+export async function browsePipeline(context: IActionContext, target: any): Promise<void> {
+    if (await isDeployToAzureExtensionInstalled()) {
```
Tatsinnit, line 17: 💡`nit` for clarity: I would extract it to a `const` with logical name.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+    return true;
+}
+
+async function executeDeployToAzureExtensionInstalled(commandToRun: string, cluster: any): Promise<void> {
+    const listOfCommands = await vscode.commands.getCommands();
+    if (listOfCommands.find((commmand: string) => commmand === commandToRun)) {
+        return vscode.commands.executeCommand(commandToRun, cluster);
+    }
+    throw new Error(utils.format(`Unable to find command ${commandToRun}. Make sure Deploy to Azure extension is installed and enabled.`));
```
Tatsinnit, line 52: 💡 `Error(utils.format())` What will this `Error` throw will do for the end-user, i.e. do you guys also need something like `showErrorMessage` to display it to the end user?

bishal-pdMSFT:
@Tatsinnit here is the doc which describes the flow: https://microsoft.sharepoint.com/:p:/t/ReleaseManagement/RM/EY3nQXsm2Q1BiV1pZth9btkBYw5KLb2TGBy-Gawaejjtgw?%7BclickTime%7D=1600265378427

Let us know if you need more details.

src/extension.ts
```diff
@@ -30,6 +31,8 @@ export async function activate(context: vscode.ExtensionContext) {
         registerCommandWithTelemetry('aks.selectSubscriptions', selectSubscriptions);
         registerCommandWithTelemetry('aks.detectorDiagnostics', detectorDiagnostics);
         registerCommandWithTelemetry('aks.periscope', periscope);
+        registerCommandWithTelemetry('azure-deploy.ConfigureCICDPipeline', configurePipeline),
```
Tatsinnit, line 34: 🐞 missing semicolon `;`

src/extension.ts
```diff
@@ -30,6 +31,8 @@ export async function activate(context: vscode.ExtensionContext) {
...
+        registerCommandWithTelemetry('azure-deploy.BrowseCICDPipeline', browsePipeline)
```
Tatsinnit, line 35: 🐞 Missing Semicolon `;`

src/commands/cicdPipeline.ts
bishal-pdMSFT: Is there a top level logic which checks for any error and shows a message to user? If no (which seems to be the case here), then yes we should be showing error to user rather than throw.

kanika1894: Just throwing the error gives the error to user  as below
image

Tatsinnit: 💡 Re: `Error Handling` - I would recommend to error-handle known `error` as extension level `error-handling` i.e. see this sample when you use: `vscode.window.showErrorMessage('Hello World!');` have a play around and possible handle this way because in the event of "firing" and let vscode handle it, it will have the button of "Report an issue" i.e. report this although what I see is that this is known issue for the work in PR and should be handle that way. 

image


Please also feel free to ping me in teams if you want to chatup.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (await isDeployToAzureExtensionInstalled()) {
+        await getClusterAndExecute(target, browsePipelineCommand);
+    }
+}
+
+async function getClusterAndExecute(target: any, command: string): Promise<void> {
+    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+    if (!cloudExplorer.available) {
+        return undefined;
+    } else {
```
Tatsinnit, line 26: 💡 Just an idea: (I feel `else` is no longer in need?)

sample Pseudo code is:

```
if (!available) {
    return undefined;
}

// do stuff because I have the stuff available? so I guess else is not longer in need here? 

```

package.json
```diff
@@ -44,13 +44,29 @@
...
             }
         ],
         "menus": {
             "commandPalette": [
                 {
                     "command": "aks.selectSubscriptions",
                     "when": "never"
+                },
+                {
+                    "command": "azure-deploy.ConfigureCICDPipeline",
```
Tatsinnit, line 64: 💡 Camel casing please, for consistency with other commands in this extension.

package.json
```diff
@@ -44,13 +44,29 @@
...
+                    "when": "never"
+                },
+                {
+                    "command": "azure-deploy.BrowseCICDPipeline",
```
Tatsinnit, line 68: 💡 Same here: camel casing please, for consistency with other commands in this extension.

package.json
```diff
@@ -66,6 +82,16 @@
                 {
                     "command": "aks.periscope",
                     "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i"
+                },
+                {
+                    "command": "azure-deploy.BrowseCICDPipeline",
```
Tatsinnit, line 87: 💡 With camel casing changes this and the one below will be effected as well. (Changes for consistency)

Tatsinnit:
💡 Suggesstion please: Grouping these new feature into separate `group` not with existing `save kubeconfig and Merge into Kubeconfig`

Like for idea sake move these 2 after the "Run" -- features? (Separated by the horizontal line) 

image

Thanks, and also cc: @itowlson

Tatsinnit:
💡 Opinions\Suggestions:  Umbrella level comment regarding error handling, because this error is not visible in the PR, but I checkout the fork just to play around, and we need to better represent these error at the scope of this extension.

We should clearly handle the errors below when run under this extension.

* How I produces below error: `right click` on cluster and choose --> `Browse All Pipelines` 

image

Thanks heaps.

kanika1894:
> 💡 `Organizing Suggestion please`: Grouping these new feature into separate `group` not with existing `save kubeconfig and Merge into Kubeconfig`
> 
> Like for idea sake move these 2 after the "Run" -- features? (Separated by the horizontal line)
> 
> image
> 
> Thanks, and also cc: @itowlson
> 
> Also please note :) there are bunch of comments which are marked as `resolved` but actually changes don't seem to have pushed yet. I believe that might be because of local copy changes, just mentioning here to record it for just in case we forget. Thanks.

The view and order of the view buttons was I guess approved before. @pulkitaggarwl will be the right person to address this.

Yes, will be pushing the local changes soon.

src/commands/cicdPipeline.ts
kanika1894: > 💡 **Suggestion**: We should handle error with better information, i.e. for example below error which is visible from `AKS Extension` as part of this work, and is not visible in the PR. I checkout the fork just to play around, and I think we should better represent these error at the scope of this extension.
> 
> I think we should clearly handle the errors below when run under this extension.
> 
> * How I produces below error: `right click` on cluster and choose --> `Browse All Pipelines`
> 
> Questions: (Without knowing much about the "deploy to azure" extension.)
> 
> Below question is to build up to the self question I asked myself that: Can we better handle this scenario?
> 
> * What is the scenario which cause folks to directly open the issue against the parent git repository for this issue?
> * Wondering why it is not handled in the extension level with possible docs information as to how to fix it, because with this `open issue` option as a user I have nothing else to report but just the information that I took one random cluster and clicked the `Browse All Pipelines`.
> 
> image
> 
> Thanks heaps.

I guess before we go much into the error handling part , I will explain a bit about how are we trying to implement the fucnctionality.
The menu items which are added in this extension, invoke the commands exposed by Deploy to Azure extension. SO, in parallel, two things are being developed. UI support on this extension as well as backend support in Deploy to Azure etension. The error which you received on clicking on Browse all pipelines is because the browse functionality isn't released yet, and that's why it is breaking the flow.

Now, when it comes to error throwing mechanism., we believe the "Source" part of the error message thrown provides more clarity from which extension the error is coming. This is internal handling by VSCode as we are not doing anything explictly for this. Also, the message that "raise a issue in this repo" is pointing to the repo responsible for this issue. Is there something which you  feel should be done additional, we can surely discuss.

kanika1894:
> 💡 **Suggestion**: We should handle error with better information, i.e. for example below error which is visible from `AKS Extension` as part of this work, and is not visible in the PR. I checkout the fork just to play around, and I think we should better represent these error at the scope of this extension.
> 
> I think we should clearly handle the errors below when run under this extension.
> 
> * How I produces below error: `right click` on cluster and choose --> `Browse All Pipelines`
> 
> Questions: (Without knowing much about the "deploy to azure" extension.)
> 
> Below question is to build up to the self question I asked myself that: Can we better handle this scenario?
> 
> * What is the scenario which cause folks to directly open the issue against the parent git repository for this issue?
> * Wondering why it is not handled in the extension level with possible docs information as to how to fix it, because with this `open issue` option as a user I have nothing else to report but just the information that I took one random cluster and clicked the `Browse All Pipelines`.
> 
> image
> 
> Thanks heaps.

I guess before we go much into the error handling part , I will explain a bit about how are we trying to implement the fucnctionality.
The menu items which are added in this extension, invoke the commands exposed by Deploy to Azure extension. SO, in parallel, two things are being developed. UI support on this extension as well as backend support in Deploy to Azure etension. The error which you received on clicking on Browse all pipelines is because the browse functionality isn't released yet, and that's why it is breaking the flow.

Now, when it comes to error throwing mechanism., we believe the "Source" part of the error message thrown provides more clarity from which extension the error is coming. This is internal handling by VSCode as we are not doing anything explictly for this. Also, the message that "raise a issue in this repo" is pointing to the repo responsible for this issue. Is there something which you feel should be done additional, we can surely discuss.

src/commands/cicdPipeline.ts
Tatsinnit: * Cool, sounds good. In that case can we please set this PR as Draft PR, because the underlining work which this PR is trying to plumb I.e. “browse all pipeline” is not ready / released yet.

* Yes please I believe regarding error handling we should be as explicit as possible that the source of the error should be clearly underlined rather then invoking the same message which the other extension is using.

A good error handling will not only make the extension control flow clear but also make users workflow cleaner.

Thanks

package.json
Tatsinnit: :bulb: I think for consistent camel casing like “selectSubscripts” I.e. I was referring to lower camel casing, so say for example: “BrowseCICDPipeline” will become “browseCICDPipeline or browserCicdPipeline”

This will impact the other resolves camel casing comments to be consistent with the AkS extension commands casing.

Just setting this one as unresolved but it stands true for both the commands. Thank you.

Tatsinnit:
> > 💡 **Suggestion**: We should handle error with better information, i.e. for example below error which is visible from `AKS Extension` as part of this work, and is not visible in the PR. I checkout the fork just to play around, and I think we should better represent these error at the scope of this extension.
> > I think we should clearly handle the errors below when run under this extension.
> > 
> > * How I produces below error: `right click` on cluster and choose --> `Browse All Pipelines`
> > 
> > Questions: (Without knowing much about the "deploy to azure" extension.)
> > Below question is to build up to the self question I asked myself that: Can we better handle this scenario?
> > 
> > * What is the scenario which cause folks to directly open the issue against the parent git repository for this issue?
> > * Wondering why it is not handled in the extension level with possible docs information as to how to fix it, because with this `open issue` option as a user I have nothing else to report but just the information that I took one random cluster and clicked the `Browse All Pipelines`.
> > 
> > image
> > Thanks heaps.
> 
> I guess before we go much into the error handling part , I will explain a bit about how are we trying to implement the fucnctionality.
> The menu items which are added in this extension, invoke the commands exposed by Deploy to Azure extension. SO, in parallel, two things are being developed. UI support on this extension as well as backend support in Deploy to Azure etension. The error which you received on clicking on Browse all pipelines is because the browse functionality isn't released yet, and that's why it is breaking the flow.
> 
> Now, when it comes to error throwing mechanism., we believe the "Source" part of the error message thrown provides more clarity from which extension the error is coming. This is internal handling by VSCode as we are not doing anything explictly for this. Also, the message that "raise a issue in this repo" is pointing to the repo responsible for this issue. Is there something which you feel should be done additional, we can surely discuss.

* Cool, sounds good. In that case can we please set this PR as Draft PR, because the underlining work which this PR is trying to plumb I.e. “browse all pipeline” is not ready / released yet.

* Yes please I believe regarding error handling we should be as explicit as possible that the source of the error should be clearly underlined rather then invoking the same message which the other extension is using.

A good error handling will not only make the extension control flow clear but also make users workflow cleaner.

Thanks

bishal-pdMSFT:
> In that case can we please set this PR as Draft PR, because the underlining work which this PR is trying to plumb I.e. “browse all pipeline” is not ready / released yet.

@Tatsinnit A suggestion - Can we please not make it draft as the current PR is actively being worked upon and reviewed. A "draft" might stall the PR itself. The other part is already being worked upon in the other extension. We will not merge this PR unless the other work is done. However, need not make it draft.

bishal-pdMSFT:
> Yes please I believe regarding error handling we should be as explicit as possible that the source of the error should be clearly underlined rather then invoking the same message which the other extension is using.

@Tatsinnit can you please suggest how we can make it better. Are you saying:
1. We put the extension name ("Deploy to Azure") in error message?
2. Or should we have two error messages shown to user - one from ASK extension and one from Deploy-to-Azure ?

Tatsinnit:
📓  Replying both the questions \ comment under one comment,
 
> > In that case can we please set this PR as Draft PR, because the underlining work which this PR is trying to plumb I.e. “browse all pipeline” is not ready / released yet.

> https://github.com/Azure/vscode-aks-tools/pull/42#issuecomment-698863207 
> @Tatsinnit A suggestion - Can we please not make it draft as the current PR is actively being worked upon and reviewed. A "draft" might stall the PR itself. The other part is already being worked upon in the other extension. We will not merge this PR unless the other work is done. However, need not make it draft.

> https://github.com/Azure/vscode-aks-tools/pull/42#issuecomment-698863978
> @Tatsinnit can you please suggest how we can make it better. Are you saying:

> We put the extension name ("Deploy to Azure") in error message?
> Or should we have two error messages shown to user - one from ASK extension and one from Deploy-to-Azure ?

 > https://github.com/Azure/vscode-aks-tools/pull/42#issuecomment-698819997 
>> The menu items which are added in this extension, invoke the commands exposed by Deploy to Azure extension. SO, in parallel, two things are being developed. UI support on this extension as well as backend support in Deploy to Azure etension. The error which you received on clicking on Browse all pipelines is because the browse functionality isn't released yet, and that's why it is breaking the flow.

Thanks heaps to you guys for engaging and replying.

+1, cool @bishal-pdMSFT sounds good, I am happy if you want to keep working on this pull request in open state. For me draft PR is **not** backburner but rather the PR which seems like still at implementation level not entirely thought through / matured. 😊 (mentioning couple of things below which might help to build our call which you guys have booked today)

+1 and Thank you @kanika1894 for further explanation around parallel work which adds more clarity to this implementation. I have also received the meeting invite for today evening my time so we can share whatever information is not clear. 

Below I will detail couple of high level key points.

* **Please Describe implementation level flow in description PR description** : (_Explanation why that will help in context building_)  Given that it was mentioned there are 2 parallel work I think this needs to be clearly detailed in description section with the workitem which needs to go first.  (I would recommend to add more details in the description of the PR as all anyone can interpret is from the implementation and the way core is handled. With every commit sometimes the entire implementation changes. Example of first commit va latest.) where earlier a const string for browse pipeline was used and in the very last or second to last commit that mechanism is changed, I am not entirely sure the context \ implementation-level design decision for the idea\reason behind that change. 

* **Error Handling information for end-user**: Please make sure that the error messaging should clearly guide user workflow to know that error caused, my thinking is that: steps user can take to rectify or report, and more information is clear from the messaging. **For example** from the messaging of current error which was flowing form the underlying extension, it just send me as a user to the GitHub issue page of "Azure Deploy" repository, which is little unclear (Or might be I am missing some context). Might be something like (Just from what I can come up with) : "Deploy Azure Failed: <THE_REASON_IT_FAILED>. Please read possible issues which might cause this (link here) and report this specific issue to the Deploy Azure Extension (Here) + (What they should report it with?)" 

* Might be link the other Parallel PR work to this PR as a dependent, hence the order of "when it is ready" is clear.

* **Supporting Readme / docs changes**: Please Don’t forget to update read me docs along with these changes, for the new command.

I have added couple of other review comments around use of internal tenant and preview URL use which will not work for non- internal audience. Also for security reasons as well we should not use any internal tenant ids. 

Hope it helps and thanks, and talk to you guys soon.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,68 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+import { CloudExplorerV1 } from 'vscode-kubernetes-tools-api';
+
+const deployToAzureExtensionId = 'ms-vscode-deploy-azure.azure-deploy';
+const configurePipelineCommand = 'configure-cicd-pipeline';
+const azurePortalUrl = 'https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/';
```
Tatsinnit, line 8: 🐛 I believe `Microsoft.onmicrosoft.com` is internal Microsoft tenant. (This will not be **valid** for general public use)

Tatsinnit, line 8: 🐛 I think `ms.portal.azure.com` the internal preview version of the portal.

* For external it should be `portal.azure.com`. (Please double check \ Test)

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export async function configurePipeline(context: IActionContext, target: any): Promise<void> {
+    if (await isDeployToAzureExtensionInstalled()) {
+        await getClusterAndExecute(target, configurePipelineCommand);
+    }
+}
+
+export async function browsePipeline(context: IActionContext, target: any): Promise<void> {
+    const deploymentCenterUrl: string = await getDeploymentCenterUrl(target);
```
Tatsinnit, line 17: 💡 `const` level annotations is not required as the `funct` will return the desired type.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (!pipelinesExtension) {
+        vscode.window.showWarningMessage('Please install/enable `Deploy to Azure` extension and start again.');
+        await vscode.commands.executeCommand('extension.open', deployToAzureExtensionId);
+        return false;
+    }
+    return true;
+}
+
+async function executeDeployToAzureExtensionInstalled(commandToRun: string, cluster: any): Promise<void> {
+    const listOfCommands = await vscode.commands.getCommands();
```
Tatsinnit, line 63: 💡 minor: pass `true` for `getCommands` unless you want the internal commands to be returned as well. more details here: https://code.visualstudio.com/api/references/vscode-api

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,68 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+import { CloudExplorerV1 } from 'vscode-kubernetes-tools-api';
+
+const deployToAzureExtensionId = 'ms-vscode-deploy-azure.azure-deploy';
+const configurePipelineCommand = 'configure-cicd-pipeline';
```
Tatsinnit, line 7: 💡 Idea: Either use this const in all the places where you are passing commands, or I believe you already have a mechanism to pass through chain of functions so you can remove this and command declaration can reside inside the function?

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,68 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+import { CloudExplorerV1 } from 'vscode-kubernetes-tools-api';
+
+const deployToAzureExtensionId = 'ms-vscode-deploy-azure.azure-deploy';
```
Tatsinnit, line 6: 💡 I believe this can also reside inside : `isDeployToAzureExtensionInstalled` Done need to be class level `const`.

src/commands/cicdPipeline.ts
Tatsinnit: 📓 This whole flow is changed to url forwarding. cc: @kanika1894 - Resolving this for and thank you for organising the chat.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,69 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+import { CloudExplorerV1 } from 'vscode-kubernetes-tools-api';
+
+export async function configurePipeline(context: IActionContext, target: any): Promise<void | undefined> {
+    const deployToAzureExtensionInstalled = await isDeployToAzureExtensionInstalled();
+    if (!deployToAzureExtensionInstalled) {
+        return undefined;
+    } else {
```
Tatsinnit, line 10: 💡 Since we bail out early this else is redundant in my opinion. So remove this else and leave rest of the code as it is. Let me know if it’s clear.

src/commands/cicdPipeline.ts
```diff
@@ -0,0 +1,69 @@
+import * as vscode from 'vscode';
```
Tatsinnit, line 1: 💡 Since your PR has griwn this far, I realised the commands in use i.e. `configure` and `browse` both are no longer coupled at all, so I have tried to just re-organise your code for you here, and put a PR to your fork.

* https://github.com/kanika1894/vscode-aks-tools/pull/1

If you like it, you can merge it to your fork and it should appear as part of this PR.

📓 Please also note: Add the reference to the **other** work item which needs to be deployed before this work in this PR as well please. I fee, under description will be a good place. 

Thanks heaps.

Tatsinnit:
🎊 Looks good to me, Thank you cc: @itowlson for his eyes before this gets merged 🙏 . We use squash merge for this. @kanika1894 please note, we need to wait for the underlying workitem to complete first as well.

Tatsinnit (approved):

src/commands/deployAzurePipeline/browsePipeline.ts
```diff
@@ -0,0 +1,24 @@
...
+        vscode.window.showErrorMessage(`Unable to browse pipelines for the resource. Select appropriate cluster.`);
+    }
+}
+
+async function getDeploymentCenterUrl(target: any): Promise<string | undefined> {
+    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+    if (cloudExplorer.available) {
+        const clusterTarget = cloudExplorer.api.resolveCommandTarget(target) as CloudExplorerV1.CloudExplorerResourceNode;
+        if (clusterTarget) {
+            return `https://portal.azure.com/#@${clusterTarget.cloudResource.session.tenantId}/resource${clusterTarget.cloudResource.id}/deloymentCenter`;
```
Tatsinnit, line 20: 💡 Worth to get idea: if we should use `root` or `session` for the `tenatnid` both as same values. Thanks.

kanika1894: @itowlson need your inputs here.
Also, @Tatsinnit , as discussed, I did the testing using a non-microsoft account as well, it worked fine. :)

itowlson, line 20: This URL includes the string `deloymentCenter` (no `p` in `deployment`).  Is this is an error in the portal URL that we need to just get along with?  If so, is there any risk that the portal will fix the typo?

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+        return undefined;
+    }
+
+}
+
+async function isDeployToAzureExtensionInstalled(): Promise<boolean> {
+    const deployToAzureExtensionId = 'ms-vscode-deploy-azure.azure-deploy';
+    const pipelinesExtension = vscode.extensions.getExtension(deployToAzureExtensionId);
+    if (!pipelinesExtension) {
+        vscode.window.showWarningMessage('Please install/enable `Deploy to Azure` extension and start again.');
```
itowlson, line 31: A function named `isXxxInstalled()` shouldn't display UI itself - it should just return whether the thing is installed or not.  Then the top level function can decide what UI to display if the thing isn't installed.  Suggest moving this message to the `configurePipeline` function.

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+        await vscode.commands.executeCommand('extension.open', deployToAzureExtensionId);
```
itowlson, line 32: I'm not sure about doing this automatically.  I see that it gives the user some context for the message; but perhaps consider putting a button on the 'not installed' message saying "View in Marketplace" or "View in Extensions Tab" or something like that.

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+
+export async function configurePipeline(context: IActionContext, target: any): Promise<void | undefined> {
```
itowlson, line 5: This is a top level command handler so I don't think anything can consume the return value?  Just `Promise<void>` if anything.

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+    const deployToAzureExtensionInstalled = await isDeployToAzureExtensionInstalled();
+    if (!deployToAzureExtensionInstalled) {
+        return undefined;
```
itowlson, line 8: I don't think there's any need to return the `undefined` value here.

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+        await executeDeployToAzureExtensionInstalled(configurePipelineCommand, clusterTarget);
+    } else {
+        vscode.window.showInformationMessage('This command only applies to AKS clusters.');
+        return undefined;
+    }
+
+}
+
+async function isDeployToAzureExtensionInstalled(): Promise<boolean> {
+    const deployToAzureExtensionId = 'ms-vscode-deploy-azure.azure-deploy';
```
itowlson, line 28: Consider pulling this out to a top-level constant (in `SHOUTY_CAPS`) so it jumps out as a true 'magic string' constant rather than just an immutable variable.

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+    if (!cloudExplorer.available) {
+        return undefined;
+    }
+
+    const clusterTarget = cloudExplorer.api.resolveCommandTarget(target);
+    if (clusterTarget && clusterTarget.cloudName === "Azure" && clusterTarget.nodeType === "resource" && clusterTarget.cloudResource.nodeType === "cluster") {
+        const configurePipelineCommand = 'configure-cicd-pipeline';
```
itowlson, line 18: Consider pulling this out to a top-level constant (in `SHOUTY_CAPS`) so it jumps out as a true 'magic string' constant rather than just an immutable variable.

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+    const pipelinesExtension = vscode.extensions.getExtension(deployToAzureExtensionId);
+    if (!pipelinesExtension) {
+        vscode.window.showWarningMessage('Please install/enable `Deploy to Azure` extension and start again.');
+        await vscode.commands.executeCommand('extension.open', deployToAzureExtensionId);
+        return false;
+    }
+    return true;
+}
+
+async function executeDeployToAzureExtensionInstalled(commandToRun: string, cluster: any): Promise<void> {
```
itowlson, line 38: Can we give a more specific type to `cluster`?

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+    const listOfCommands = await vscode.commands.getCommands(true);
+    if (listOfCommands.find((commmand: string) => commmand === commandToRun)) {
+        try {
+            vscode.commands.executeCommand(commandToRun, cluster);
```
itowlson, line 42: I thought the point of putting this code into the AKS extension instead of the Deploy To Azure extension was that DTA could define a contract, and the code on the AKS side would marshal the cluster data into that format, so that if AKS changed its internal format we wouldn't break DTA.  If you're just passing the cluster tree object, then you don't get this decoupling or the safety that comes with it, and you might as well just hang the DTA command directly off the cluster tree rather than implementing and maintaining a stub command here...?

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+        }
+        catch (error) {
+            vscode.window.showErrorMessage("Please report the issue with [Deploy to Azure extension](https://github.com/microsoft/vscode-deploy-azure/issues), execute command failed with error:" + error);
```
itowlson, line 45: Comma splice 
```suggestion
            vscode.window.showErrorMessage("Command failed. Please report this issue with [Deploy to Azure extension](https://github.com/microsoft/vscode-deploy-azure/issues). Error details:" + error);
```

src/commands/deployAzurePipeline/configurePipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+        vscode.window.showWarningMessage('Please install/enable `Deploy to Azure` extension and start again.');
+        await vscode.commands.executeCommand('extension.open', deployToAzureExtensionId);
+        return false;
+    }
+    return true;
+}
+
+async function executeDeployToAzureExtensionInstalled(commandToRun: string, cluster: any): Promise<void> {
+    const listOfCommands = await vscode.commands.getCommands(true);
+    if (listOfCommands.find((commmand: string) => commmand === commandToRun)) {
```
itowlson, line 40: Shouldn't need the type annotation here?

itowlson, line 40: Consider using `includes` rather than `find`: 
```suggestion
    if (listOfCommands.includes(commandToRun)) {
```

package.json
```diff
@@ -44,13 +44,29 @@
             {
                 "command": "aks.periscope",
                 "title": "Run AKS Periscope"
+            },
+            {
+                "command": "azure-deploy.configureCicdPipeline",
+                "title": "Configure CI/CD"
```
itowlson, line 50: This is going out to all AKS users, not just people who choose to install the CI/CD pipelines extension.  So it needs consistent and explanatory language.  I don't know enough about the feature to be sure what to suggest, but maybe `Configure CI/CD Pipelines` and `Browse CI/CD Pipelines`.  Possibly even go as far as `Configure Azure CI/CD Pipelines` because all _my_ CI/CD pipelines are on GitHub and I would be puzzled if the AKS extension thought it knew about them!

itowlson (changes requested):
This looks broadly okay, but there are a few UI issues, and one key question relating to the justification for having this code in the AKS extension at all.  Other things are mostly nitpicks/suggestions.

src/commands/deployAzurePipeline/configurePipeline.ts
Tatsinnit: 💡 I believe it means moving this back to the top level with : sample: `const MY_CONSTANT = "blah";` 

* Please note you need to make sure this is well documented in the your extension so that the command line et. al. don't change and if it does people should be aware of this.

package.json
kanika1894: @pulkitaggarwl please let me know what will the updated strings be

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: @Tatsinnit we deliberately moved it to function level as it is being used only in that particular function. Just trying to understand the reason behind moving it back at the top level?

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: Will update it.

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: sure

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: @Tatsinnit I remember we put this up on the suggestion to have a proper failure/success scenario. 
Just double checking because thought this value isn't causing any difference to the flow but just conflicting review comments. Trying to minimize the rework.

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: ``` await vscode.commands.executeCommand('extension.open', deployToAzureExtensionId); ```
This command only opens the extension in VSCode and doesn't install it automatically. 
@itowlson any particular concern you have around this?

src/commands/deployAzurePipeline/browsePipeline.ts
kanika1894: yeah, it's an error in the portal error.
If the portal fixes the same, this will have to be updated.

src/commands/deployAzurePipeline/configurePipeline.ts
Tatsinnit: @kanika1894 cool, I guess check the signature of other places please essentially I just did. So please make it consistent I.e. do not return promise undefined from main function hence you might not need to return undefined.

src/commands/deployAzurePipeline/configurePipeline.ts
Tatsinnit: Thanks, @kanika1894 this one I totally remember so here is my take: I am easy with any favour but remember what you had earlier was all lower casing: example: https://github.com/Azure/vscode-aks-tools/pull/42/commits/5fe6ea16e4796c079a7fbeab9f64a9bf6c0d087d#diff-5b5e105bbccdadd756c4442822456a4bR7

Like I have iterated in past I am easy with whatever According to what code design needs to look like and I can see making all in CAPS here kind of highlights the fact that hey these are key Const. so make sense to do it that way. What do you say?

(Just to keep as side note) Also please note we should highlight in the parent extension that all these constant which are actually DTA extension dependencies cannot be changed as it will break these functionality.

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: I didn't really understand this?

Tatsinnit: @kanika1894 here is what I think it means :) check out : your find and the. You have “: string” annotation inside, you don’t need it here. 

Also please note there is another comment somewhere where it is proposed to try and use “.includes” instead and you might end up but efficient code 😊☕️ What do you think? 

Extra: I guess: includes: use it when you want to check if an array contains an exact value in it. This is more efficient than some if you can use it.

src/commands/deployAzurePipeline/configurePipeline.ts
Tatsinnit: 💡 Just adding my 2 cents as observation and merit to the proposed idea: (but it’s open for discussion)

Pro/advantage (button for user): If we follow the button approach: That makes user more aware about the actions they are taking I.e. for first time user it will display message below on right hand corner and followed by automatic market place and I wonder that will be just overwhelming for some user as experience.

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: yeah, using includes makes it more efficient.
I guess the presence of both the commands created the confusion. But all good now. :)

src/commands/deployAzurePipeline/browsePipeline.ts
```diff
@@ -0,0 +1,24 @@
...
+    } else {
+        vscode.window.showErrorMessage(`Unable to browse pipelines for the resource. Select appropriate cluster.`);
+    }
+}
+
+async function getDeploymentCenterUrl(target: any): Promise<string | undefined> {
+    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+    if (cloudExplorer.available) {
+        const clusterTarget = cloudExplorer.api.resolveCommandTarget(target) as CloudExplorerV1.CloudExplorerResourceNode;
+        if (clusterTarget) {
```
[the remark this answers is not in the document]
kanika1894: updated

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: updated

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: updated

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: updated

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: done

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: True. Have defined a new structure which should solve the issue.

src/commands/deployAzurePipeline/configurePipeline.ts
itowlson: @kanika1894 Yes, I understand it only opens the extension, doesn't install it. I was just concerned about whether it was good to shift the user's focus in this way without further interaction, particularly since the visual shift will likely draw their attention more than the message which explains the visual shift.

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksClusterNode.ts
```diff
@@ -0,0 +1,13 @@
+import { CloudExplorerV1 } from "vscode-kubernetes-tools-api";
+
+export class aksClusterNode{
```
Tatsinnit, line 3: 💡 I think if you can declare an interface that is a nice clean way to contract bound `Object` with well defined expected types. Explore this: https://www.typescriptlang.org/docs/handbook/interfaces.html

Tatsinnit (changes requested):

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksClusterNode.ts
bishal-pdMSFT, line 3: We should not have `node` in the name as this definition's purpose is to decouple tree node from the cluster object which DTA extension understands

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksClusterNode.ts
```diff
@@ -0,0 +1,13 @@
+import { CloudExplorerV1 } from "vscode-kubernetes-tools-api";
+
+export class aksClusterNode{
+    nodeType: string;
```
bishal-pdMSFT, line 4: can we call it `resourceType`

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksClusterNode.ts
bishal-pdMSFT: +1. It needs to be an interface.

bishal-pdMSFT: Actually, would it make more sense to use `GenericResource` interface from `azure-arm-resource` package/ This is a very generic definition. AKS extension already takes dependency on this package so we won't adding a new dependency.

We might have to extend it though for adding `subscriptionid`

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: I guess @pulkitaggarwl might be able to better suggest in terms of expected user experience.

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksClusterNode.ts
kanika1894: keeping it type as in GenericResource

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksCluster.ts
```diff
@@ -0,0 +1,13 @@
+import { GenericResource } from 'azure-arm-resource/lib/resource/models';
+import { CloudExplorerV1 } from 'vscode-kubernetes-tools-api';
+
+export class aksCluster implements GenericResource {
```
Tatsinnit, line 4: ❓ Wondering how extending `GenericResource` cater what is discussed here? : https://github.com/Azure/vscode-aks-tools/pull/42#discussion_r499921565

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksClusterNode.ts
Tatsinnit: 💡 I am unsure of using `GenericResource` in this context, and how you guys think it caters this comment here: https://github.com/Azure/vscode-aks-tools/pull/42#discussion_r499921565 

Possibly missing a context so please feel free to add more details? : )

bishal-pdMSFT: @Tatsinnit the goal is to define a contract which is de-coupled from AKS extension's tree node structure. We can either define a custom resource model which can contain properties which are currently required by DTA extension.
Another solution would be to use this `GenericResource` model. The benefits are:
1. It can be used for any other resource as well. For example, if DTA has to integrate with Web App resource, we will not define to update the contract. This will save us from unnecessarily updating AKS extension when we integrate with App service extension.
2. If we need to pass more properties of the cluster to DTA extension in future, we can do so without changing the contract. GenericResource is more flexible and has other fields which can be populated if need be.

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksCluster.ts
bishal-pdMSFT: I think we should _extend_ GenericResource and not _implement_ it

kanika1894: We cannot extend the GenericResource interface.
image

bishal-pdMSFT: Actually extend might not even be needed here. You can compose new interface by including GenericResource and subscriptionId.

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
```diff
@@ -0,0 +1,47 @@
...
+        }
+    } else {
+        vscode.window.showWarningMessage('Please install/enable `Deploy to Azure` extension and start again.');
+        await vscode.commands.executeCommand('extension.open', DEPLOY_TO_AZURE_EXTENSION_ID);
+    }
+}
+
+async function isDeployToAzureExtensionInstalled(): Promise<boolean> {
+    const pipelinesExtension = vscode.extensions.getExtension(DEPLOY_TO_AZURE_EXTENSION_ID);
+    if (!!pipelinesExtension) {
```
Tatsinnit, line 29: 💡 with this new commit why we are using double bang operator here -> `!!` ~~ `(!!pipelinesExtension)` ?

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
```diff
@@ -0,0 +1,47 @@
...
+                vscode.window.showInformationMessage('This command only applies to AKS clusters.');
+            }
+        }
+    } else {
+        vscode.window.showWarningMessage('Please install/enable `Deploy to Azure` extension and start again.');
+        await vscode.commands.executeCommand('extension.open', DEPLOY_TO_AZURE_EXTENSION_ID);
+    }
+}
+
+async function isDeployToAzureExtensionInstalled(): Promise<boolean> {
```
Tatsinnit, line 27: 🐛 This functionalists is no longer `async`, no await.

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksClusterNode.ts
bishal-pdMSFT: @Tatsinnit Does this answer your question?

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
```diff
@@ -0,0 +1,47 @@
...
+const DEPLOY_TO_AZURE_EXTENSION_ID = 'ms-vscode-deploy-azure.azure-deploy';
+
+export async function configurePipeline(context: IActionContext, target: any): Promise<void> {
+    const deployToAzureExtensionInstalled = await isDeployToAzureExtensionInstalled();
+    if (deployToAzureExtensionInstalled) {
+        const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+        if (cloudExplorer.available) {
+            const clusterTarget = cloudExplorer.api.resolveCommandTarget(target);
+            if (clusterTarget && clusterTarget.cloudName === "Azure" && clusterTarget.nodeType === "resource" && clusterTarget.cloudResource.nodeType === "cluster") {
+                const cluster: resourceNode = {resource: clusterTarget.cloudResource, subscriptionId: clusterTarget.cloudResource.subscription.subscriptionId};
```
Tatsinnit, line 15: 💡 No annotation is needed here.

src/commands/deployAzurePipeline/configureCicdPipeline/models/aksClusterNode.ts
Tatsinnit: @bishal-pdMSFT Now after [this commit](https://github.com/Azure/vscode-aks-tools/pull/42/commits/c521426abb70e8043fee5be826d8c12b01daa8f7), it makes sense, and more aligned to what I was thinking. I will let Ivan kindly do a quick take.

https://github.com/Azure/vscode-aks-tools/pull/42/commits/c521426abb70e8043fee5be826d8c12b01daa8f7

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
kanika1894: Isn't it good to have?

Tatsinnit: :bulb: Why is it good? i.e. what are we achieving by the const level type annotation?  **Because** 

* By assigning `{ resource: blah, subscriptionId: blahs};` is achieving that purpose? **Or** please let me know if I missed a context. further you had similar discussion hereh: https://github.com/Azure/vscode-aks-tools/pull/42#discussion_r499922583

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
kanika1894: Since ``` pipelinesExtension  ``` is a non-boolean value.
`` !! `` is used to convert it to a boolean one. 

If it's causing a problem, I'll reverse the condition as ws before.

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
kanika1894: yes, nice catch!
updated :)

src/commands/deployAzurePipeline/browsePipeline.ts
```diff
@@ -0,0 +1,26 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+import { CloudExplorerV1 } from 'vscode-kubernetes-tools-api';
+import { AksClusterTreeNode } from '../../tree/aksClusterTreeItem';
+
+export async function browsePipeline(context: IActionContext, target: any): Promise<void> {
+    const deploymentCenterUrl = await getDeploymentCenterUrl(target);
```
Tatsinnit, line 8: 💡 :) since its no longer an `async`, this specific `await` is redundant. (you forgot this place) https://github.com/Azure/vscode-aks-tools/pull/42/commits/e162e3de51ec5d686421fdefc56c24655adf308b

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
```diff
@@ -0,0 +1,47 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+import { resourceNode } from './models/resourceNode';
+const CONFIGURE_PIPELINE_COMMAND = 'configure-cicd-pipeline';
+const DEPLOY_TO_AZURE_EXTENSION_ID = 'ms-vscode-deploy-azure.azure-deploy';
+
+export async function configurePipeline(context: IActionContext, target: any): Promise<void> {
+    const deployToAzureExtensionInstalled = await isDeployToAzureExtensionInstalled();
```
Tatsinnit, line 9: 💡 :) since its no longer an `async`, this specific `await` is redundant. (you forgot this place) https://github.com/Azure/vscode-aks-tools/pull/42/commits/e162e3de51ec5d686421fdefc56c24655adf308b

kanika1894: ohho!
yeah, updated. thanks :)

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
```diff
@@ -0,0 +1,47 @@
...
+function isDeployToAzureExtensionInstalled(): boolean {
+    const pipelinesExtension = vscode.extensions.getExtension(DEPLOY_TO_AZURE_EXTENSION_ID);
+    if (!pipelinesExtension) {
+        return false;
+    }
+    return true;
+}
+
+async function executeDeployToAzureExtensionInstalled(commandToRun: string, cluster: resourceNode): Promise<void> {
+    const listOfCommands = await vscode.commands.getCommands(true);
```
Tatsinnit, line 36: 💡 Can you please double check if `getCommands` is an `async` function? definition here doesn't specify that. https://code.visualstudio.com/api/references/vscode-api

src/commands/deployAzurePipeline/configureCicdPipeline/models/resourceNode.ts
```diff
@@ -0,0 +1,6 @@
+import { GenericResource } from 'azure-arm-resource/lib/resource/models';
+
+export interface resourceNode extends GenericResource{
```
Tatsinnit, line 3: 💡 **Question**: Why we need `extend` here? because we are `annotating` property -> `resource` as `GenericResource`.

package.json
pulkitaggarwl: Sure we can be more descriptive. we can use

- Configure CI/CD Workflow
- Browse CI/CD Workflows

@itowlson how does these sound ?

src/commands/deployAzurePipeline/configureCicdPipeline/models/resourceNode.ts
kanika1894: yeah, we need nto extend

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
kanika1894: getCommands in not an async function
But usage of ``` includes ``` will require an await. WiIl update.

src/commands/deployAzurePipeline/configurePipeline.ts
pulkitaggarwl: Yeah, we can add a notification with a CTA "Open Extension". 
Here is the notification I am thinking -  "Deploy to Azure" VSCode extension is needed for this command. Please install/enable the extension and try again.  

@itowlson how does that look?

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
Tatsinnit: 💡 I wonder why `await` is added here: `includes` just runs on the `list`.

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes 
* Typescript Playground for `includes` here: https://www.typescriptlang.org/play?#code/MYewdgzgLgBADgUyhGBeGBtA5MAhlLAGhiwBMQBzIkgI3ywF0BuAWACh3RIQAbBAOh6UAFAAMQAVyhwpMAJYoAJAG9EyfnLDAeE0ggjCc9AJQBfUcfZA

src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
```diff
@@ -0,0 +1,47 @@
...
+    const pipelinesExtension = vscode.extensions.getExtension(DEPLOY_TO_AZURE_EXTENSION_ID);
+    if (!pipelinesExtension) {
+        return false;
+    }
+    return true;
+}
+
+async function executeDeployToAzureExtensionInstalled(commandToRun: string, cluster: resourceNode): Promise<void> {
+    const listOfCommands = vscode.commands.getCommands(true);
+    if ((await (listOfCommands)).includes(commandToRun)) {
```
Tatsinnit, line 37: 💡 Please refer to my comments here: https://github.com/Azure/vscode-aks-tools/pull/42#discussion_r502048401 , Thank you.

Tatsinnit: Cool, I see the return type is thenable <string[]> sire the you can await for the line above or whatever you find is best. Can you please double check. Thank you.

kanika1894: cool, will update.

src/commands/deployAzurePipeline/configurePipeline.ts
kanika1894: image
Attaching screenshot for reference.

Tatsinnit (approved):
💡 LGTM, Approved from my end, final take from Ivan will be key. Thank you so much.  🙏 I guess there is couple of last agreement in terms of command naming is left but that is cool with me.

src/commands/deployAzurePipeline/configurePipeline.ts
Tatsinnit: Thanks minor consistency nit: see here : https://github.com/Azure/vscode-aks-tools/pull/42/files#diff-e6f9b94f8c32fa85137a98cfd4e5f259R46 How you have link to the "Deploy to Azure" we could do same link here i.e. `(Deploy to Azure)[link]` Just an Idea.

**Approved** from my end. Passing on for Ivan's eyes for a last take. Thank you so much to everyone.

kanika1894: @itowlson suggested "Show extension" instead of "Open". Have updated the same.

itowlson (approved):
Looks good - thanks for your tenacity and patience!

PR #43, opened 2020-09-28
updating ReadMe for AKS Extension supporting CI/CD pipelines

pulkitaggarwl:


Tatsinnit:
💡 Linked with this work: https://github.com/Azure/vscode-aks-tools/pull/42

README.md
```diff
@@ -68,6 +68,15 @@ Running the AKS Periscope requires you to have a storage account associated with
 
 For more information on Diagnostics settings, visit [Create diagnostic settings to send platform logs and metics to different destinations](https://docs.microsoft.com/azure/azure-monitor/platform/diagnostic-settings).
 
+### Configure CI/CD 
+
+Right click on your AKS cluster and click on **Configure CI/CD** to create a CI/CD pipeline targetting your cluster.
```
Tatsinnit, line 73: 💡 Just for other folks as thought share, not sure if we want to underline that these 2 commands are the hooks for the extension `Deploy to Azure`? aka https://marketplace.visualstudio.com/items?itemName=ms-vscode-deploy-azure.azure-deploy

Tatsinnit:
💡 Please note: you might have missed another place to add this commands which is line 6 or line 7 ( the bullet points which define what all commands fucntiaonlity does via this extension) 

image

README.md
pulkitaggarwl: Expanding CI/CD makes sense.  However we should not need to take a dependency on Deploy to Azure extension for "Browse Pipelines". Why is that needed ? cc: @kanika1894

Tatsinnit: * Indeed let me correctly rephrase that out of 2 command only one I.e. configurePipeline is the one which has dependency and seems like browsePipeline is just url forwarding. 

* rectified that in my comment above, thank you 🙏

pulkitaggarwl: awesome. Have updated the readme for configure.

itowlson (approved):

PR #46, opened 2020-10-16
Typecasting resource to have a common contract

kanika1894:
Explicitly typecasting to have a comman contract between both the extensions.

### DIFF
--- a/src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
+++ b/src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
@@ -2,6 +2,7 @@ import * as vscode from 'vscode';
 import * as k8s from 'vscode-kubernetes-tools-api';
 import { IActionContext } from 'vscode-azureextensionui';
 import { resourceNode } from './models/resourceNode';
+import { GenericResource } from 'azure-arm-resource/lib/resource/models';
 const CONFIGURE_PIPELINE_COMMAND = 'configure-cicd-pipeline';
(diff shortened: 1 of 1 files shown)


src/commands/deployAzurePipeline/configureCicdPipeline/configurePipeline.ts
```diff
@@ -12,7 +13,7 @@ export async function configurePipeline(context: IActionContext, target: any): P
         if (cloudExplorer.available) {
             const clusterTarget = cloudExplorer.api.resolveCommandTarget(target);
             if (clusterTarget && clusterTarget.cloudName === "Azure" && clusterTarget.nodeType === "resource" && clusterTarget.cloudResource.nodeType === "cluster") {
-                const cluster = { resource: clusterTarget.cloudResource, subscriptionId: clusterTarget.cloudResource.subscription.subscriptionId };
```
itowlson, line 15: Rather than making a type assertion, it might be better to specify that `cluster` is of the contract type, and let the TypeScript compiler _verify_ that the RHS conforms to that side, e.g.

`const cluster: YourContractType = { resource: { id: ...,, type: ... }, subscriptionId: ... };`

This would, I believe, allow the compiler to check and report if things changed in a way that made the object no longer compatible with the contract.  By contrast, the assertions tell the compiler to ignore any potential incompatibilities and to treat the value of being the given type regardless.

kanika1894: sure, updated

itowlson (approved):

PR #51, opened 2020-11-23
ASO Feature integration.

Tatsinnit:
Thank you at first taking your time and looking into this :) Detail as to how logic flows etc, is mentioned below.

This PR contains whole E2E working logic ~~along with tree `Pin vs UnPin` logic as well.~~

Command gets run via the `cloudExplorer` and user can see the node cluster experience via the `clusterExplorer` 

* **Working of ASO**: detailed here: https://operatorhub.io/operator/azure-service-operator (Note: as a side note bunch of stuff I have update in their document and they will flow one MarkDom at openshift has merge his big changes)
   * `kubectl apply : cert-manager.yaml`
   * Check if the `cer-manager rollout` was successful.
   * `Install OLM using kubectl`, https://github.com/operator-framework/operator-lifecycle-manager/releases/
   * `Kubectl - apply - Issuer and Certificate cert-manager resources`
   * Ask user for the Service Principal `APPID and Password` this is something they need to provide.
   * `Kubectl apply : azureoperatorsettings Resource`: I formulate rest of the object needed for the azure operator settings resrouce for the operators namespace.
   * Finally check, `kubectl get pods -n operators` 

~~**TreeNode View**~~

* ~~For Tree node stuff intial code was very helpful but I have used the explicit function implementation so that I can add `.contextValue` as a class and attach the `Pin` or `Unpin` command according to the context values~~

2 screeshots below for idea:

image

image

image


src/commands/azureServiceOperators/pinned.ts
```diff
@@ -0,0 +1,23 @@
...
+    context: IActionContext,
+    target: any
+): Promise<void> {
+    const kubectl = await k8s.extension.kubectl.v1;
+
+    const clusterExplorer = await k8s.extension.clusterExplorer.v1;
+
+    if (clusterExplorer.available && kubectl.available) {
+        let pinnedItems = <any>vscode.workspace.getConfiguration().get('aso.pinned');
+        const itemPinned = target.impl.kind;
```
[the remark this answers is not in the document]
itowlson: Sorry for all the emphasis but I cannot, er, overemphasise this.  The pattern obviously feels complicated and unnecessary to you, but I couldn't see a way to make it any simpler while guaranteeing compatibility over time.  I implore you to use it even if it feels like "oh I found this super convenient shortcut and it worked."

src/commands/azureServiceOperators/ui/azureservicebrowser.ts
```diff
@@ -0,0 +1,117 @@
...
+        treeItem.contextValue = "aks.aso.allservices";
+        return treeItem;
+    }
+
+ }
+
+ class AzureServiceKindNode  implements k8s.ClusterExplorerV1.Node {
+    constructor(private readonly explorer: k8s.ClusterExplorerV1, private readonly kind: AzureServiceKind) {}
+    async getChildren(): Promise<k8s.ClusterExplorerV1.Node[]> {
+        const allFolderChildren = this.explorer.nodeSources.resourceFolder(this.kind.displayName, this.kind.displayName, this.kind.manifestKind, this.kind.abbreviation);
```
itowlson, line 51: This appears to be nesting a resource folder instead each custom folder.  Won't this give you an appearance of

```
|- context
   |- Azure Services
      |- EventHub
         |- EventHub
      |- SQLServer
         |- SQLServer
```

A resource folder _is_ a folder - not the _contents_ of a folder.

src/commands/azureServiceOperators/ui/azureservicebrowser.ts
```diff
@@ -0,0 +1,117 @@
...
+        treeItem.contextValue = "aks.aso.allservices.serviceitem";
+
+        return treeItem;
+    }
+}
+
+class PinnedAzureServiceKindNode  implements k8s.ClusterExplorerV1.Node {
+    constructor(private readonly explorer: k8s.ClusterExplorerV1, private readonly kind: AzureServiceKind) {}
+    async getChildren(): Promise<k8s.ClusterExplorerV1.Node[]> {
+        const allFolderChildren = this.explorer.nodeSources.resourceFolder(this.kind.displayName, this.kind.displayName, this.kind.manifestKind, this.kind.abbreviation);
```
itowlson, line 65: Again I'm confused by nesting the resource folder inside the custom folder.

src/commands/azureServiceOperators/ui/azureservicebrowser.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+export interface AzureServiceKind {
+    readonly displayName: string;
+    readonly manifestKind: string;
+    readonly abbreviation: string;
+}
+
+async function pinnedServiceKinds(): Promise<AzureServiceKind[]> {
+    const pinnedKindNames = <AzureServiceKind[]>vscode.workspace.getConfiguration().get('aso.pinned');
```
itowlson, line 83: I'm puzzled by this type assertion.  Your `package.json` says that `aso.pinned` is an array of strings, not an array of structured objects.

itowlson: I mean I am all in favour of being able to override the display name in the pinned folder - it just didn't look like the configuration schema supported it.

src/commands/azureServiceOperators/ui/azureservicebrowser.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    return pinnedKindNames;
+}
+
+// should be errorable but skip for prototype
```
itowlson, line 88: Something to address before merge.

src/commands/azureServiceOperators/ui/azureservicebrowser.ts
```diff
@@ -0,0 +1,117 @@
...
+function parseServiceResource(text: string): AzureServiceKind {
+    const bits = text.split(' ').filter((s) => s.length > 0);
+    if (bits.length === 4) {
+        // name, apigroup, namespaced, kind
+        return { displayName: bits[3], manifestKind: bits[3], abbreviation: bits[0] };
+    } else if (bits.length === 5) {
+        // name, shortnames, apigroup, namespaced, kind
+        return { displayName: bits[4], manifestKind: bits[4], abbreviation: bits[0] };
+    } else {
+        return { displayName: "WAT " + text, manifestKind: "WAT", abbreviation: "wat" };
```
itowlson, line 115: I would advise something more dignified than 'WAT'.

src/commands/azureServiceOperators/unpinned.ts
```diff
@@ -0,0 +1,24 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from 'vscode-azureextensionui';
+
+export default async function unpinned(
```
itowlson, line 5: Again, consider a verb rather than what sounds like a predicate.

src/commands/azureServiceOperators/unpinned.ts
```diff
@@ -0,0 +1,24 @@
...
+    target: any
+): Promise<void> {
+
+    const kubectl = await k8s.extension.kubectl.v1;
+
+    const clusterExplorer = await k8s.extension.clusterExplorer.v1;
+
+    if (clusterExplorer.available && kubectl.available) {
+        let unpinnedItems = <any>vscode.workspace.getConfiguration().get('aso.pinned');
+        const itemToUnpin = target.impl.kind;
```
itowlson, line 16: Again, _please_ don't do this.

itowlson (changes requested):
When this is all over I am going to buy you the beverage of your choice and bore you all evening with why the k8s extension API is designed the way it is.  Long story short, though, _it is this way for a reason_ - please don't try to bypass it unless you have established that it can't do what you need (in which case raise a bug report).

Other than that, mostly UI and tidying suggestions.  You're doing a great job on simplifying the install - fantastic to see!

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -0,0 +1,334 @@
...
+        return undefined;
+    }
+}
+
+export async function getAzureServicePrincipal(
+    target: AksClusterTreeItem
+): Promise<OperatorSettings | undefined> {
+    // Note: Will be removed soon.
+    // So what is happenign here?
+    // There is no way I could see we can generate ServicePricipal credential for the user.
```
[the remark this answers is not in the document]
Tatsinnit: 💡 Old comment indeed: essentially the `servicepriciple` here are the few gotchas: 

We decided to get user input for `AppID and Password`  

* This `sp create` is not actually create all the time, if the `service principle is existing` it does `update` and gives back new `password` or `clientsecret` everytime. 
* From Portal, when creating `service principle` the password generated comes with warning that `user should save this passwrod somewhere because this is the only time they are seeing this password`.
* Also, found out that for AZ CLI `2.13.1 and below` this create command is not working.

Hence it made more sense that instead of generating new password for the user all the time for the user, `User provide us` their AppID and their saved credentials.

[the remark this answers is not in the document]
Tatsinnit: 💡 📓  Note:

Another idea: (yet to reachout to AAD and some ASO folks to check on this)
* What I will again try and reachout is that if the AAD folks have any `sdk` and we can use it internally at vscode level to generate the unique `ServicePrinciple` and then we can do more organic, no user input required scenario if that is feasiable.

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -0,0 +1,287 @@
...
+            (f) => kubectl.api.invokeCommand(`create -f ${asoCrdYamlFile} --kubeconfig="${f}"`));
+
+        return runResult;
+    } catch (e) {
+        vscode.window.showErrorMessage(`Installing operator lifecycle manager CRD resource had following error: ${e}`);
+        return undefined;
+    }
+}
+
+export async function installOlm(
```
itowlson, line 106: Same comments

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -0,0 +1,287 @@
...
+            (f) => kubectl.api.invokeCommand(`create -f ${asoOlmYamlFile} --kubeconfig="${f}"`));
+
+        return runResult;
+    } catch (e) {
+        vscode.window.showErrorMessage(`Installing operator lifecycle manager resource had following error: ${e}`);
+        return undefined;
+    }
+}
+
+export async function installOperator(
```
itowlson, line 123: Again, same comments and same pattern.

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -0,0 +1,287 @@
...
+    clusterKubeConfig: string
+): Promise<k8s.KubectlV1.ShellResult | undefined> {
+    try {
+        const extensionPath = getExtensionPath();
+        const templateYAML = tmp.fileSync({ prefix: "aso-issuer", postfix: `.yaml` });
+        const yamlPathOnDisk = vscode.Uri.file(path.join(extensionPath!, 'resources', 'yaml', 'issuerandcertmanager.yaml'));
+        const issuerandcertmanager = fs.readFileSync(yamlPathOnDisk.fsPath, 'utf8');
+        fs.writeFileSync(templateYAML.name, issuerandcertmanager);
+
+        const runCommandResult = await tmpfile.withOptionalTempFile<k8s.KubectlV1.ShellResult | undefined>(
```
itowlson, line 151: Again I think the helper function from the previous functions could help you here - you have more setup to do but from here on it looks like the same structure of 'with temp file run a kubectl with --kubeconfig and report errors'.

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -0,0 +1,287 @@
...
+    kubectl: k8s.APIAvailable<any>,
+    operatorSettingInfo: OperatorSettings,
+    clusterKubeConfig: string
+): Promise<k8s.KubectlV1.ShellResult | undefined> {
+    try {
+        const extensionPath = getExtensionPath();
+        const yamlPathOnDisk = vscode.Uri.file(path.join(extensionPath!, 'resources', 'yaml', 'azureoperatorsettings.yaml'));
+
+        const azureoperatorsettings = fs.readFileSync(yamlPathOnDisk.fsPath, 'utf8');
+        azureoperatorsettings
```
itowlson, line 172: Bit confused by this. The `.replace` function returns a _new_ string, it does not modify the string on which it is called. So it looks like all these replacements are bring thrown away. And yet this clearly works when you test it.  What is going on?

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -0,0 +1,287 @@
...
+    const issuerCertResult = installationResponse.installIssuerCertResult;
+    const olmCrdResult = installationResponse.installOlmCrdResult;
+    const olmResult = installationResponse.installOlmResult;
+    const operatorResult = installationResponse.installOperatorSettingsResult;
+    const operatorSettingsResult = installationResponse.installOperatorResult;
+    const getOperatorsPodResult = installationResponse.getOperatorsPodResult;
+
+    let exitCode = 0;
+
+    // If total of outputCode is more than 0 then obviously something failed and display the html error, with std ouput.
```
itowlson, line 215: Be very careful! It is possible for an exit code to be interpreted as a negative number I believe... Consider instead creating an array of results and writing `allSucceeded = results.all(sr => !!sr && sr.code == 0)`.  It may feel less efficient but it is safer and more intention-revealing.

itowlson: Oh. your function only does anything if the input exit code is 0. So once you have a non-zero you discard all others anyway, so the comment is out of date.

I still prefer the `.all(... === 0)` though. *grin*

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -0,0 +1,287 @@
...
+    // Standard output code of all the installation response.
+    return installShellResult ? installShellResult.stderr + installShellResult.stdout : undefined
+}
+
+function installationExitCodeResponse(
+    exitCode: number,
+    installShellResult: k8s.KubectlV1.ShellResult | undefined
+): number {
+
+    if (exitCode === 0 && installShellResult) {
```
itowlson, line 257: This means that if `installShellResult` is undefined (which means we couldn't even run the program) you will leave the exit code at 0; that's probably not what you want.

itowlson: I tactfully note that the `all(...)` version of the process would not have this bug. *grin*

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -0,0 +1,287 @@
...
+        exitCode += installShellResult.code;
```
itowlson, line 258: `exitCode` must be 0 so this is actually `=` - the `+` will always be a no-op.

src/commands/azureServiceOperators/installAzureServiceOperator.ts
```diff
@@ -0,0 +1,158 @@
...
+            vscode.window.showInformationMessage('This command only applies to AKS clusters.');
+        }
+    }
+}
+
+async function install(
+    aksCluster: AksClusterTreeItem
+): Promise<void> {
+
+    const kubectl = await k8s.extension.kubectl.v1;
```
itowlson, line 49: You already get this in the calling function (though maybe you don't use it) - why not check it there so that everything is checked up top and the resolved, known-good API objects are passed down to here?

```
function installAzureServiceOperator(...) {
    const kubectlAPI = await k8s.extension.kubectl.v1;
    const cloudExplorerAPI = await k8s.extension.cloudExplorer.v1;
    const clusterExplorerAPI = await k8s.extension.clusterExplorer.v1;

    // check prerequisites
    if (!kubectlAPI.available) {
        // error and exit
    }
    if (!clusterExplorerAPI.available) {
        // error and exit
    }
    if (!cloudExplorerAPI.available) {
        // error and exit
    }

    const kubectl = kubectlAPI.api;
    const clusterExplorer = clusterExplorerAPI.api;
    const cloudExplorer = cloudExplorerAPI.api;

    // Now use kubectl, clusterExplorer and cloudExplorer freely

    install(aksCluster, kubectl);
}

function install(
  aksCluster: AksClusterTreeItem,
  kubectl: k8s.KubectlV1
) { ... }
```

itowlson (changes requested):
Few niggles and suggestions (mostly around code consolidation) but looking good!

tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "target": "es6",
         "outDir": "out",
         "lib": [
+            "es2017",
```
Tatsinnit, line 7: 💡 This is added to avoid the `Object.values` - `compiler error`, more here : ) https://stackoverflow.com/questions/42966362/how-to-use-object-values-with-typescript

resources/webviews/azureserviceoperator/azureserviceoperator.html
```diff
@@ -0,0 +1,121 @@
...
+            <a href="https://aka.ms/aks/aso-debug">Learn more</a></a>
+    </div>
+    {{#if getUserInput}}
+    <div class="container">
+        <div class="loader"></div>
+        <div class="panel panel-default data-container spform">
+            <i class="fa fa-wpforms" aria-hidden="true"
+                style="font-size:14px;padding-right:5px;padding-bottom:15px;"></i>
+            Please provide service principal appid and password, as detailed in <a
+                href="https://operatorhub.io/operator/azure-service-operator">step 4 of azure service operator docs</a>.
```
itowlson, line 65: capitalise:
```suggestion
                href="https://operatorhub.io/operator/azure-service-operator">step 4 of Azure Service Operator docs</a>.
```

itowlson (changes requested):
One teeny tiny proofreading change please, and then we can ship

itowlson (approved):
Yay!  Thanks for driving this through!

PR #58, opened 2021-02-22
Handle sceanrio when api-resource returns empty for all cluster without ASO installed.

Tatsinnit:
I found this last night when I was just playing around with this feature in extension, that `api-resources --api-group azure.microsoft.com --no-headers` always return `Empty` result, with `exit code == 0` when the ASO service is not installed. This can only be re-pro when you follow specific steps mentioned below.

**For example:** (I created new acs-cluster which has ASO not installed, and ran : kubectl for aspi-resource), this is what I get as result. Essentially when the resource is installed,  the kubectl will return something.

Screen Shot 2021-02-23 at 10 04 06 AM

**Specific steps to repro in Extension** 

* Create new aks-cluster and then set it as current context.
* Now close the extension and re-open it, what will happen is that the `registerNode` function will trigger and hit this line of code where is `return` of `api-resource` is empty it will fire a false warning message.

The easiest fix is that if the `kubectl` returns empty for the `api-resource` we do nothing. 

Thank you for taking time and reviewing This.

### DIFF
--- a/src/commands/azureServiceOperators/ui/azureservicebrowser.ts
+++ b/src/commands/azureServiceOperators/ui/azureservicebrowser.ts
@@ -1,10 +1,15 @@
 import * as k8s from 'vscode-kubernetes-tools-api';
 import * as vscode from 'vscode';
+import { Errorable } from '../../utils/errorable';
 
 interface AzureServiceKind {
     readonly displayName: string;
     readonly manifestKind: string;
     readonly abbreviation: string;
+}
+
+enum ASOInstallation {
+    ASONotInstalled = 'ASONotInstalled'
 }
 
 export async function AzureServiceBrowser(explorer: k8s.ClusterExplorerV1): Promise<k8s.ClusterExplorerV1.NodeContributor> {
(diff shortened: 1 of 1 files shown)


src/commands/azureServiceOperators/ui/azureservicebrowser.ts
```diff
@@ -16,21 +21,34 @@ export async function AzureServiceBrowser(explorer: k8s.ClusterExplorerV1): Prom
...
-        vscode.window.showWarningMessage(`Azure Service Operator api-resources command failed with following error: ${asoAPIResourceCommandResult?.stderr}.`);
-        return undefined;
-    }
 
-    const treeResourceItems = asoAPIResourceCommandResult.stdout.split("\n").map((line: string) => line.trim()).filter((line: string | any[]) => line.length > 0);
-
-    return treeResourceItems.map((item: string) => parseServiceResource(item)!);
+    if (!asoAPIResourceCommandResult) { // Fail to invoke command.
+        return { succeeded: false, error: `Azure Service Operator api-resources failed to invoke command.` };
+    } else if (asoAPIResourceCommandResult.code !== 0 && !asoAPIResourceCommandResult.stdout) { // Error result and Faild execution.
```
itowlson, line 44:
```suggestion
    } else if (asoAPIResourceCommandResult.code !== 0 && !asoAPIResourceCommandResult.stdout) { // kubectl errored and there is no partial result
```

Another way to approach this that might be clearer is to say: if there is _any_ stdout then (regardless of exit code) parse and return.  Then tackle the other cases:

```
if (execFailed) { return genericerr }
else if (thereIsStdout) { return success }
else if (code === 0) { return specialNotInstalledError }
else { return stderr }
```

This avoids the complex and slightly counterintuitive if-clause here.

src/commands/azureServiceOperators/ui/azureservicebrowser.ts
```diff
@@ -16,21 +21,34 @@ export async function AzureServiceBrowser(explorer: k8s.ClusterExplorerV1): Prom
...
+        return { succeeded: false, error: `Azure Service Operator api-resources command failed with following error: ${asoAPIResourceCommandResult?.stderr}.` };
+    } else if (asoAPIResourceCommandResult.code === 0 && !asoAPIResourceCommandResult.stdout) { // No ASO installed.
+        return { succeeded: false, error: ASOInstallation.ASONotInstalled };
+    } else { // success sceanrio.
+        const treeResourceItems = asoAPIResourceCommandResult.stdout.split("\n").map((line: string) => line.trim()).filter((line: string | any[]) => line.length > 0);
```
itowlson, line 49: Don't get how `line` would be anything other than a `string` in the `filter` clause; why do you need `| any[]` here?

src/commands/azureServiceOperators/ui/azureservicebrowser.ts
```diff
@@ -16,21 +21,34 @@ export async function AzureServiceBrowser(explorer: k8s.ClusterExplorerV1): Prom
 }
 
 async function allServiceKinds(): Promise<AzureServiceKind[] | undefined> {
+    const apiResult = await getAPIResourceCommandResult();
+
+    if (apiResult.succeeded) {
+        return apiResult.result;
+    } else if (apiResult.error !== ASOInstallation.ASONotInstalled) {
```
itowlson, line 28: Consider dropping the `else` here and inserting a blank line above.  "If it's not the special error then display it, and in any case return undefined" is your error handling "paragraph"; the two arms of the current `if` statement address different concerns (return value vs user interaction).

src/commands/azureServiceOperators/ui/azureservicebrowser.ts
```diff
@@ -16,21 +21,34 @@ export async function AzureServiceBrowser(explorer: k8s.ClusterExplorerV1): Prom
...
-    const treeResourceItems = asoAPIResourceCommandResult.stdout.split("\n").map((line: string) => line.trim()).filter((line: string | any[]) => line.length > 0);
-
-    return treeResourceItems.map((item: string) => parseServiceResource(item)!);
+    if (!asoAPIResourceCommandResult) { // Fail to invoke command.
+        return { succeeded: false, error: `Azure Service Operator api-resources failed to invoke command.` };
+    } else if (asoAPIResourceCommandResult.code !== 0 && !asoAPIResourceCommandResult.stdout) { // Error result and Faild execution.
+        return { succeeded: false, error: `Azure Service Operator api-resources command failed with following error: ${asoAPIResourceCommandResult?.stderr}.` };
+    } else if (asoAPIResourceCommandResult.code === 0 && !asoAPIResourceCommandResult.stdout) { // No ASO installed.
+        return { succeeded: false, error: ASOInstallation.ASONotInstalled };
+    } else { // success sceanrio.
```
itowlson, line 48:
```suggestion
    } else { // kubectl returned a list of resources (even if it errored part way through)
```

itowlson (approved):

PR #59, opened 2021-02-26
Updating the menu screenshot

yunjchoi:
When I kept the same file name, the file kept reverting back to the old one, so I changed the file name and the reference.


itowlson (approved):
Thanks for catching this!

Issue #54, opened 2021-02-02
Extension issue


sandipchajjed:
- Issue Type: `Bug`
- Extension Name: `vscode-aks-tools`
- Extension Version: `0.0.8`
- OS Version: `Darwin x64 20.3.0`
- VSCode version: `1.52.1`

:warning: We have written the needed data into your clipboard. Please paste! :warning:

sandipchajjed:
```json 
{
	"messages": [],
	"activationTimes": {
		"codeLoadingTime": 22377,
		"activateCallTime": 0,
		"activateResolvedTime": 34,
		"activationReason": {
			"startup": true,
			"extensionId": {
				"value": "azuredevspaces.azds",
				"_lower": "azuredevspaces.azds"
			},
			"activationEvent": "*"
		}
	},
	"runtimeErrors": []
}
```

itowlson:
@Tatsinnit We should look at webpacking the extension as we now (finally!) do for the core extension.  Are you interested in looking at that or shall I do it - I know you have a lot on your plate right now.

Issue #17, opened 2019-12-24
Extension causes high cpu load


welloncn:
- Issue Type: `Performance`
- Extension Name: `vscode-aks-tools`
- Extension Version: `0.0.4`
- OS Version: `Windows_NT x64 10.0.17134`
- VSCode version: `1.41.0`

:warning: Make sure to **attach** this file from your *home*-directory:
:warning:`C:\Users\le.wang\ms-kubernetes-tools.vscode-aks-tools-unresponsive.cpuprofile.txt`

Find more details here: https://github.com/microsoft/vscode/wiki/Explain-extension-causes-high-cpu-load

daniv-msft:
Thank you for reporting this issue!
Could you please attach the cpuprofile.txt file to this issue, so that the team in charge of this extension can investigate it?
Thanks!

rpus:
[ms-kubernetes-tools.vscode-aks-tools-unresponsive.cpuprofile.txt](https://github.com/Azure/vscode-aks-tools/files/4414873/ms-kubernetes-tools.vscode-aks-tools-unresponsive.cpuprofile.txt)

rpus:
(There's one from me, since I just got the modern equivalent of this issue!)

Version: 1.43.2
Commit: 0ba0ca52957102ca3527cf479571617f0de6ed50
Date: 2020-03-24T07:34:57.037Z
Electron: 7.1.11
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Darwin x64 19.4.0

itowlson:
Thanks.  We probably need to webpack the extension... _shudders in fear_

ezam-ws:
[ms-azure-devops.azure-pipelines-unresponsive.cpuprofile.txt](https://github.com/Azure/vscode-aks-tools/files/5233016/ms-azure-devops.azure-pipelines-unresponsive.cpuprofile.txt)

ezam-ws:
[ms-kubernetes-tools.vscode-aks-tools-unresponsive.cpuprofile.txt](https://github.com/Azure/vscode-aks-tools/files/5233022/ms-kubernetes-tools.vscode-aks-tools-unresponsive.cpuprofile.txt)

flakd:
Just got this 5min ago....

[...]

PR #62, opened 2021-03-15
Working webpack x other relevant changes.

Tatsinnit:
This PR aims for the enable webpack aks-vscode extension. With the deprecation work merged it was much quicker and simpler to get the Webpack. 

Currently there is one thing which I am exploring is the missing Key icon from `azureextensionui`:
* This probably will be due to `*.svc` inclusion in `Webpack` 

This work achieves:
* Webpacking the extension.
* Some related work like: Removal of the `replace` dependencies altogether because it was never needed and Webpack was not playing with it nicely either.

For double checking: I have done feature testing from the webpacked extension and it looks good feature wise. 

Should take care of these:  

* https://github.com/Azure/vscode-aks-tools/issues/17
* https://github.com/Azure/vscode-aks-tools/issues/54

Thank you. 🙏

cc: @itowlson 

Screen Shot 2021-03-16 at 11 33 07 AM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 out
+dist
 node_modules
 vsix
 .vscode-test/
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,6 +2,7 @@
 	// See http://go.microsoft.com/fwlink/?LinkId=827846
 	// for the documentation about the extensions.json format
 	"recommendations": [
-		"eg2.tslint"
+		"eg2.tslint",
+		"eamodio.tsl-problem-matcher"
 	]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,7 +16,7 @@
             "env": {
             },
             "outFiles": [
-                "${workspaceFolder}/out/**/*.js"
+                "${workspaceFolder}/dist/**/*.js"
             ],
             "preLaunchTask": "npm: watch"
         },
@@ -32,7 +32,7 @@
             "outFiles": [
                 "${workspaceFolder}/out/test/**/*.js"
             ],
-            "preLaunchTask": "npm: watch"
+            "preLaunchTask": "npm: test-compile"
         }
     ]
 }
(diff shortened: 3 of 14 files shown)


lstocchi:
Hey @Tatsinnit, i couldn't test it fully as i don't have access to an AKS cluster. Hopefully someone else can give you a better answer :crossed_fingers: 

What i noticed is that i don't see any icons in the tree
image

image

Based on the code, the tree is handled by a npm dependency, maybe the problem is there.
Can you try by adding `'vscode-azureextensionui': 'vscode-azureextensionui'` to the externals inside `webpack.config.js`?

If the key icon still doesn't work, can you please copy/paste the path where the extension tries to find it?

Tatsinnit:
Thank you @lstocchi :) That worked. 🙏😊 It took little around 5 second to load the tree node `Azure` but setting 'vscode-azureextensionui': 'vscode-azureextensionui' to the externals inside `webpack.config.js` did the trick. 

Screen Shot 2021-03-17 at 12 21 03 AM

Tatsinnit:
💡 🚢  For historic archival reasons I am adding all possible ways this could have worked but the issue we faced when we tried to `vsce package` the extension hence the preferred way is the way in this PR:

* Option 1: `external` for `azureextensionui` which worked in debug mode or developer mode but with package it continued to fail with package cannot be resolved when set as external.
* Option 2: **This is what we have done** we followed the `Webpack ~~> FileManagerPlugin` way as mentioned here as well: https://github.com/microsoft/vscode-azuretools/blob/main/dev/src/webpack/getDefaultWebpackConfig.ts#L118 and setting the `IgnoreBundle` flag to `false`. 

Thanks.

Tatsinnit:
💡`Vsce Package` works.

Below is just a documenting of an issue which I was able to see and the fix is discussed here: https://github.com/Azure/ms-rest-nodeauth/issues/83#issuecomment-606522618  

 I ran, `nom install` along with `npm run webpack` and then `vsce package` all went fine but when I use the new `VSIX` there is a known bug which is causing an issue: with `Error: Expected signal to be an instanceof AbortSignal`

* More TLDR version of this issue is here: https://github.com/Azure/ms-rest-nodeauth/issues/83#issuecomment-606500496 

What I\we was see:

Screen Shot 2021-08-02 at 5 25 01 PM

**After the fix** 

Screen Shot 2021-08-02 at 6 00 52 PM

PR #71, opened 2021-06-09
Configure to be disabled in untrusted workspaces.

Tatsinnit:
This work is to handle the untrusted workspace scenario.

Please refer to https://github.com/microsoft/vscode/issues/120251 for more details. Covers:  https://github.com/Azure/vscode-aks-tools/issues/68 

Thank you. cc: @itowlson

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,12 @@
     "engines": {
         "vscode": "^1.33.0"
     },
+    "capabilities": {
+        "untrustedWorkspaces": {
+            "supported": false,
+            "description": "This extension requires workspace trust."
+        }
+    },
     "license": "MIT",
     "categories": [
         "Azure"
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
     },
     "capabilities": {
         "untrustedWorkspaces": {
-            "supported": false,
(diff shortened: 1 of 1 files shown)


itowlson:
Do we know *why* it requires workspace trust?

Tatsinnit:
💡 Best example how I understood this was if foo@malicious user sending a zip file and I am opening that unzip project in vscode vs cloning the repo.

Since the default behaviour is that it will prompt,  Honestly I believe this change is **not needed** because by-default it doesn't trust the workspace behaviour: 

* It is upto user to decide : after VS Code prompts them to trust or distrust any newly opened workspace. With default behaviour that vscode will show the message that it is untrusted and ask user to select, I wonder why it is needed now.

Like without any changes recently I was asked this: (image below default behaviour)

Also, like this https://github.com/microsoft/vscode/issues/118584#issuecomment-802407888 

image

Thank you so much 🙏❤️

itowlson:
I've reviewed the current AKS code and I agree it's safe to run in untrusted workspaces.  We will need to keep this in mind as we carry out future changes - either be sure to keep things safe, or gate them with trust checks.

Issue #73, opened 2021-06-17
Migrate vscode module to latest.


Tatsinnit:
This workitem aims to upgrade our extension to latest `vscode` module  https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode 

<img width="1771" alt="Screen Shot 2021-06-18 at 10 42 45 AM" src="https://user-images.githubusercontent.com/6233295/122481010-0cf85980-d022-11eb-91ae-9d9977702744.png">

* https://code.visualstudio.com/updates/v1_36#_splitting-vscode-package-into-typesvscode-and-vscodetest 

This will have ripple effect to:

* current `monitor`, `storage` packages.

PR #74, opened 2021-06-17
vscode module updrage.

Tatsinnit:
This PR aims to upgrade `vscode` module to avoid consistent deprecation warning for vscode: more details in here. https://github.com/Azure/vscode-aks-tools/issues/73 

Thanks,

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,6 @@
     "packages": {
         "": {
             "version": "0.0.8",
-            "hasInstallScript": true,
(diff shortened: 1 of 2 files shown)


itowlson:
From experience on the k8s extension, it might be better to specify something a couple of versions back - we had a few folks report issues when we jumped to latest.  k8s extension requires 1.52, so maybe peg it to that unless you need a more recent feature?

itowlson (commented):
🎉

Issue #72, opened 2021-06-17
Scope work for zipping blob within vscode and decouple dependency on zip file from periscope tool.


Tatsinnit:
## Initial description

This work-item is created to scope what all work will entail to stock take work needed to write extension level `*.zip` for case when\if periscope tool removes `*.zip` from their upload to `azure storage`.

* How much upfront work will be needed to render the `azure storage` and use `jszip`.
* With initial investigation it will definitely involve writing the `zipper` function within the vscode.

Origination of this thought/idea and key issue: 

* Currently for the the `vscode x periscope` feature leverage the  `*.zip` file which was always there by design form the initial days of periscope tool
    * Now, given Periscope tool is cleaning up and as part of clean up holding the `*.zip` of something which already exist, is in the possible re-think.

Key challenge:

* Current way we can rely on the vscode default browser behaviour of download since we will handle the `zip` at vscode level how can we trigger default save behaviour?

## Update 20th June

Gained more clarity by implementing the functionality and there were challenges but more succinct version is below and I will tag a PR with this for the associated changes.

**Goal** To implement the `vscode` level download and zipping mechanism and to move away from the browser click mechanism. This enhancement decouples the underlying periscope implementation to calling mechanism only and no point of being dependent on a zip file which was populated earlier by the tool.

**Implementation**

* For the `click` mechanism I will utilise the `vscode -> webview` messaging call back to invoke 2 key aspects.
    * Call the `storage` to get the `node` logs, and stream read it and download  like here: https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-nodejs-legacy#download-a-blob 
    * Then from in-memory stream `zip` those files using `jszip` and store it ad the `downloadFolder` location

## Update 21 June - Draft POC PR

* Draft PR reside here for the implementation: https://github.com/Azure/vscode-aks-tools/pull/75 

## ⚠️ 💡  Implication of doing this in VsCode are following:

Here is what we found the implications/Poor experience are from the draft POC implementation here:  https://github.com/Azure/vscode-aks-tools/pull/75 

* Poor Download U/X experience: The current downside to move away from the `link to browser` will be gone, and replaced with only message driven approach. ( this is something which seems inevitable)
* Slower U/X experience: This will happen because the underlying code now deal with directly downloading the blob from storage and stream the download into the zip file and save it into the download folder all within vscode.
[...]

PR #75, opened 2021-06-21
Enable download blob and zip within vscode for persicope uploaded logs to storage.

Tatsinnit:
This PR caters need to handle the download and zip the dowloaded folder into the download folder location.

Good amount of detail here: https://github.com/Azure/vscode-aks-tools/issues/72 

Currently, the download is achieved by user clicking on hyperlink which redirect user to the browser based experienced, but in the new way the experience the user will never need to leave the vscode eco-system.

These changes also are needed because the periscope clean-up PR which will result in clearing this `*.zip` file because it is not formed correctly and not needed. [I've checked the `az-cli` implementation ](https://github.com/feiskyer/azure-cli-extensions/blob/3faf1c812d44c1087c053864b1a041b7437d7cd4/src/aks-preview/azext_aks_preview/custom.py#L1315 )and essentially the `az-cli` also don't use the `*.zip` from periscope either, so this implementation makes more aligned to other tools as well. 

This will also decouples the dependency on the `*.zip` link we generate which is uploaded by the periscope tool.

Along with various advantages we do have much better control over the zip structure and hierarchy.

## Update: 22nd June:

Please refer more as to how thought process for this work evolved: https://github.com/Azure/vscode-aks-tools/issues/72

Introducing these changes in vscode is not that straight forward as it sounded: some key complications:

* Poor Download U/X experience: The current downside to move away from the `link to browser` will be gone, and replaced with only message driven approach. ( this is something which seems inevitable)
* Slower U/X experience: This will happen because the underlying code now deal with directly downloading the blob from storage and stream the download into the zip file and save it into the download folder all within vscode.
* Maintenance of this code needs some guarantee and contract form the persicope tool that the blob folder structure will not change in future.
* Breaking change: As a user-experience and tool related breaking change the `vscode` changes need to go first before any tool related changes to remove `zip` upload is removed. (This will take time to re-group with team - and is it worth it to break the tool to vscode contract effort.)

### DIFF
--- a/src/commands/periscope/helpers/periscopehelper.ts
+++ b/src/commands/periscope/helpers/periscopehelper.ts
@@ -261,8 +261,9 @@ export async function downloadableLinks(
             // and extract *.zip files which are individual node logs.
             const latestBlobUploadedByPeriscope = blob.name.indexOf(listCurrentUploadedFolders.sort().reverse()[0]);
 
-            if (latestBlobUploadedByPeriscope !== -1 && blob.name.indexOf('.zip') !== 0 && blob.name.indexOf(nodeName) !== -1) {
+            if (latestBlobUploadedByPeriscope !== -1 && path.extname(blob.name) !== '.zip' && blob.name.indexOf(nodeName) !== -1) {
                 // Get a block blob client
+                // https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-nodejs-legacy#download-a-blob
                 const blockBlobClient = containerClient.getBlockBlobClient(blob.name);
 
                 // Get blob content from position 0 to the end
@@ -270,27 +271,37 @@ export async function downloadableLinks(
                 const downloadBlockBlobResponse = await blockBlobClient.download(0);
                 console.log('\nDownloaded blob content...');
                 const contentOfFile = await streamToString(downloadBlockBlobResponse.readableStreamBody);
-                console.log('\t', contentOfFile);
+                // console.log('\t', contentOfFile);
+                // The zip structure folder will be parent folder along with the log file.
+                // Example: /path/to/node-name/iptables/iptables.log ==> will get zipped as iptables/iptables.log
                 const filename = blob.name.replace(/^.*[\\\/]/, '');
-
-                zip.file(filename, `${contentOfFile}`);
+                const parentDir = path.dirname(blob.name).split(path.sep).pop();
+                zip.file(path.join(parentDir!, filename), `${contentOfFile}`);
             }
 
         }
-        const downloadsFolder = require('downloads-folder');
-        console.log('\nDownloads folder is...');
-        console.log(downloadsFolder());
-        const filename = `${downloadsFolder()}/${nodeName}.zip`;
-        zip.generateAsync({ type: "nodebuffer" }).then(function (content) {
-            fs.writeFileSync(filename, content);
-        });
+
+        // Write zip file to download location.
+        writeZipToDownloadLocation(nodeName, zip);
+
     } catch (e) {
         vscode.window.showErrorMessage(`Error generating downloadable link: ${e}`);
         return undefined;
     }
 }
 
+function writeZipToDownloadLocation(nodeName: string, zip: JSZip) {
+    const downloadsFolder = require('downloads-folder');
+    console.log('\nDownloads folder is...');
+    console.log(downloadsFolder());
+    const filename = `${downloadsFolder()}/${nodeName}.zip`;
+    zip.generateAsync({ type: "nodebuffer" }).then((content) => {
+        fs.writeFileSync(filename, content);
+    });
+}
+
 // A helper function used to read a Node.js readable stream into a string
+// https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-nodejs-legacy#download-a-blob
 async function streamToString(readableStream: any) {
     return new Promise((resolve, reject) => {
         const chunks: any[] = [];
--- a/src/commands/periscope/helpers/periscopehelper.ts
+++ b/src/commands/periscope/helpers/periscopehelper.ts
@@ -10,6 +10,7 @@ import * as path from 'path';
 import * as fs from 'fs';
 import AksClusterTreeItem from '../../../tree/aksClusterTreeItem';
 import * as axios from 'axios';
+import JSZip = require('jszip');
 const tmp = require('tmp');
 
 const {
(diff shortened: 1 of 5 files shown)


src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -214,6 +216,105 @@ export async function generateDownloadableLinks(
...
+        // Hide this all under single pupose funct which lik elike : getZipDir or get me dirname
+        const containerClient = blobServiceClient.getContainerClient(containerName);
+
+        // List all current blob.
+        const listCurrentUploadedFolders = [];
+        for await (const item of containerClient.listBlobsByHierarchy("/")) {
+            if (item.kind === "prefix") {
+                listCurrentUploadedFolders.push(item.name);
+            }
+        }
```
itowlson, line 252: You may be able to replace this loop with something like `const currentFolders = (await containerClient.list(...)).filter(item => item.kind === prefix).map(item => item.name)` but I am not sure if that enumeration is async - in that case filter and map may not be available.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -214,6 +216,105 @@ export async function generateDownloadableLinks(
...
+
+        // Sort and get the latest uploaded folder under the container used by periscope.
+        // const periscopeHtmlInterfaceList = [];
```
itowlson, line 255: Tidy commented-out code here and throughout.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -214,6 +216,105 @@ export async function generateDownloadableLinks(
...
+        // jszip for zipping
```
itowlson, line 256: Does this comment add anything?

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -214,6 +216,105 @@ export async function generateDownloadableLinks(
...
+        const zip = new JSZip();
+
+        for await (const blob of containerClient.listBlobsFlat()) {
+            // Get the latest uploaded folder, then Identify the Zip files in the latest uploaded logs within that folder
+            // and extract *.zip files which are individual node logs.
+            const latestBlobUploadedByPeriscope = blob.name.indexOf(listCurrentUploadedFolders.sort().reverse()[0]);
```
itowlson, line 262: This naming is confusing.  It clearly isn't a blob - it appears to be an index into a string.  Also it appears to be used only for whether the substring is present or not, not the index itself.  Perhaps consider:

```suggestion
            const latestFolder = listCurrentUploadedFolders.sort().reverse()[0];
            const blobMatchesLatestFolder= blob.name.includes(latestFolder);
```

_If_ I have understood this correctly, of course... big if!

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -214,6 +216,105 @@ export async function generateDownloadableLinks(
...
+
+            if (latestBlobUploadedByPeriscope !== -1 && path.extname(blob.name) !== '.zip' && blob.name.indexOf(nodeName) !== -1) {
```
itowlson, line 264: Again consider `includes` instead of `indexOf(...) === -1`.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -214,6 +216,105 @@ export async function generateDownloadableLinks(
...
+                // Get a block blob client
+                // https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-nodejs-legacy#download-a-blob
+                const blockBlobClient = containerClient.getBlockBlobClient(blob.name);
```
itowlson, line 267: Could everything from here to the bottom of the loop be a well-named helper function?  It seems like a `downloadBlobToFile` kind of chunk.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -214,6 +216,105 @@ export async function generateDownloadableLinks(
...
+
+                // Get blob content from position 0 to the end
+                // In Node.js, get downloaded data by accessing downloadBlockBlobResponse.readableStreamBody
+                const downloadBlockBlobResponse = await blockBlobClient.download(0);
+                console.log('\nDownloaded blob content...');
```
itowlson, line 272: Users won't see `console.log`.  Consider using a progress notification if this is going to take a while.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -214,6 +216,105 @@ export async function generateDownloadableLinks(
...
+
+    } catch (e) {
+        vscode.window.showErrorMessage(`Error generating downloadable link: ${e}`);
+        return undefined;
+    }
+}
+
+function writeZipToDownloadLocation(nodeName: string, zip: JSZip) {
+    const downloadsFolder = require('downloads-folder');
+    console.log('\nDownloads folder is...');
```
itowlson, line 295: Remove debugging code

src/commands/periscope/helpers/periscopehelper.ts
itowlson: Oh looking below maybe this isn't downloading the blob to a file.

src/commands/periscope/periscope.ts
```diff
@@ -150,7 +151,13 @@ async function createPeriscopeWebView(
 
                     panel.webview.html = getWebviewContent(clusterName, extensionPath, outputResult, periscopeStorageInfo, downloadableAndShareableNodeLogsList);
                 }
-
+                if (message.command === "downloadLink") {
+                    // Generate link mechanism is in place due to current behaviour of the aks-periscope tool. (which seems by design for now)
+                    // more detail here: https://github.com/Azure/aks-periscope/issues/30
+                    await longRunning(`Downloading ${message.text} logs in download location.`,
```
itowlson, line 157: What happens when they are downloaded?  Do you open the folder for the user?

itowlson (changes requested):
This is hard for me to review because I'm not sure what it's doing!  I've left a few comments where I don't understand the code in itself; otherwise I trust you and the PMs to test it and make sure it does the right thing and is understandable to the users!

davidkydd:
Did you consider simply dropping support for zipping the outputs in VSCode? This is a convenience function that could be left up to user to do afterwards if they wished - and as you mention az-cli doesn't make use of the .zip, so there isn't a conventional periscope experience to uphold here or anything

Tatsinnit:
> Did you consider simply dropping support for zipping the outputs in VSCode? This is a convenience function that could be left up to user to do afterwards if they wished - and as you mention az-cli doesn't make use of the .zip, so there isn't a conventional periscope experience to uphold here or anything

Yep, we considered all options but as vscode user-experience the current U/X experience is the best solution for vscode ecosystem. 

The `zip` provides end to end experience with quick download and user-scenario for this is mostly to quickly run the tool and share the zip with CSS folks for diagnostics.

The `az-cli` has its own different ecosystem which has its own usage, and since its pure command line they get away with bunch of text and asking user to download the `storage-explorer`.

If we drop support for `zip` it will break the whole end-to-end experience for vscode which we designed this feature for.

Tatsinnit:
💡 Thank you so much for discussion, this is no longer necessary, further discussion and whole update reside here: https://github.com/Azure/vscode-aks-tools/issues/72 , periscope tool kept the zip as part of their output contract.

Thanks 🙏

Issue #28, opened 2020-06-10
End-to-end- CI test integration investigation.


Tatsinnit:
This is a placeholder for thoughts\ideas to investigate mechanism to integrate test hooks for ask-tools build and that way we can counter any breaking changes pre-emptively and earlier then depending on manual tests:

[x] Possible approaches?
[x] Do we want to go functionalists level test and possibly think of UI level tests latter?

Thanks heaps,

PR #78, opened 2021-07-17
Test suite mechanism for aks extension.

Tatsinnit:
This PR is inspired by test suite which is present in Kubernetes extension.

Why Mocha: I find it easy to use, flexible and easily work with current project which we have.

Thanks in advance - cc: @itowlson, @squillace and @lstocchi if you can please eyeball and see if its cool to add as test suite and we can grow this once framework is in. ❤️☕️🙏

Motivation:

Key motivations are: 

* Quality checks in place: By adding tests will help in locating data contracts and functionality surety that with any future breaking change anything is broken we can easily identify them. 
* We will make these test as part of git workflow as well. 
* This work is also motivated form the testSuite work in Kubernetes extension.

Possibly also caters this workitem as well: https://github.com/Azure/vscode-aks-tools/issues/28 

**Latest**

Screen Shot 2021-12-06 at 10 16 03 PM

**Old**
Screen Shot 2021-07-17 at 1 00 08 PM

Thanks, 
^Tats


Tatsinnit:
💡 For archival reasons: Please note - Due to this existing bug - We cannot use `rooters` for nom run test` hence moving the `tests` folder inside the project makes more sense. 

* Bug\Issue is : https://github.com/microsoft/nodejstools/issues/1949

Thanks.

lstocchi (approved):
LGTM

- A suggestion is to give a look at `sinon`. I've used it in another project and i found it very helpful to write tests. Easier than simple mocha.
- I would update the gh action workflow to run tests in this PR as well.
- Not sure how interested you are in it but you can also think about adding a code coverage step

Tatsinnit:
💡 Thank you so much @lstocchi for above comment, they were very helpful. I have added 2 of 3 suggestions, SinonJS and also GH action to be part of smoke test as a workflow. ☕️🙏 Regarding code coverage, I would improve some test coverage before I add code coverage, initially this PR could be a good framework PR and small test of existing helper just as a sample. What do you think.

Please let me know if this looks good.

New GH action changes looks good: 

image

.github/workflows/build.yml
```diff
@@ -25,6 +25,12 @@ jobs:
       run: |
         npm install
         npm run compile
+    - name: Smoke test (Linux)
+      run: xvfb-run -a npm run test
```
lstocchi, line 29: Why did you use `xvfb-run` ? Couldn't you just execute `npm run test` in all platforms?

lstocchi (commented):
It looks great :rocket:

.github/workflows/build.yml
Tatsinnit: 💡☕️ Thank you @lstocchi,  It is interesting because the latest Ubuntu fails with `code ELIFECYCLE`, which I figured because reasons why `vscode-kubernetes` extension uses same code https://github.com/Azure/vscode-kubernetes-tools/blob/master/.github/workflows/build-dev.yml#L38 

* A sample of failed tests Ubuntu-Latest with `npm run test` [here](https://github.com/Azure/vscode-aks-tools/runs/4586936249?check_suite_focus=true). 
* Borrowed context which works: https://github.com/Azure/vscode-kubernetes-tools/blob/master/.github/workflows/build-dev.yml#L38

lstocchi: Thanks for trying @Tatsinnit ... not sure about it failing, i'll dig on it when i'll have some time. Good to merge for me. :+1:

Issue #80, opened 2021-08-10
AKS Periscope breaking change: Need to supply ContainerName to AKS Periscope from future release v0.5 onwards


qpetraroia:
AKS Periscope used to read the container name with a hardcoded Kubeconfig. This [PR](https://github.com/Azure/aks-periscope/pull/96) and this [issue](https://github.com/Azure/aks-periscope/issues/99) discuss the breaking changes that will occur to each service that has a working contract with AKS Periscope. The removal of the hardcoded code in Periscope will allow for other tools to onboard onto AKS Periscope easily.

Vscode-aks-tools is one service that has a working contract with AKS Periscope. To keep the default behavior as it is pass the API Server Name as the ContainerName in the AKS Periscope Deployment File.

CC: @Tatsinnit

PR #82, opened 2021-08-26
Periscope work for kustomise related changes.

Tatsinnit:
Following work in this PR takes care of workitem raised here: https://github.com/Azure/vscode-aks-tools/issues/80 

This work enable and supply `containerName` along with `sas` and `storageName`. This work  enables the tool to support via its own deployment yaml file rather than dependent on a centralised version as the periscope tool not contains changes which support ARC, SMI and OSM scenarios. 

Gyst: (background)

* To enable other kind of cluster periscope removed hardcoded `kubeconfig` file which was only ever used to get the `APIServerName` for formulating the `containerName`, as a result from latest image onwards `containerName` needs to be supplied form consuming tools. 
* Periscope image is now MCR onboarded as well.

Given, for formulating the storage link we were already utilising the `DNS Host Name` which is same as the `APIServerName` following change will enable, adaptation for new images.

YAML file now reside in consuming tools, because having persicope now uses kustomise and let consuming tool form their deployment file, which suites variety of teams like ARC, OSM, SMI and avoids issues we had where 1 change in persiceop repo will result in breaking change for others.

Thanks heaps. cc: @itowlson 🙏 

cc: @qpetraroia - fyi.

Working with latest image:

image


src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -95,7 +97,11 @@ export async function getStorageInfo(
         const storageAccKeyList = await storageClient.storageAccounts.listKeys(resourceGroupName, accountName);
         const storageKey = storageAccKeyList.keys?.find((it) => it.keyName === "key1")?.value!;
 
+        // Get container name from cluster-info default behaviour was APIServerName without
+        const containerName = await extractContainerName(clusterKubeConfig!);
```
itowlson, line 101: How do you know it's safe to apply `!` to `clusterKubeconfig` - there doesn't seem to be any validation.  If you're relying on the caller not passing `undefined` then the function signature should be `string` not `string | undefined`...?

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -95,7 +97,11 @@ export async function getStorageInfo(
...
+
         const clusterStorageInfo = {
+            containerName: containerName!,
```
itowlson, line 104: Again, you can't just `!` this away without validating.  If `extractContainerName` is guaranteed to return a string, _change the signature._  If it's not, _validate before asserting._

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -109,51 +115,67 @@ export async function getStorageInfo(
     }
 }
 
+async function extractContainerName(clusterKubeConfig: string): Promise<string | undefined> {
+        const kubectl = await k8s.extension.kubectl.v1;
+
+        if (!kubectl.available) {
+            vscode.window.showWarningMessage(`Kubectl is unavailable.`);
+            return;
```
itowlson, line 123: Recommend `return undefined;` to make it clear that the return value is significant.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -109,51 +115,67 @@ export async function getStorageInfo(
...
+        }
+
+        // Run cluster-info to get DNS Core hostname.
+        const runCommandResult = await tmpfile.withOptionalTempFile<k8s.KubectlV1.ShellResult | undefined>(
+            clusterKubeConfig, "YAML",
+            (f) => kubectl.api.invokeCommand(`cluster-info --kubeconfig="${f}"`));
```
itowlson, line 129: Formatting suggestion: if you format this as

```
const runCommandResult = await tmpfile.withOptionalTempFile(
  clusterKubeConfig, "YAML",
  (f) => invoke(...)
);  // This closing bracket is what has moved
```

then it gives the visual appearance of nesting a la braces, and helps the reader see the scope of the callback/with-function.

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -109,51 +115,67 @@ export async function getStorageInfo(
...
+
+        // Get DNS Core hostname which Periscope use it as name of the container.
+        const matches = runCommandResult?.stdout.match(/(https?:\/\/[^\s]+)/g);
+        let containerName: string;
+        if (matches![0].indexOf('://') !== -1) {
```
itowlson, line 134: Again, how are you validating this?  You specifically allow `matches` to be undefined through null propagation, but then skip that possibility here.

I would suggest that if `runCommandResult` is undefined, you should detect that before this stanza, and _return a meaningful error_ (or return undefined if there is no meaningful error).  Proceed with line 132 _only after_ you have validated `runCommandResult` to be a ShellResult (and, I would suggest, checking that it succeeded via the exitCode).

PR #88, opened 2021-10-31
Upgrade deprecated azure-storage package.

Tatsinnit:
This PR is continuation for the dependency fix, but this is final of it, reason being left over packages which are showing as dependency alert for this repository is actually just the decency of `@azure/` packages which are new storage packages. 

Following changes moves this code base 100 % away from the use of deprecated `azure-storage` package and use `@azure/storage-xxx` packages. more detail here: https://github.com/Azure/azure-storage-node or https://www.npmjs.com/package/azure-storage 

I have smoke tested this locally, adding PR for a quick look for more experienced eyes.

Thank you so much. ❤️🙏☕️

fyi: with new `storage` API they have removed the `Const` hence the `string` values are used like they have reopened in their sample API.

Screen Shot 2021-11-01 at 10 57 56 AM

### DIFF
--- a/src/commands/utils/azurestorage.ts
+++ b/src/commands/utils/azurestorage.ts
@@ -41,10 +41,9 @@ export function getSASKey(
 
     const permissionsForSas = sasPermission(linkDuration);
 
-    // Please refer to the new implementation of azure-storage api here: https://github.com/Azure/azure-storage-node 
-    // Constans used to be available externallbut not in new packages they dont.
-    // https://github.com/Azure/azure-storage-node/blob/master/lib/common/util/constants.js#L179
-    const cerds = new storage.StorageSharedKeyCredential(storageAccount, storageKey);
+    // The Azure storage client doesn't export constants.
+    // The ones we declare here are copied from https://github.com/Azure/azure-storage-node/blob/c4226315f037f2791f7c938e900b3497c9c0a67a/lib/common/util/constants.js#L179
+    const creds = new storage.StorageSharedKeyCredential(storageAccount, storageKey);
     const accountSharedAccessSignature = storage.generateAccountSASQueryParameters({
         expiresOn : expiryDate,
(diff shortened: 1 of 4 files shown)


src/commands/utils/azurestorage.ts
```diff
@@ -40,23 +39,23 @@ export function getSASKey(
...
-            Start: startDate,
-            Expiry: expiryDate
-        }
-    };
+    const permissionsForSas = sasPermission(linkDuration);
+
+    // Please refer to the new implementation of azure-storage api here: https://github.com/Azure/azure-storage-node 
+    // Constans used to be available externallbut not in new packages they dont.
+    // https://github.com/Azure/azure-storage-node/blob/master/lib/common/util/constants.js#L179
+    const cerds = new storage.StorageSharedKeyCredential(storageAccount, storageKey);
```
itowlson, line 47:
```suggestion
    const creds = new storage.StorageSharedKeyCredential(storageAccount, storageKey);
```

src/commands/utils/azurestorage.ts
```diff
@@ -40,23 +39,23 @@ export function getSASKey(
...
-                AccountSasConstants.Resources.CONTAINER +
-                AccountSasConstants.Resources.OBJECT,
-            Permissions: permissionsForSas,
-            Start: startDate,
-            Expiry: expiryDate
-        }
-    };
+    const permissionsForSas = sasPermission(linkDuration);
+
+    // Please refer to the new implementation of azure-storage api here: https://github.com/Azure/azure-storage-node
```
itowlson, line 44: Readers of the code aren't very interested in how storage used to work.  Focus on what is current e.g. "The Azure storage client doesn't export constants.  The ones we declare here are copied from <permalink>."  (Also use a GitHub permalink rather than a line in the master branch, because code in master will move around over time.)

itowlson (approved):
Couple of things that are not showstoppers but would be good to tidy up - otherwise looks good!

PR #89, opened 2021-11-02
Fix another damned packaging break

itowlson:
I'm also turning on package build on PRs which I somehow got wrong last time _sigh_


Tatsinnit (approved):
👍

PR #93, opened 2021-12-05
fix for the runtime node_module asynciterator error.

Tatsinnit:
This PR is open to cater a fix for development time only issue `asyniterator` error which is detailed more here: https://stackoverflow.com/questions/43694281/ts2318-cannot-find-global-type-asynciterableiterator-async-generator/43694282 

the behaviour appears after successful `nom install` and `nom run webpack` and when dev will open the code in `vscode` and hit `f5` its then we see some error like this: 

image

image

Thanks,
^Tats

cc: @itowlson, @squillace or @lstocchi - for quick reference and if it looks good.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,6002 @@
 {
     "name": "vscode-aks-tools",
-    "version": "0.0.10",
-    "lockfileVersion": 1,
+    "version": "0.0.11",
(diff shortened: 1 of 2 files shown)


itowlson (approved):
LGTM

PR #94, opened 2021-12-23
Add comment for cloud node dependency by other extension.

Tatsinnit:
This PR is outcome from a quick ideas for a very old wi #8 , this extension has grown since then and other azure extension is using this cloud node tree for their extension hook as well, adding comment for clarity.

cc: @itowlson ☕️❤️🙏 What do you think? Thank you so much in advance.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -32,6 +32,9 @@ export async function activate(context: vscode.ExtensionContext) {
 
         registerUIExtensionVariables(uiExtensionVariables);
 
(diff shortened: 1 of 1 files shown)


itowlson:
*cough* okay, this is a bit embarrassing but... I was going to say this was in the wrong place and it should be on the tree item declarations in `src/tree`, but when I went to check the file names... the files already contained suitable comments.

So you can close this.  Apologies for putting unnecessary work in your way.

Tatsinnit:
Thank you so much as always, : ) <3 @itowlson. 🎅

PR #99, opened 2022-02-24
Feature/configure aks starter workflow

Tatsinnit:
💡 This work is to enable a `new feature` for the `dev experienc`e for generation of a starter workflow for azure Kubernetes services workflow file which can be easily saved to respective projects for quick deployment.

Currently this work will serve user with a pre-baked copy of GitHub `starter workflow for AKS` and all user need to do is save and add left over secrets.

☕️ Please note: This work should not be merged until this PR becomes official: https://github.com/actions/starter-workflows/pull/1403

Thanks heaps, 🙏☕️❤️

cc: @gambtho , @rzhang628 , @peterbom 
FYi: @squillace for visibility.

Images below for more visibility of this feature.

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
         "onView:kubernetes.cloudExplorer",
         "onCommand:aks.detectorDiagnostics",
         "onCommand:aks.periscope",
-        "onCommand:aks.installAzureServiceOperator"
+        "onCommand:aks.installAzureServiceOperator",
+        "onCommand:aks.deployStarterWorkflow"
     ],
     "main": "./dist/extension",
     "contributes": {
@@ -63,6 +64,10 @@
             {
                 "command": "aks.installAzureServiceOperator",
                 "title": "Install Azure Service Operator"
+            },
+            {
+                "command": "aks.deployStarterWorkflow",
+                "title": "Configure AKS Starter Workflow"
             }
         ],
         "menus": {
@@ -109,6 +114,11 @@
                 {
                     "command": "aks.installAzureServiceOperator",
                     "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i"
+                },
+                {
+                    "command": "aks.deployStarterWorkflow",
+                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i"
+                
                 }
             ]
         }
--- a/src/commands/aksStarterWorkflow/deployStarterWorkflowHelper.ts
+++ b/src/commands/aksStarterWorkflow/deployStarterWorkflowHelper.ts
@@ -1,0 +1,19 @@
+import path = require("path");
+import * as fs from 'fs';
+import { getExtensionPath } from "../utils/host";
+import * as vscode from 'vscode';
+
+export function configureStarterConfigDataForAKS(
+    resourceName: string,
+    clusterName: string
+): string {
+    const extensionPath = getExtensionPath();
+
+    // Load vscode resoruce yaml file and replace content.
+    const yamlPathOnDisk = vscode.Uri.file(path.join(extensionPath!, 'resources', 'yaml', 'azure-kubernetes-service.yml'));
+    const starterFileAutoFillContent = fs.readFileSync(yamlPathOnDisk.fsPath, 'utf8')
+                                            .replace('<RESOURCE_GROUP>', resourceName)
+                                            .replace('<CLUSTER_NAME>', clusterName);
+
+    return starterFileAutoFillContent;
+}
--- a/README.md
+++ b/README.md
@@ -86,13 +86,17 @@ Right click on your AKS cluster and click on **Browse CI/CD Workflows** to view 
 
 ### Install Azure Service Operator
 
-Right click on your AKS cluster and click on **Install Azure Service Operator** to easily deploy the latest version of Azure Service Operator (ASO) on your AKS cluster and provision and connect applications to Azure resources within Kubernetes. When you select this option, you'll be prompted for a service principal for ASO to use when performing Azure resource operations. This service principal must have appropriate permissions (typically Contributor at suitable scope). Fill out the service principal details and click **Submit** to kick off the installation of Azure Servic Operator.
+Right click on your AKS cluster and click on **Install Azure Service Operator** to easily deploy the latest version of Azure Service Operator (ASO) on your AKS cluster and provision and connect applications to Azure resources within Kubernetes. When you select this option, you'll be prompted for a service principal for ASO to use when performing Azure resource operations. This service principal must have appropriate permissions (typically Contributor at suitable scope). Fill out the service principal details and click **Submit** to kick off the installation of Azure Service Operator.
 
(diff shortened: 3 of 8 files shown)


src/commands/aksStarterWorkflow/deployStarterWorkflowHelper.ts
```diff
@@ -0,0 +1,19 @@
...
+import * as vscode from 'vscode';
+
+export function configureStarterConfigDataForAKS(
+    resourceName: string,
+    clusterName: string
+): string {
+    const extensionPath = getExtensionPath();
+
+    // Load vscode resoruce yaml file and replace content.
+    const yamlPathOnDisk = vscode.Uri.file(path.join(extensionPath!, 'resources', 'yaml', 'azure-kubernetes-service.yml'));
```
OliverMKing, line 13: In the future it might be better to pull the workflow file from Github directly instead of storing a copy locally.

curl https://api.github.com/repos/tbarnes94/starter-workflows/contents/deployments/azure-kubernetes-service.yml | jq -r ".content" | base64 --decode

OliverMKing:
This looks really cool. I'm very excited about it!

OliverMKing (approved):

davefellows:
Love this, really great work Tats!

Is there a better name than "Configure AKS Starter Workflow"? At first, I didn't have a clue what a "Starter Workflow" was until I figured out it was in context of GitHub Actions. Could we call it "Configure GitHub Starter Workflow"? We're already in the context of an AKS clusters so "AKS" is redundant. "Configure GitHub Actions Starter Workflow" would be even more useful but feels way too long... 😄

Actually, how about just "Configure GitHub Actions"?? You click it and it gives you a Starter Workflow for your GH Action, seems pretty intuitive...

src/commands/aksStarterWorkflow/deployStarterWorkflowHelper.ts
Tatsinnit: ❤️🙏 Thank you so much for this comment, let me add some background design thoughts Ivan and I had for keeping local copies of files under resources for organisations:

* keeping a local copy under the extension will mean that in case you have no internet then this extension can still generate the files for template action. Also will avoid any file read failure in case other end point is unavailable.
* another advantage will be in case the online version changed and has breaking impact atleast the local copy of extension will keep working.

What do you think about above rationals, feel free to share thoughts please or I can ping you in teams sometime today. thank you for taking a look and really appreciate it your time for this. 🙏

OliverMKing: I appreciate the background and rationale. 

Being able to work for users with no internet makes lots of sense. A workaround might be favoring the pull method but still having a local copy in case the user is offline. An alternative approach could be to setup a Github action in this repo that would automatically create a PR whenever the upstream starter workflow changed.

This could be done with [this](https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md#spider-and-download-a-website) and would look like:

```md
name: Update starter workflow
on:
  schedule:
    - cron:  '0 10 * * *'
jobs:
  format:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: Download starter workflow
        run: |
          curl https://api.github.com/repos/tbarnes94/starter-workflows/contents/deployments/azure-kubernetes-service.yml | jq -r ".content" | base64 --decode > resources/yaml/azure-kubernetes-service.yml
      - name: Create Pull Request
        uses: peter-evans/create-pull-request@v3
        with:
          commit-message: update starter workflow
          title: Automated updates to starter workflow
```

This approach would also ensure that there would be no breaking changes since you'd be able to review any generated PRs. Although, either approach should be fine because in theory there should be no breaking impacts possible. The only breaking impact I can think of is the variable names changing.

Tatsinnit: Cool sounds good, let’s chant more about this and once we have scoped good possible ideas we can scout them in separate PR. Thank you so much for this, closing this for now. ❤️🙏

package.json
```diff
@@ -63,6 +64,10 @@
             {
                 "command": "aks.installAzureServiceOperator",
                 "title": "Install Azure Service Operator"
+            },
+            {
+                "command": "aks.deployStarterWorkflow",
+                "title": "Configure Github Actions"
             }
```
peterbom, line 71: Is the title correct? It looks inconsistent with the command

Tatsinnit: 🐈 💡 Cool, yeah, basically, title is the only the `"display purpose"` lik3e right click and `title` gets displayed, under the hood its just helping deploying theStarterWorkflow. let me know if you want to rename the internal classes we could do that. Could do `configureStarterWirkflow`. mostly will be asthetic chnage, will take 2 mins, let me make the change any ways :)

itowlson: From the PR description it sounds like this _creates_ a GH workflow that... does what, deploys the built project to the selected AKS cluster?  _If_ that's correct, and it may not be, then perhaps `Create GitHub Deployment Workflow` or `Create GitHub Starter Workflow`, though they're a bit long... agree with Peter that "Configure GitHub Actions" is just a bit too open-ended.

Tatsinnit: Done, have a look, just changed it to this, also redid the whole class name structure to reflect that : (so `deploy...` in name become `configure...`

right-click-menu

peterbom: I hadn't noticed Dave's earlier comment about 'Starter Workflow' not being obvious. 'Create GitHub Deployment Workflow' sounds good to me..

davefellows: +1 for  'Create GitHub Deployment Workflow'

Tatsinnit: +1, exactly I like with `create` and `Starter` it's explicit, shall we go with this one please : `Create GitHub Starter Workflow` it kind of contains everyone thoughts I guess.

What do you guys think? 

Agree

peterbom (approved):

resources/yaml/aks-periscope.yaml
```diff
@@ -174,7 +174,7 @@ spec:
             name: nodelogs-config
         - secretRef:
             name: azureblob-secret
-        image: mcr.microsoft.com/aks/periscope:v0.6
+        image: ghcr.io/tatsinnit/aks/periscope:test-image-tag-live-demo
```
peterbom, line 177: Should this be changed?

Tatsinnit:
Thank you so much to all in this review. 🙏❤️☕️ I think everything is catered so I guess I will go head and merge this sometime latter today and create a release for this feature for The work can get release by Monday our time. 

Cc:  @qike-ms, @rzhang628, @gambtho , @itowlson , @squillace , @davefellows ,  tor thoughts or plus one if this sounds like good idea. Thank you so much. 🙏☕️

qpetraroia:
@Tatsinnit +1 let's merge this!

gambtho:
thanks so much for the quick work @Tatsinnit !  this looks great

gambtho (approved):

gambtho (approved):

PR #102, opened 2022-03-10
retiring/deleting deploy to azure extension dependency.

Tatsinnit:
Opening this PR for removing commands for bridge which were dependent on [Deploy to Azure extension.](https://marketplace.visualstudio.com/items?itemName=ms-vscode-deploy-azure.azure-deploy) 

Deploy to azure extension is going out of service and as a result these command will not longer be supported as a bridge to the retiring extension.

Thank you so much ❤️☕️ cc and for review: @squillace , @bishal-pdMSFT , @bnookala , please let me know if I forgot anyone else.

* Here is the official deprecation notice here: https://github.com/Azure/github/issues/142 thank you 🙏 

After removal:

image


peterbom (approved):

rzhang628 (approved):
Approved!

PR #104, opened 2022-03-18
Some refactoring to ease creation of new commands

peterbom:
The goal here is to simplify and reduce the amount of code required to create new VS Code commands, particularly those involving clusters and WebViews.

It involves a collection of changes that move some common, repeated code from command definitions into shared locations. Namely:
- Getting (or not) an AksClusterTreeItem from a command target (the context from which a user launched the command)
- Performing the API calls to retrieve AppLens info for detectors (and sub-detectors)
- Creating a WebView
- Rendering a WebView (template+data -> output), registering helpers, getting VSCode resource URIs, and disentangling these operations in consumers

Along the way this led to some other adjustments:
- Flattening chains of nested method calls (e.g. WebView creation + rendering)
- More use of the `succeeded`/`failed` methods whose type predicates make errorable code flow really nicely
- More use of `Errorable` in general (in preference to union with `undefined`)
- Explicit handling of more error conditions
- Extra type declarations on methods
- Smaller try/catch blocks (so that the error message is more specific to the location the error is thrown)

One other rabbit hole I went down was this...

I like to handle user interaction (like displaying info/errors) at the top level, rather than guessing whether lower-level code is going to be showing notifications to the user. There were a couple of comments along the lines of "this method I'm calling reports errors so I don't have to".

Along with my enjoyment of using `Errorable`, this led me to play with moving those notifications up the call stack, so that top-level code can (and must) decide how it will relay errors to the users.

This had a cascade effect, touching more files than I would've liked, increased the code slightly, and introduced some cosmetic behavioural changes (since some lower-level code was raising notifications with different severities within the same function). I personally like the style it encourages, and think it's a worthwhile change, but others are welcome to disagree. For that reason, it's done in a couple of separate commits that can be axed if necessary.

(All the commits are pretty much self-contained, and can be reviewed individually or submitted as separate PRs if that's preferable.)


src/commands/utils/webviews.ts
```diff
@@ -0,0 +1,119 @@
+import * as fs from 'fs';
+import * as htmlhandlers from "handlebars";
+import { HelperDelegate } from 'handlebars';
+import * as path from 'path';
+import * as vscode from 'vscode';
+
+export enum HtmlHelper {
```
peterbom, line 7: If someone points out this is unnecessary, I'll agree. I was trying to match the existing behaviour in which different commands would register different sets of helpers. But since nothing is ever unregistered, with enough use all the helpers will get registered in the end anyway. It'd probably be clearer if we just registered them all at the start, and take the responsibility away from the commands entirely.

But I'd prefer to get input from others before going ahead and doing that.

src/commands/aksCRUDDiagnostics/aksCRUDDiagnostics.ts
```diff
@@ -1,103 +1,74 @@
...
-import path = require('path');
-import { htmlHandlerRegisterHelper } from '../utils/detectorhtmlhelpers';
+import { AppLensARMResponse, getDetectorInfo, getDetectorListData, getPortalUrl } from '../utils/detectors';
+import { failed } from '../utils/errorable';
 import AksClusterTreeItem from '../../tree/aksClusterTreeItem';
+import { createWebView, getRenderedContent, getResourceUri, HtmlHelper } from '../utils/webviews';
 
 export default async function aksCRUDDiagnostics(
-    context: IActionContext,
+    _context: IActionContext,
```
Tatsinnit, line 12: 💡 Thank you for this, can we please add more context for this change and benefits.

peterbom: It was something I changed to keep the linter in vscode happy (otherwise it complains that `context` is unused). This prompted me to go searching for an official style guide that says unused parameters should be prefixed by `_`.

[This](https://github.com/microsoft/TypeScript/issues/9458) is the closest thing to that I was able to find. It's apparently a little controversial (some people like underscore prefixes for parameters that _are_ used), and an alternative would be to disable the linter rule. I don't think the convention introduces any inconsistencies with code in this project, so I think it's fine here.

src/commands/utils/webviews.ts
peterbom: Discussed with @Tatsinnit and removed.

Tatsinnit (approved):
:shipit:💡 Thank you for this. Most of this is the Code move around in order to make the diagnostics detector stuff easy and any cross functionalities to be moved to common file. Thank you for detailed description in this PR. for future reference.

📓 As sanity check I will definitely would like at least, @itowlson, @lstocchi, or @squillace for any final eyeballing in case I missed anything. Also cc: @rzhang628 as well.

itowlson:
@Tatsinnit I no longer know this code well enough to provide any useful review - sorry!  But from a cursory glance this looks like a really valuable, pragmatic consolidation of the common code.  Nice one @peterbom!

Tatsinnit:
> @Tatsinnit I no longer know this code well enough to provide any useful review - sorry! But from a cursory glance this looks like a really valuable, pragmatic consolidation of the common code. Nice one @peterbom!

Thank you so much @itowlson this is good enough, just needed a quick eyeballing, will await for another few hours so that everyone take a quick glance and merge it. ❤️❤️🙏

PR #106, opened 2022-03-25
Integrating Best Practices detector.

Tatsinnit:
This is PR to show the behaviour form left over detectors which have some anomalous behaviour, we do have a proposed solution but need agreement as requirement from the PMs because some detectors from the list dont work and we dont want to add special code for anomalous behaviour.

Here is the working version:

Thank you, cc: @peterbom , @rzhang628 - here is PR for the idea.

image


src/commands/utils/detectors.ts
```diff
@@ -24,7 +24,12 @@ export async function getDetectorListData(
 
     let results: Errorable<AppLensARMResponse>[] = [];
     try {
-        results = await Promise.all(crudDetectorList.map((detector) => getDetectorInfo(cloudTarget, detector)));
+        const promiseResults = await Promise.all(crudDetectorList.map((detector) => getDetectorInfo(cloudTarget, detector)));
+        // TODO: Temporry code - Heavily commenting.
+        // This is temporarily put jsut to show that if we agree i.e. PM agree as requirement.
+        // We can safely swallow the detectors in the detectorId list which are not available.
+        // But handling Random API anamolous behaviour should not be handled form this VsCode.
+        results = promiseResults.filter((x) => x.succeeded !== false);
```
peterbom, line 32: Style question: do we always prefer explicit comparisons (like `!== false`)? I know that always doing that rules out bugs arising from JS's interpretation of truthiness. But when the property is a `boolean`, I'd find something like `promiseResults.filter((x) => x.succeeded)` easier to read. Or at least `x.succeeded === true` to avoid the double negative?

src/commands/aksBestPractices/aksBestPractices.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function getWebviewContent(
+  clusterdata: AppLensARMResponse,
+  detectorMap: Map<string, AppLensARMResponse>,
+  vscodeExtensionPath: string
+  ): string {
+    const webviewClusterData = clusterdata?.properties;
+    const styleUri = getResourceUri(vscodeExtensionPath, 'common', 'detector.css');
+    const templateUri = getResourceUri(vscodeExtensionPath, 'akscrud', 'aksCRUD.html');
```
peterbom, line 64: Does this use the exact same template as the CRUD detector?

src/commands/utils/detectors.ts
Tatsinnit: 100 % agree, Yeah, this is super POC for sharing with PM to see our theory. = )

src/commands/aksBestPractices/aksBestPractices.ts
Tatsinnit: 💡☕️ Indeed, Yeah, I think we finally have achieved a common html way to achieve displaying of data for end user. Famous last words until detector changes their data representation in any other format. Again, I have kept all this for POC for now, once we have agreement the solution will be formed between us. 🥷

rzhang628 (approved):
good to me, please move to forward.

Tatsinnit:
Nice, we are all in same page now, we will go ahead and implement this. thank you so much 🙏☕️❤️

peterbom (approved):

PR #114, opened 2022-04-08
Add workflow to update starter workflows

OliverMKing:
This workflow runs every 6 hours (or on user dispatch) and will open a pull request if the starter workflows in this repo don't match the official ones. If a pull request is closed, this workflow will not open a PR for these specific changes again.

This allows this repo to be aware of workflow changes but have complete control over how / when they are merged in.

The configureWorkflowHelper.ts changes take advantage of our starter workflow's env variables and makes it so you don't have to manually insert brackets to every workflow. Note that this change breaks this repo until the poll starter workflows workflow executes and the resulting PR is merged in. This lets you see the workflow in action.

Overall, this change lets this project stay up-to-date with starter workflow changes easily and non-intrusively. This is particularly important because security updates to the starter workflows should be merged in as soon as possible.

### DIFF
--- a/.github/workflows/poll-starter-workflows.yml
+++ b/.github/workflows/poll-starter-workflows.yml
@@ -21,7 +21,7 @@ jobs:
         id: download
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          workflow: poll-starter.yml
+          workflow: poll-starter-workflows.yml
           workflow_conclusion: success
           name: seen
           path: /tmp
--- a/.github/workflows/poll-starter-workflows.yml
+++ b/.github/workflows/poll-starter-workflows.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 
 env:
-  STARTER_WORKFLOW_PATH: "./resources/yaml"
+  STARTER_WORKFLOW_PATH: "./resources/yaml/"
   STARTER_WORKFLOW_REPO_BRANCH: "partner_templates"
 
 jobs:
--- a/.github/workflows/poll-starter-workflows.yml
+++ b/.github/workflows/poll-starter-workflows.yml
@@ -1,0 +1,80 @@
+name: Poll Starter Workflows
+
+on:
+  schedule:
+    - cron: "0 */6 * * *"
+  workflow_dispatch:
+
+env:
+  STARTER_WORKFLOW_PATH: "./resources/yaml"
+  STARTER_WORKFLOW_REPO_BRANCH: "partner_templates"
+
+jobs:
+  poll:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Download run information
+        uses: dawidd6/action-download-artifact@v2
+        continue-on-error: true
+        id: download
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          workflow: poll-starter.yml
+          workflow_conclusion: success
+          name: seen
+          path: /tmp
+
+      - name: Create new run information
+        if: steps.download.outcome != 'success'
+        run: |
+          echo "Creating new run information record"
+          declare -A seenWorkflows
+          touch /tmp/seenWorkflows
+          declare -p seenWorkflows > /tmp/seenWorkflows
+
+      - name: Download starter workflows
+        run: |
+          # get deployment workflows
+          echo "Calling GitHub API for starter workflows"
+          curl \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/actions/starter-workflows/contents/deployments?ref=$STARTER_WORKFLOW_REPO_BRANCH \
+            | jq -c '.[]' | while read -r workflow;
+          do
+            # extract fields
+            name=$(echo $workflow | jq '.name' -r)
+            downloadUrl=$(echo $workflow | jq '.download_url' -r)
+            sha=$(echo $workflow | jq '.sha' -r)
+
+            # get previous seen workflows (bash runs while loops in new shell)
+            source /tmp/seenWorkflows
+            
+            # download workflow if it's an AKS workflow and not seen before
+            echo "Checking workflow $name"
+            if [[ ( $name == azure-kubernetes-service* ) && ( ${seenWorkflows["$name"]} != "$sha" ) ]] ;
+            then
+              echo "Downloading starter workflow $name"
+              mkdir -p $STARTER_WORKFLOW_PATH && touch $STARTER_WORKFLOW_PATH$name
+              wget -O $STARTER_WORKFLOW_PATH$name $downloadUrl
+
+              echo "Writting to seen workflows"
+              seenWorkflows["$name"]="$sha"
+              declare -p seenWorkflows > /tmp/seenWorkflows
+            fi
+          done
+
+      - name: Upload run information
+        uses: actions/upload-artifact@v3
+        with:
+          name: seen
+          path: /tmp/seenWorkflows
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: update starter workflows
+          title: Automated Update Starter Workflows
+          body: This is an auto-generated PR. The starter workflow repo has changes not in this repository. Closing this PR stops all future alerts for these particular file changes.
+          branch: starter-workflow-updates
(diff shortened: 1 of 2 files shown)


OliverMKing:
@Tatsinnit

.github/workflows/poll-starter-workflows.yml
```diff
@@ -0,0 +1,80 @@
+name: Poll Starter Workflows
+
```
Tatsinnit, line 2: 💡 question: so every 6 hours there will be a commit in this repo for new workflows? 

I am keen to possibly understand the use case for it? And then we can follow a different approach for this solution! So whatever repo it is, the Master branch contains the true copy of the AKS workflows?

src/commands/utils/configureWorkflowHelper.ts
```diff
@@ -17,8 +17,8 @@ export function configureStarterConfigDataForAKS(
       // Load vscode resoruce yaml file and replace content.
     const yamlPathOnDisk = vscode.Uri.file(path.join(extensionPath.result, 'resources', 'yaml', `${workflowName}.yml`));
     const starterFileAutoFillContent = fs.readFileSync(yamlPathOnDisk.fsPath, 'utf8')
-                                            .replace('<RESOURCE_GROUP>', resourceName)
-                                            .replace('<CLUSTER_NAME>', clusterName);
+                                            .replace('your-resource-group', resourceName)
```
Tatsinnit, line 20: 💡 there will be no scenario where this could defer between env var or the deployment level use, if yes then I would recommend change the placeholder in current exiting deployment files.

Tatsinnit (commented):
🙏💡 thank you this PR, I have few questions and to understand better I have opened up with 2 small ones. Thank you.

.github/workflows/poll-starter-workflows.yml
OliverMKing: Every 6 hours this workflow executes which checks if there has been an update to the starter workflows housed in their main repository. If there's no new changes, no PR or commit is made. If there's new changes a PR is made suggesting those changes. If the PR is rejected, no PR is made until new changes are made to the starter workflows housed in their main repository.

OliverMKing: The use case is it informs us if starter repos have changed since we last saw them by making a PR if they have. This lets this repo keep up to date (with security vulnerability fixes and other changes).

src/commands/utils/configureWorkflowHelper.ts
OliverMKing: Correct! This should work.

.github/workflows/poll-starter-workflows.yml
Tatsinnit: 💡 Thanks for reply, Cool, so will be better to do this, I can add a PR with this, but doing this we can work on whatever the the latest for that repo.

``` *.ts
export async function workOnWhateverLatestProductionCopyis(): Promise<string | undefined> {
     const tempFile = tmp.fileSync({ prefix: "azure-kubernetes-service-", postfix: `.yaml` });

     try {
         const response = await axios.default.get('https://raw.githubusercontent.com/actions/starter-workflows/main/deployments/azure-kubernetes-service.yml');
         fs.writeFileSync(tempFile.name, response.data);
         replaceDeploymentResourceAndClusterName(some_object, tempFile.name);
         return tempFile.name;
     } catch (e) {
         vscode.window.showErrorMessage(`Azure-kubernetes-service Deployment file had following error: ${e}`);
         return undefined;
     }
 }
```

Tatsinnit: 💡 So the key question here is will the Yaml file ever contain something which is a breaking change and cause the vscode to fail in this case all I can think off that if `resoruce name` and `cluster name` (placeholder values) are removed then there is nothing to replace. (Thinking out loud I guess then also nothing will break) cc: @peterbom

.github/workflows/poll-starter-workflows.yml
```diff
@@ -0,0 +1,80 @@
...
+on:
+  schedule:
+    - cron: "0 */6 * * *"
+  workflow_dispatch:
+
+env:
+  STARTER_WORKFLOW_PATH: "./resources/yaml/"
+  STARTER_WORKFLOW_REPO_BRANCH: "partner_templates"
```
Tatsinnit, line 10: 💡 Just for archive reasons: What was the reason that there is a difference between `main` branch and `partner_templates` branch?  (I know that's a repo level question but I really want to avoid multiple vscode level commits which will clutter the `main` branch commit history against this repo.) Overhead of too many PRs.

.github/workflows/poll-starter-workflows.yml
OliverMKing: Partner_templates is what they ask us to raise PRs against. 

Pulling the workflows directly seems scary in case there ever is a breaking change. The workflow that I have created essentially does the same thing while giving this repository clear visibility into any changes.

In the case that resource name and cluster name placeholders are replaced it wouldn't technically break anything but it would be helpful to alert us that this change was made (which this workflow does).

Tatsinnit: Ah, okies, so breaking changes is something to watch out for, Thanks this is helping to understand the use-case and associated changes.

**Another Quick Scenario Question** - Lets say there was a change of a line in first 6 hours and then after that just before second polling someone added another small line of change. For with first poll there will be PR-1 with one line of change and then in second round there will be PR-2 with 2 lines of change (including the one in PR-1) so essentially the PR-1 is now redundant - in that scenario we will need to physically go and close the PR-1 and keep PR-2 as real PR?

peterbom:
Hey @OliverMKing - I really like this approach!

I gave it a try on my own repo (with one minor change that I'll point out in the yaml), and it worked like a charm.

[Here is the PR](https://github.com/peterbom/vscode-aks-tools/pull/5/files) it generated.

The contents of the PR raise a slight concern to me (I may not have been paying attention and missed the answer elsewhere). This might for @Tatsinnit and me to figure out, and not really related to this PR, but:

The resource group and cluster name are now expected to be env vars in GitHub, right? I.e. they're not hard-coded in the yaml. Which does make sense as a recommended approach. But the current behaviour of the VSCode extension is to replace those values with _actual_ resource group and cluster name in the yaml (based on the cluster the user right-clicked on).

I wonder if we need to get together and figure out how we really want the VSCode extension to behave here, so it's consistent with the starter templates...

.github/workflows/poll-starter-workflows.yml
```diff
@@ -0,0 +1,80 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Download run information
+        uses: dawidd6/action-download-artifact@v2
+        continue-on-error: true
+        id: download
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          workflow: poll-starter.yml
```
peterbom, line 24: Think this should be `poll-starter-workflows.yml`?

.github/workflows/poll-starter-workflows.yml
Tatsinnit: Nice, 👍 sounds good! Thank you 🙏 rest will be leftover yaml changes for placeholder value, all depends if you are comfortable to make those changes or we do it separately.

peterbom:
> 



> Hey @OliverMKing - I really like this approach!
> 
> I gave it a try on my own repo (with one minor change that I'll point out in the yaml), and it worked like a charm.
> 
> [Here is the PR](https://github.com/peterbom/vscode-aks-tools/pull/5/files) it generated.
> 
> The contents of the PR raise a slight concern to me (I may not have been paying attention and missed the answer elsewhere). This might for @Tatsinnit and me to figure out, and not really related to this PR, but:
> 
> The resource group and cluster name are now expected to be env vars in GitHub, right? I.e. they're not hard-coded in the yaml. Which does make sense as a recommended approach. But the current behaviour of the VSCode extension is to replace those values with _actual_ resource group and cluster name in the yaml (based on the cluster the user right-clicked on).
> 
> I wonder if we need to get together and figure out how we really want the VSCode extension to behave here, so it's consistent with the starter templates...

D'oh! I just figured out I was missing something. The env vars are defined in the yaml, so it all makes sense to me now :)

peterbom (approved):
LGTM - just the reference to `poll-starter.yml` that I _think_ needs changing to match the actual filename.

Tatsinnit (approved):
❤️🙏🎉 let’s rename the workflow file as mentioned and merge this so that we can manually run the workflow and create whatever new changes on workflow are.

.github/workflows/poll-starter-workflows.yml
OliverMKing: good catch! will fix

.github/workflows/poll-starter-workflows.yml
OliverMKing: partner_templates is where we are asked to make PRs against.

OliverMKing:
> LGTM - just the reference to `poll-starter.yml` that I _think_ needs changing to match the actual filename.

Fixed this 😄 . Good catch! 

Thanks for the feedback, everyone.

PR #116, opened 2022-04-14
Add tests to validate expected placeholders in starter workflows

peterbom:
The intent here is to reduce the manual burden of reviewing any automated PRs created as a result of the starter workflow templates changing.

Depends on #117 (tests should pass after this is merged).

Other things to note:
- To make it easier to test the template files themselves, the function to read the template and substitute the text was split into two.
- To avoid duplication of UI code, a common command handler is shared between all workflow commands (in one file: if there's a good reason to have one command per file I can split them back out).
- Minor change to bring focus to the generated document in the VS Code editor (previously it would show up behind other tabs and it wasn't always obvious the command worked).


Tatsinnit (commented):
💡 Just one observation that seems like the tests are failing because the `yaml` are removed. Might be add the `yaml` back.

Tatsinnit (approved):
❤️👍🎉

PR #119, opened 2022-04-15
Feature upgrade - Adapt ASO v2

Tatsinnit:
This PR is to update the new ASO (Azure Service Operator) release V2 mentioned here: https://azure.github.io/azure-service-operator/ 

Essentially, old ASO v1 is deprecated and latest update is published in their official repo.

Thanks heaps, cc: @peterbom & @rzhang628  for review please ❤️☕️🙏

### DIFF
--- a/src/commands/azureServiceOperators/helpers/azureservicehelper.ts
+++ b/src/commands/azureServiceOperators/helpers/azureservicehelper.ts
@@ -34,11 +34,6 @@ export async function startInstallation(
     );
     if (!isInstallationSuccessfull(webview, extensionPath, installationResponse.installCertManagerResult, installationResponse)) return undefined;
 
-    // installationResponse.installOlmResult = await longRunning(`Installing Operator Lifecycle Manager resource...`,
-    //     () => installOlm(kubectl, clusterKubeConfig.result)
-    // );
-    // if (!isInstallationSuccessfull(webview, extensionPath, installationResponse.installOlmResult, installationResponse)) return undefined;
-
     installationResponse.installOperatorResult = await longRunning(`Installing Opreator Namespace...`,
         () => installOperator(kubectl, clusterKubeConfig.result)
     );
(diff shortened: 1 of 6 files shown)


src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -64,47 +59,35 @@ async function getOperatorsPod(
...
-    const failureDescription = "Get operator pod had following error";
+    const command = `get pods -n azureserviceoperator-system`;
+    const failureDescription = "Get azureserviceoperator-system pod had following error";
 
     const result = await invokeKubectlCommand(kubectl, clusterKubeConfig, command, failureDescription);
     return result;
 }
 
-async function installOlmCrd(
+async function installCerManager(
```
peterbom, line 69: small typo - `installCertManager` (I only mention it because sometimes I search for keywords in the code to help understand how things work)

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
```diff
@@ -64,47 +59,35 @@ async function getOperatorsPod(
...
-    return result;
-}
-
 async function installOperator(
     kubectl: k8s.KubectlV1,
     clusterKubeConfig: string
 ): Promise<k8s.KubectlV1.ShellResult | undefined> {
-    const asoYamlFile = "https://operatorhub.io/install/azure-service-operator.yaml";
+    const asoYamlFile = "https://github.com/Azure/azure-service-operator/releases/download/v2.0.0-beta.0/azureserviceoperator_v2.0.0-beta.0.yaml";
     const command = `create -f ${asoYamlFile}`;
```
peterbom, line 91: I know this was already like this, but could we use `apply` instead of `create` here? When I run the command, I get lots of errors because resources already exist (I previously installed an older ASO version). I _think_ those failures would be avoided by using `apply`...

peterbom (approved):

src/commands/azureServiceOperators/helpers/azureservicehelper.ts
Tatsinnit: 💡 We used create for the `error` reason, I guess if `ASO` is already installed, we want user to know that you either installed or underlying stuff already installed so take a look. That also takes out unnecessary  handling of if installed half way etc sceanrios.

PR #120, opened 2022-04-16
Node health diagnostics integration.

Tatsinnit:
This PR enables the `Node-Health` detector diagnostics integration with extension.

cc: @peterbom and @rzhang628 - thanks in advance for review. ❤️🙏☕️

Screenshots:

Screen Shot 2022-04-16 at 8 17 59 AM

Screen Shot 2022-04-16 at 8 17 46 AM


resources/webviews/common/detector.html
```diff
@@ -30,6 +30,28 @@ <h2 class="content-header">{{name}}</h2>
         </div>
         <hr/>
             {{#eachProperty detectorData}}
+            {{#if (hasResourceThrottle value)}}
```
peterbom, line 33: Could we use an `ifEquals` helper here (like [this](https://stackoverflow.com/a/34252942))?

That way we won't need to keep adding additional helpers for similar string equality checks in future?

resources/webviews/common/detector.html
```diff
@@ -30,6 +30,28 @@ <h2 class="content-header">{{name}}</h2>
...
+                    <div class="panel-heading togglesubnet">
+                        <span class="fa status-icon fa-check-circle"></span>
+                        <h5 class="panel-title">{{value}}</h5>
+                    </div>
+                    <div class="throttle panel-body subnet">
+                        <table>
+                            <tbody>
+                                <tr>
+                                {{#each property}}
+                                    <td class="rectangle table-key-graph" {{{setStyleVar this.[2] }}}> </td>
```
peterbom, line 45: Is it possible to do this without the helper, like:
`<td class="rectangle table-key-graph" style="background-color: {{{this.[2]}}}"></td>`

I'm afraid that future detectors might have _other_ magic styling properties appearing in the data, then we start needing to add extra parameters to the helper, and it'll get really hard to understand.

By similar logic, I'm not sure about introducing the new `rectangle` css class here. There may be future rectangle-like fields with other dimensions. At the moment, these dimensions look very specific to this particular context.

What if we had something like this:
`<td class="table-key-graph" style="background-color: {{{this.[2]}}}; width: 20px; height: 50px;"></td>`
?

peterbom (approved):

resources/webviews/common/detector.html
Tatsinnit: 💡 Cool, I did tried the first approach but Handlebar don't allow, hence strictly driven by helper I guess.

I have made the changes with later recommendation which seems good, and just tested it as well. (will push it next.)

PR #121, opened 2022-04-16
Read changelog from right place.

Tatsinnit:
This PR fixes the empty `publish` field in release tag. 

The changelog was never getting read from changelog reader but from package reader hence it was empty. It is fixed now.

* Changes made according to this `underlying GH action` : https://github.com/mindsers/changelog-reader-action 

Thanks cc: @peterbom and @rzhang628 for review thanks guys! ❤️🙏☕️


peterbom (approved):
ohhhh, good job figuring that out!!

PR #127, opened 2022-05-05
Fix warning on Windows when loading clusters pane

peterbom:
This fixes a warning that appears when loading the cluster hierarchy pane on Windows: `unable to match a printer suitable for the output format "'custom-columns=Name:.metadata.name'"`.

Because of the single quotes, this entire expression was being interpreted as an output format, when in fact the output format should just be 'custom-columns'.

Using the alternative syntax `-o custom-columns=...` is equally valid and appears to work.


Tatsinnit (approved):
🥷👍 nice catch for windows! I see nothing of this error in Mac. Let’s merge this.

PR #128, opened 2022-05-06
Enable navigation to azure portal feature.

Tatsinnit:
This PR enables the navigation to specific cluster issue, now user can `right-click` and choose `Show in Azure Portal` for their cluster and browser will be called for the portal to open the `cluster overview` page. I know this will make @peterbom really happy, as he shared the real use-case.

* **Use case:** Imagine you have too many clusters and it gets really hard if you are logging into the azure portal and your recent history has lost your cluster only way you could search is via search resource or navigating inside AKS in Azure portal, this way simplifies all this into just one-click and navigate to the AKS cluster azure portal page.

Thanks you so much guys! cc: @peterbom,  @gambtho and @rzhang628 + @qpetraroia ❤️🙏

**Working Screenshots:**

image

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-aks-tools",
-    "version": "0.0.17",
+    "version": "0.0.18",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-aks-tools",
-            "version": "0.0.17",
+            "version": "0.0.18",
             "license": "MIT",
             "dependencies": {
                 "@azure/arm-containerservice": "^7.0.1",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
     "name": "vscode-aks-tools",
-    "version": "0.0.17",
+    "version": "0.0.18",
```
peterbom, line 3: Should this change have happened as part of another update?

peterbom (approved):

package-lock.json
Tatsinnit: 👍 you are correct, so when we create/write publish PR along with readme change et al.

PR #132, opened 2022-05-20
Show cluster property feature.

Tatsinnit:
This specific PR adds the feature enable the fetching of key property especially the `provisioningState` of cluster as long as information for `agentPool`. cc: @peterbom and @rzhang628 ❤️🙏☕️ Thanks

Screen Shot 2022-05-20 at 10 25 04 PM

Screen Shot 2022-05-22 at 2 10 17 PM

### DIFF
--- a/resources/webviews/aksclusterproperties/clusterproperties.html
+++ b/resources/webviews/aksclusterproperties/clusterproperties.html
@@ -43,8 +43,14 @@
                                     <td>Kubernetes Version</td>
                                     <td>{{clusterData.kubernetesVersion}}</td>
                                 </tr>
+                            </tbody>
+                        </table>
+                        <b>Agent Pool Properties</b>
+                        <table>
+                            <tbody>
                                 <tr>
-                                    <td><b>Agent Pool Properties</b></td>
+                                    <th>Properties</th>
+                                    <th>Values</th>
                                 </tr>
                                 {{#each clusterData.agentPoolProfiles}}
                                 <tr>
--- a/resources/webviews/aksclusterproperties/clusterproperties.html
+++ b/resources/webviews/aksclusterproperties/clusterproperties.html
@@ -41,8 +41,9 @@
                                 </tr>
                             </tbody>
                         </table>
-                        <b>Agent Pool Properties</b>
+
                         {{#each clusterData.agentPoolProfiles}}
+                        <b>Agent Pool: {{this.name}}</b>
                         <table>
                             <tbody>
                                 <tr>
--- a/resources/webviews/aksclusterproperties/clusterproperties.html
+++ b/resources/webviews/aksclusterproperties/clusterproperties.html
@@ -7,10 +7,6 @@
     <title>AKS Cluster Properties</title>
     <!-- Link to the css file -->
     <link rel="stylesheet" href="{{cssuri}}">
-    <script src="https://use.fontawesome.com/7dddd54b5c.js"></script>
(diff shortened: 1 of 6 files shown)


rzhang628 (approved):
good to go!

resources/webviews/aksclusterproperties/clusterproperties.html
```diff
@@ -0,0 +1,84 @@
...
+                        </table>
+                        <b>Agent Pool Properties</b>
+                        <table>
+                            <tbody>
+                                <tr>
+                                    <th>Properties</th>
+                                    <th>Values</th>
+                                </tr>
+                                {{#each clusterData.agentPoolProfiles}}
+                                <tr>
```
peterbom, line 56: I wonder if it'd be a good idea to move the `{{#each}}` block outside the `<table>` element. That way, if there's more than one agent pool you get a table for each one. The bold `Agent Pool Properties` could also contain {{this.name}}, so it's clear which is which. What do you reckon?

resources/webviews/aksclusterproperties/clusterproperties.html
```diff
@@ -0,0 +1,84 @@
...
+                                    <td>Node Version</td>
+                                    <td>{{this.nodeImageVersion}}</td>
+                                </tr>
```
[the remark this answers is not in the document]
Tatsinnit: +1 sounds good!! will add these 2 in next push.

resources/webviews/aksclusterproperties/clusterproperties.html
```diff
@@ -0,0 +1,84 @@
...
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>AKS Cluster Properties</title>
+    <!-- Link to the css file -->
+    <link rel="stylesheet" href="{{cssuri}}">
+    <script src="https://use.fontawesome.com/7dddd54b5c.js"></script>
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js"
```
peterbom, line 11: Can we remove these two JS libs (fontawesome and jquery)? I can't see them used -- am I missing something?

resources/webviews/aksclusterproperties/clusterproperties.html
```diff
@@ -0,0 +1,87 @@
...
+                                    <td>FQDN</td>
+                                    <td>{{clusterData.fqdn}}</td>
+                                </tr>
+                                <tr>
+                                    <td>Kubernetes Version</td>
+                                    <td>{{clusterData.kubernetesVersion}}</td>
+                                </tr>
+                            </tbody>
+                        </table>
+                        <b>Agent Pool Properties</b>
```
peterbom, line 44: How about we move this line inside the `{{#each}}` and put the name in it, like maybe:
```
<b>Agent Pool: {{this.name}}</b>
```
Then we get a nice heading + table for each agent pool (and we're not missing out their names, which I think are needed).

I was also trying to figure out if the preferred term is 'agent pool' or 'node pool', and I can't figure it out. The docs are a complete mix...

Tatsinnit: 👍 🥷 sounds good, let me add this, mostly the property name is the actual object name is `agentProolProfiles` so might be `agent pool` is the right one for now. let me add this change, I found one of your cluster with both linux and windows node, so will be a nice one to checkout.

peterbom (approved):

PR #136, opened 2022-05-25
Use remote Kustomize base for deploying Periscope

peterbom:
The latest Periscope release provides a Kustomize base file for deploying Periscope with Windows support.

These changes make use of Kustomize to remove the need for magic string replacement, and minimize the required changes to VS Code when new Periscope versions are released (we'll only have to update the release version and image tag, not the entire deployment yaml).

NOTE: the minimum supported `kubectl` version for Kustomize is [1.14](https://github.com/kubernetes/sig-release/blob/master/releases/release-1.14/release-notes-draft.md) (released March 2019).

This also moves this configuration into VSCode settings. This means users can override the default container registries, image tags, and Kustomize base repositories.


src/commands/periscope/periscope.ts
```diff
@@ -67,47 +70,57 @@ async function runAKSPeriscope(
...
+    overlayDir: string,
+    clusterKubeConfig: string
+): Promise<Errorable<k8s.KubectlV1.ShellResult>> {
+    return await tmpfile.withOptionalTempFile<Errorable<k8s.KubectlV1.ShellResult>>(clusterKubeConfig, "YAML", async kubeConfigFile => {
+        // Clean up running instance (without an error if it doesn't yet exist).
+        const deleteResult = await invokeKubectlCommand(kubectl, kubeConfigFile, 'delete ns aks-periscope --ignore-not-found=true');
+        if (failed(deleteResult)) return deleteResult;
+
+        // Deploy aks-periscope.
+        const applyResult = await invokeKubectlCommand(kubectl, kubeConfigFile, `apply -k ${overlayDir}`);
```
Tatsinnit, line 119: 💡- Keen to mention that we should make sure `-k` option don't have dependency on `kustomise` https://kustomize.io i..e currently we have min version of `kubectl` from parent extension and if for -k flag we need `customise` pre-baked from that route for host will not have this.

I thought I will share. Either way if that is the case we have other options handy for this to deliver but worth checking. Thanks heaps.

Tatsinnit (commented):
👍 Added the only key concern into this PR, rest looks cool, there are some space etc which we can do it off-line and I feel like worth double checking for it. 

Nice, and thank you for this. 🎉🙏

src/commands/periscope/periscope.ts
peterbom: Great point - I've noted the minimum `kubectl` version in the main PR description.

peterbom:
> 👍 Added the only key concern into this PR, rest looks cool, there are some space etc which we can do it off-line and I feel like worth double checking for it.
> 
> Nice, and thank you for this. 🎉🙏

Whitespace and error message fixed - thanks!!

Tatsinnit (approved):

PR #142, opened 2022-06-21
Update to latest k8s extension API

peterbom:
This upgrades the `vscode-kubernetes-tools-api` package reference from v1.0.0 to v1.3.0. The API has been extended in recent versions, including access to the configuration API, which we can use to retrieve the user's KUBECONFIG.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5135,9 +5135,9 @@
             }
         },
         "node_modules/vscode-kubernetes-tools-api": {
-            "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


peterbom:
> ☕️👍 thanks for this, just a note: is this correct? Might be missing the associated “package.json” file as it’s just “-lock” file.

That surprised me too! The `package.json` has a `^` in front of the version, which I think matches all minor versions from 0 upwards, so that's actually unchanged.

Issue #130, opened 2022-05-12
Fix or remove 'Azure Services' initialization code


peterbom:
When the extension is loaded, it executes some code which initializes the 'Azure Services' treeview item:
<img width="98" alt="image" src="https://user-images.githubusercontent.com/1817696/168179488-cacb23ef-ba6c-4805-b069-fdff0894569e.png">

There are some things that need looking at with this:
- It is not documented, AFAIK -- @Tatsinnit - am I wrong about that?
- It doesn't appear to work, at least for me (as shown in the screenshot, there is nothing inside when you expand it).
- It can slow down extension loading a *lot* if the active cluster has a lot of CRDs.

I'm not sure, but I *think* the intended functionality behind it is already covered by the 'Custom Resources' item (the difference being that 'Azure Services' is filtered so that it only includes Azure/ASO custom resources).

Things to consider:
- Is my understanding correct?
- If working, would the 'Azure Services' list provide value to users beyond that of the 'Custom Resources' item?
- If we really want that functionality, could we retrieve it from the Kubernetes extension, rather than repeating the same, expensive `kubectl get crd` operations?
- If users can survive without this, can we remove the code and vastly decrease initialization time?

itowlson:
Background:

This was created partly to provide a list of specifically Azure services, so the user could find them without grovelling through the full list of CRDs, but also as a place where people could pin the CRDs they regularly worked with, because, even filtered, the set of Azure CRDs is unmanageably huge!

This is not to say it can’t be removed or changed!  Just providing context on the reason it was done back in the day.

peterbom:
Thanks @itowlson! @Tatsinnit also gave me a bit of history and I now have a much better understanding of why it's there.

I'm wondering if it might be possible for us to create a `NodeContributor` that wraps and filters the one that's registered for `Custom Resources`, rather than duplicating it (and all the long-running retrieval logic) in the AKS extension.

Any initial thoughts whether it might be worth pursuing that idea?

itowlson:
That sounds like a good idea - loading a particular set of CRDs and representing them in a more relevant way is definitely a key use case for the API, and it would be great to have additional support for this.

peterbom:
I've played around with a couple of ideas here, with pretty limited success. The API [provides a couple of ways](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/master/docs/extending/clusterexplorernodes.md) of initializing a folder:
[...]

PR #143, opened 2022-06-22
Move to NodeContributor implementation for ASO CRDs

peterbom:
As discussed in #130 we need to fix the initialization code for the 'Azure Services' resource grouping folder in Cluster Explorer.

The previous code had a couple of issues:
- it didn't populate the resource folder properties (name, kind, abbreviation) correctly, so resource folders were not created.
- it tried to determine all the CRDs at initialization time, rather than when the grouping folder was expanded, meaning the cluster explorer took much longer to load when there were a lot of CRDs.

These changes use the CRD loading logic from the Kubernetes extension to correctly describe the resources, and change the loading behaviour to use a `NodeContributor` implementation which only lists the CRDs when required.

@Tatsinnit: I moved the implementation outside of the ASO command folder. Although this functionality refers to resource types that _can be_ created by the ASO command, it's not part of the command itself. I placed it under `tree` because that seemed the most appropriate place that already exists, but feel free to suggest otherwise.

cc: @itowlson (for sanity checking of Kubernetes extension API usage)


Tatsinnit (approved):
+1 looks good to me, I hope you have tested this functionally, Thanks.

src/tree/azureResourceNodeContributor.ts
```diff
@@ -0,0 +1,97 @@
...
+        if (!crdTypes.succeeded) {
+            vscode.window.showErrorMessage(`Unable to retrieve Azure Services resources: ${crdTypes.error}.`);
+            return [];
+        }
+
+        // Map the custom resource definitions into NodeSources that the K8s extension knows how to turn into nodes.
+        const nodeSources = crdTypes.result.map(t => this.explorer.nodeSources.resourceFolder(t.displayName, t.pluralDisplayName, t.manifestKind, t.abbreviation));
+
+        // Turn the NodeSources into nodes.
+        const nodes = (await Promise.all(nodeSources.map(ns => ns.nodes()))).flat();
```
itowlson, line 41: I am not sure programmatically getting the nodes from a `NodeContributor` and using them elsewhere works - I recall having this same idea and one point and finding some odd behaviour - but I can't substantiate that.  I am probably hallucinating (or was doing something else, like a fully custom contributor, or it was an earlier version of the API where these worked a bit differently).  But I'd encourage particularly thorough testing just in case.

itowlson (commented):
Looks fine to me.  There is one area where you use a technique that I _vaguely, vaguely_ recall having problems with in the past.  But my memory of it is completely faded so I might be imagining it or mistaking it for something else.  So I'd give the created nodes more than the usual level of testing, but if they hold up, then write it off as Old Man Yells At Clouds and ship away!

src/tree/azureResourceNodeContributor.ts
peterbom: Thanks @itowlson - @Tatsinnit and I have been testing and it appears to hold up.

One thing we did notice is that the last two options (watch and stop watching) are missing:

image

I'm not sure if that's anything at all to be concerned about, but thought I'd call it out in case it's related to the odd behaviour you were mentioning. My uninformed guess is it relates to the API version of the cluster explorer (v1 vs. v1.1 or something?) that we're using.

As it seems to function okay, we'll merge this for now, but please shout if anything sounds any alarm bells and I'll dig into it some more.

itowlson: Watch-related commands require a specific context string on the tree item, which is added here:

https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/92582a57ce722e166d25482b443c37a58d16305a/src/components/clusterexplorer/explorer.ts#L118

Not sure why this wouldn't appear on the 'duplicated' nodes if it appears on the 'ordinary' ones.  But that's the line where I'd start looking.

itowlson: Forgot to say, I'm delighted this is working for you and thanks for all the feedback!

peterbom: Ahh, I think that makes sense. It's missing the `apiName` property because the v1 `resourceFolder` function doesn't take that as the last argument: [v1](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/92582a57ce722e166d25482b443c37a58d16305a/src/api/contract/cluster-explorer/v1.ts#L104) vs. [v1_1](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/92582a57ce722e166d25482b443c37a58d16305a/src/api/contract/cluster-explorer/v1_1.ts#L106).

Thank you for the hint, @itowlson!

Any thoughts about whether we should be adopting the newer API, @Tatsinnit? I have no idea what's involved in doing that...

itowlson: The 1.1 API should be fully backward compatible except for using the `V1_1` namespace, i.e. you shouldn't need to change anything except for where you're passing that new optional argument.  They both back onto identical implementation code so I'd be confident there would be no migration issues.

PR #144, opened 2022-07-12
Update deprecated vscode-extensionui package.

Tatsinnit:
Hiya All, This PR intends to cater the update of the the deprecated `vscode-extensionui` package. https://www.npmjs.com/package/vscode-azureextensionui 

Thank you so much for your time and eyeballing, 

* Please note: I have tested locally and all functionalities are retained I will involve internally few folks to kindly test it as well.

Further there is a a **know issue** past web packing for the `core-tracing` which is well documented here: https://github.com/Azure/azure-sdk-for-js/issues/21745 hence the tsconfig changes.

Thanks heaps, cc: @rzhang628 , @gambtho , @peterbom , @itowlson , @philliphoff  and @squillace ❤️🙏☕️

### DIFF
--- a/src/azure-api-utils.ts
+++ b/src/azure-api-utils.ts
@@ -1,5 +1,3 @@
-import { Subscription } from '@azure/arm-subscriptions';
-import { ISubscriptionContext } from '@microsoft/vscode-azext-utils';
 
 export interface PartialList<T> extends Array<T> {
     nextLink?: string;
@@ -23,12 +21,4 @@ export function parseResource(armId: string): { resourceGroupName: string | unde
 function bitAfter(bits: string[], after: string): string | undefined {
     const afterIndex = bits.indexOf(after);
     return bits[afterIndex + 1];
-}
-
-export function toSubscription(context: ISubscriptionContext): Subscription {
-    return {
-        id: context.subscriptionPath,
-        subscriptionId: context.subscriptionId,
-        displayName: context.subscriptionDisplayName
-    };
 }
--- a/src/azure-api-utils.ts
+++ b/src/azure-api-utils.ts
@@ -1,5 +1,5 @@
-import { ISubscriptionContext } from "vscode-azureextensionui";
-import { SubscriptionModels } from '@azure/arm-subscriptions';
+import { Subscription } from '@azure/arm-subscriptions';
+import { ISubscriptionContext } from '@microsoft/vscode-azext-utils';
 
 export interface PartialList<T> extends Array<T> {
     nextLink?: string;
@@ -25,7 +25,7 @@ function bitAfter(bits: string[], after: string): string | undefined {
     return bits[afterIndex + 1];
 }
 
-export function toSubscription(context: ISubscriptionContext): SubscriptionModels.Subscription {
+export function toSubscription(context: ISubscriptionContext): Subscription {
     return {
         id: context.subscriptionPath,
         subscriptionId: context.subscriptionId,
--- a/src/tree/aksClusterTreeItem.ts
+++ b/src/tree/aksClusterTreeItem.ts
@@ -1,9 +1,9 @@
-import { AzExtParentTreeItem, AzureTreeItem, ISubscriptionContext } from "vscode-azureextensionui";
+import { AzExtParentTreeItem, AzExtTreeItem , ISubscriptionContext } from "@microsoft/vscode-azext-utils";
 import { CloudExplorerV1 } from "vscode-kubernetes-tools-api";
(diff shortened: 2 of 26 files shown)


src/azure-api-utils.ts
```diff
@@ -25,7 +25,7 @@ function bitAfter(bits: string[], after: string): string | undefined {
     return bits[afterIndex + 1];
 }
 
-export function toSubscription(context: ISubscriptionContext): SubscriptionModels.Subscription {
+export function toSubscription(context: ISubscriptionContext): Subscription {
```
peterbom, line 28: I see that some uses of this are now removed. Is it still used anywhere? Wondering if it can be removed...

src/tree/aksClusterTreeItem.ts
```diff
@@ -1,9 +1,9 @@
-import { AzExtParentTreeItem, AzureTreeItem, ISubscriptionContext } from "vscode-azureextensionui";
+import { AzExtParentTreeItem, AzExtTreeItem , ISubscriptionContext } from "@microsoft/vscode-azext-utils";
 import { CloudExplorerV1 } from "vscode-kubernetes-tools-api";
-import { SubscriptionModels } from '@azure/arm-subscriptions';
-import { toSubscription } from "../azure-api-utils";
+import { Subscription } from '@azure/arm-subscriptions';
 import { Resource } from "@azure/arm-resources/esm/models";
 import { assetUri } from "../assets";
+// import { toSubscription } from "../azure-api-utils";
```
peterbom, line 6: nit: clean up commented-out code

src/tree/aksClusterTreeItem.ts
```diff
@@ -44,12 +44,12 @@ export default class AksClusterTreeItem extends AzureTreeItem implements AksClus
...
 
     public get session(): ISubscriptionContext {
-        return this.root;
+        return this.session;
     }
 
-    public get subscription(): SubscriptionModels.Subscription {
-        return toSubscription(this.root);
-    }
+    // public get subscription(): ISubscriptionContext {
```
peterbom, line 50: cleanup

src/tree/subscriptionTreeItem.ts
```diff
@@ -27,23 +28,23 @@ export default class SubscriptionTreeItem extends SubscriptionTreeItemBase imple
     }
 
     public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
-        const client = new ResourceManagementClient(this.root.credentials, this.root.subscriptionId);
+        const client = new ResourceManagementClient(this.subscription.credentials!, this.subscription.subscriptionId!);
```
peterbom, line 31: I wonder why credentials can now be nullable/undefined? Can we be certain it's populated at this point? Do the docs give any guidance about under what circumstances it might not be set, and what to do about it if that's the case?

Just wondering if there's some re-authenticating dance we may need to do.

src/tree/subscriptionTreeItem.ts
```diff
@@ -27,23 +28,23 @@ export default class SubscriptionTreeItem extends SubscriptionTreeItemBase imple
...
 
     public get session(): ISubscriptionContext {
-        return this.root;
+        return this.session;
     }
 
-    public get subscription(): SubscriptionModels.Subscription {
-        return toSubscription(this.root);
-    }
+    // public get subscription(): ISubscriptionContext {
```
peterbom, line 45: cleanup

src/tree/subscriptionTreeItem.ts
Tatsinnit: Nice, I think I was playing around kept `!` by mistake, changed, and just tested it as well. Thank you for noticing this.

src/tree/aksClusterTreeItem.ts
```diff
@@ -12,10 +12,10 @@ export interface AksClusterTreeNode {
     readonly armId: string;
     readonly name: string;
     readonly session: ISubscriptionContext;
-    readonly subscription: SubscriptionModels.Subscription;
+    readonly subscription: Subscription;
```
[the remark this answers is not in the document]
Tatsinnit: sounds good, let me change it for the `clusterProperties` stuff as well. there is a utility function `parseResrouce` which can come handy and now target will have `resrouceGroupname` property available. just test and the cluster properties worked.

peterbom (approved):
Love it! Definitely makes more sense to me now we don't have `target.root` all over the place.

rzhang628 (approved):
Tested functions within the extension (diagnostic, periscope, show properties, github workflow, create cluster, etc.) all work well to me.

PR #145, opened 2022-07-12
AKS managed service start and stop cluster feature

Tatsinnit:
This feature enables the ease of use and menu item to enable `AKS Managed Services` operations like `start and stop`.

Some of the screen flow which will look like is as follows:

Thanks heaps in advance, ❤️🙏☕️ cc: @rzhang628 , @peterbom and @gambtho 

**New Screen flow**

Screen Shot 2022-07-28 at 10 46 29 PM

Screen Shot 2022-07-28 at 10 46 36 PM

Screen Shot 2022-07-28 at 10 46 46 PM

Screen Shot 2022-07-28 at 10 47 17 PM

Screen Shot 2022-07-28 at 11 12 16 PM

**Old Screenflow**
Screen Shot 2022-07-11 at 5 49 39 PM

Screen Shot 2022-07-13 at 11 36 11 AM

Screen Shot 2022-07-13 at 11 36 52 AM


resources/webviews/aksclusterproperties/clusterproperties.html
```diff
@@ -7,6 +7,34 @@
...
     <!-- Link to the css file -->
     <link rel="stylesheet" href="{{cssuri}}">
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
+    <script src="https://use.fontawesome.com/7dddd54b5c.js"></script>
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js"
+        integrity="sha256-9/aliU8dGd2tb6OSsuzixeV4y/faTqgFtohetphbbj0=" crossorigin="anonymous"></script>
+
+    <script>
+        // https://code.visualstudio.com/api/extension-guides/webview#passing-messages-from-a-webview-to-an-extension
+        const vscode = acquireVsCodeApi();
```
peterbom, line 17: The docs in the link say "For security reasons, you must keep the VS Code API object private and make sure it is never leaked into the global scope."

Could we move it inside the document-ready function?

resources/webviews/aksclusterproperties/clusterproperties.html
```diff
@@ -7,6 +7,34 @@
...
+
+        $(document).ready(function () {
+            $('#Startcluster').click(function () {
+                $(".spinner").show();
+                $(this).hide();
+                vscode.postMessage({ command: "startCluster" });
+                // https://docs.microsoft.com/en-us/rest/api/storageservices/setting-timeouts-for-blob-service-operations
+                setTimeout(function() {$(".spinner").hide()}, 30000);
```
peterbom, line 25: Do we need to hide the spinner at all?

First off, I thought the extension could notify the webview when the operation finished, and then it could hide the spinner.

But then I noticed that when the operation finishes, the entire webview html gets replaced. So effectively the spinner will be shown for the rest of the current document's life, and I *think* that means we don't need to hide it?

src/commands/aksClusterProperties/aksClusterProperties.ts
```diff
@@ -19,37 +19,112 @@ export default async function aksClusterProperties(
...
+
+    await longRunning(`Loading webview ${clustername} for cluster properties.`,
+      async () => {
+        let clusterData: ClusterARMResponse | undefined;
         const webview = createWebView('AKS Cluster Properties', `AKS properties view for: ${clustername}`);
 
-        webview.html = getWebviewContent(clusterInfo.result, extensionPath);
+        webview.onDidReceiveMessage(
+          async (message) => {
+              if (message.command === "startCluster") {
```
peterbom, line 65: Could we simplify this by just doing:
```javascript
clusterData = await onReceivePerformOperations(cloudTarget, clustername, message.command);
```

src/commands/aksClusterProperties/aksClusterProperties.ts
```diff
@@ -19,37 +19,112 @@ export default async function aksClusterProperties(
...
+                clusterData = await onRecievePerformOperations(cloudTarget, clustername, 'start');
+              } else if (message.command === "stopCluster") {
+                clusterData = await onRecievePerformOperations(cloudTarget, clustername, 'stop');
+              }
+              const clusterState = await determineClusterState(cloudTarget, clustername);
+              if (!clusterState || !clusterData) {
+                  return;
```
peterbom, line 72: This should 'never' happen, right? If we end up here, we don't actually have the required information to display cluster properties correctly. So, in that case, should we close the webview and maybe show an error?

src/commands/aksClusterProperties/aksClusterProperties.ts
```diff
@@ -19,37 +19,112 @@ export default async function aksClusterProperties(
...
+
+    return await longRunning(`Loading ${cloudTarget.name} cluster data.`,
+      async () => {
+        const clusterInfo = await getClusterProperties(cloudTarget, cloudTarget.name);
+        return clusterInfo;
+      }
+    );
+}
+
+async function onRecievePerformOperations(
```
peterbom, line 96: nit: spelling (receive)

src/commands/aksClusterProperties/aksClusterProperties.ts
```diff
@@ -19,37 +19,112 @@ export default async function aksClusterProperties(
...
+  cloudTarget: AksClusterTreeItem,
+  clusterName: string,
+  eventName: string
+): Promise<ClusterARMResponse | undefined> {
+
+    let startStopClusterInfo: Errorable<string>;
+    if (eventName === 'start') {
```
peterbom, line 103: Could we use a `switch` instead?

src/commands/aksClusterProperties/aksClusterProperties.ts
```diff
@@ -19,37 +19,112 @@ export default async function aksClusterProperties(
...
+     startStopClusterInfo = await longRunning(`Starting cluster.`, () => startCluster(cloudTarget, clusterName) );
+    } else {
+      startStopClusterInfo = await longRunning(`Stopping cluster.`, () => stopCluster(cloudTarget, clusterName));
```
peterbom, line 106: Just to protect against weird bugs in the future I'd prefer to use an explicit `case` (or `else if`) for `stop`, rather than the catch-all `else`. And then the `default`/`else` case would be reserved for reporting an error that should never happen.

src/commands/utils/clusters.ts
```diff
@@ -16,6 +16,13 @@ export interface ClusterARMResponse {
     readonly type: string;
 }
 
+export enum ClusterState {
```
peterbom, line 19: If this is an overall 'cluster' state, we could end up with these states increasing exponentially as we combine more features. For example, say we add in scaling and upgrade we might end up with things like 'StartedAndScalingUp', along with a complicated translation layer between RP-level states (provisioning state, power state, etc.) and our VSCode representation of cluster states.

However, I see their utility here. I think my worry above could be sidestepped completely if we just rename this enum to `ClusterStartStopState`. What do you reckon?

src/commands/utils/clusters.ts
```diff
@@ -102,3 +110,71 @@ export async function getClusterProperties(
         return { succeeded: false, error: `Error invoking ${clusterName} managed cluster: ${ex}` };
     }
 }
+
+export async function determineClusterState(
+    target: AksClusterTreeItem,
+    clusterName: string
+): Promise<string | undefined> {
```
peterbom, line 117: My preference would be to return `Promise<errorable<string>>` and let callers decide whether to show the error message. This would make it easier to consume in other places (I'm thinking `startCluster` and `stopCluster`).

src/commands/utils/clusters.ts
```diff
@@ -102,3 +110,71 @@ export async function getClusterProperties(
...
+    }
+}
+
+export async function startCluster(
+    target: AksClusterTreeItem,
+    clusterName: string
+): Promise<Errorable<string>> {
+    try {
+        const containerClient = new azcs.ContainerServiceClient(target.subscription.credentials, target.subscription.subscriptionId);
+        const clusterInfo = (await containerClient.managedClusters.get(target.resourceGroupName, clusterName));
```
peterbom, line 144: It looks as if we're duplicating the logic in `determineClusterState` here. I'd love it if we could just call that instead... And if `determineClusterState` is changed to return an errorable, then if it fails we can just return the result of that directly.

peterbom (approved):

Issue #152, opened 2022-08-04
Handle cases for Sovereign Clouds and other Gov Scenarios.


Tatsinnit:
I guess from initial day the primary target of this extension was common Azure/sub scenarios, It will be worth to document what are the current VsCode extension command which don't support USGov or Sovereign cloud due to underlying SDK or any other usage. For example Detector Diagnostics work fine as its a direct ARM resourceClient call. Like Periscope currently has no support for sovereign cloud.

For example: I see `ContainerServices -> arm` API has pre-requisite of "azure subscription"  - it best to test if this will work, as with sovereign clouds and possibly document and come out with moving forward solution.

* https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/containerservice/arm-containerservice#prerequisites 
    * We use this API for `show Properties` `get kubeconfig` and few other places.
* 💡💡 Idea: How about information message early for those commands which have dependency on ContainerService api?

Here is an example how endpoint under the hood for ARM Client SDK's differ hence one works and another breaks.

A very early triage so just adding as discussion point:

## **Solution** 

* Lets handle the `endpoint` based on the `sovereign` cloud : for us gov it will be :  `"https://management.usgovcloudapi.net" ` 

### **extra detail**

* ResourceManagemt client call works because the baser is fed correctly:

<img width="755" alt="Screen Shot 2022-08-04 at 11 30 05 PM" src="https://user-images.githubusercontent.com/6233295/182837322-93c616df-cbcc-489f-aa29-d7589472baed.png">

* ConatainerClient ARM API has only management.azure as endpoint:

<img width="929" alt="Screen Shot 2022-08-04 at 11 32 01 PM" src="https://user-images.githubusercontent.com/6233295/182837268-0d32f6dd-4eff-4a52-aee2-b26d910df679.png">

Thanks, ❤️🙏

PR #153, opened 2022-08-04
support usgov cloud endpoint.

Tatsinnit:
This PR checks for the location which determine the usgov cloud and then we handle the `endpoint` for the `ContainerServiceClient` which kind of fixes most of the command support for the Support of USGov cloud AKS cluster.

and hence fixes/enhances #152 

Thanks heaps and fyi cc: @peterbom, @gambtho , @rzhang628 

Thanks. ❤️🙏☕️


rzhang628 (approved):
looks good and approved

peterbom (approved):

PR #156, opened 2022-08-09
Poll Starter Workflow: Remove comparison between 'seen' and latest templates

peterbom:
There is a bug with the current polling mechanism, which goes roughly as follows:

1. A workflow run detects a change in the source templates. This updates both the 'seen' mapping (which then gets uploaded as an artifact), and the local workspace. It then creates a PR from the local changes. This is all working as intended.
2. The PR is not merged before the next run (maybe it's out of work hours or we're busy with something else).
3. The workflow runs again. This time, the 'seen' mapping contains the latest SHA hashes of the files and so no changes are made to the local workspace. It then creates/updates the PR, but because there are no changes [the PR is automatically closed](https://github.com/peter-evans/create-pull-request#action-behaviour).

What _should_ happen is that the PR is kept open unchanged (i.e. the latest changes are preserved). There are a few ways of achieving this:
1. Only run the `create-pull-request` step if there have actually been changes (i.e. make it conditional on an output from the comparing/downloading step). This approach is tricky to reason about and might suffer from some weird edge cases if some but not all of the source templates are updated between runs.
2. Check out the `starter-workflow-updates` branch _before_ the comparing/downloading step. This will ensure the PR won't get 'reset' and is probably simpler than (1), but I think it'd entail extra conditional logic depending on whether the remote branch already exists.
3. _Always_ update the local workspace with the latest templates and create/update a PR from that. This is the most straightforward option (all the 'seen' tracking can be removed). It loses some functionality which I think was added to mitigate concerns we had around churn or unwanted updates in the source templates. However, given that the source templates are updated very infrequently, I don't believe this will be an issue in practice.

These changes are for (3).

@Tatsinnit, @OliverMKing: any thoughts/comments?

### DIFF
--- a/.github/workflows/poll-starter-workflows.yml
+++ b/.github/workflows/poll-starter-workflows.yml
@@ -15,25 +15,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Download run information
-        uses: dawidd6/action-download-artifact@v2
-        continue-on-error: true
-        id: download
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          workflow: poll-starter-workflows.yml
-          workflow_conclusion: success
-          name: seen
-          path: /tmp
-
-      - name: Create new run information
-        if: steps.download.outcome != 'success'
-        run: |
-          echo "Creating new run information record"
-          declare -A seenWorkflows
-          touch /tmp/seenWorkflows
-          declare -p seenWorkflows > /tmp/seenWorkflows
-
       - name: Download starter workflows
         run: |
           # get deployment workflows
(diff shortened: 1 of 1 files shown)


OliverMKing:
There's another fix that I added to the draft repo [here](https://github.com/Azure/draft/blob/main/.github/workflows/poll-starter.yml). Their version of the workflow only creates / updates the PR and branch if there were changes found (solution 1 you proposed). The changes that are proposed in this PR would make it not work well if you ever needed to modify the starter workflows that get merged in.

We also swapped to [comparing against the main starter workflow branch](https://github.com/Azure/draft/blob/main/.github/workflows/poll-starter.yml#L10) due to a recent change in GitHub's process.

peterbom:
> There's another fix that I added to the draft repo [here](https://github.com/Azure/draft/blob/main/.github/workflows/poll-starter.yml). Their version of the workflow only creates / updates the PR and branch if there were changes found (solution 1 you proposed). The changes that are proposed in this PR would make it not work well if you ever needed to modify the starter workflows that get merged in.
> 
> We also swapped to [comparing against the main starter workflow branch](https://github.com/Azure/draft/blob/main/.github/workflows/poll-starter.yml#L10) due to a recent change in GitHub's process.

Thanks @OliverMKing!

I've updated the branch to `main`, and indeed there are [some changes](https://github.com/peterbom/vscode-aks-tools/pull/7/files) there that we hadn't been picking up.

Still, I don't see anything in those changes that would break our extension (we're just looking for two strings, `"your-resource-group"` and `"your-cluster-name"`, and those are still unchanged). If we _do_ at some point encounter a change that'd break our extension, we'd probably need to handle it in our code rather than fixing up the auto-generated PR.

So my preference is still to go for the simple approach for now, noting that the fix in the `draft` repo exists and that we can adopt it further down the line if it turns out we do need it.

@Tatsinnit - does that make sense to you?

Tatsinnit (approved):

PR #157, opened 2022-08-11
Update latest Periscope version and support optional Windows features

peterbom:
This updates the extension to use the latest version of Periscope (0.0.10) by default.

This version of Periscope has an optional Kustomize component, which should only be deployed if the cluster supports it. For this, there is some extra checking in the Periscope helper module to determine if this component is supported. Although there is just one feature/component right now, the code is structured such that the pattern can be extended to add further components later.

The `DIAGNOSTIC_RUN_ID` configuration value is required to support the new optional Periscope component, but may in future be required by the Periscope application.

### DIFF
--- a/src/commands/periscope/helpers/periscopehelper.ts
+++ b/src/commands/periscope/helpers/periscopehelper.ts
@@ -7,12 +7,15 @@ import { PeriscopeStorage, PeriscopeHTMLInterface } from '../models/storage';
 import * as amon from '@azure/arm-monitor';
 import * as path from 'path';
 import * as fs from 'fs';
+import * as semver from 'semver';
 import AksClusterTreeItem from '../../../tree/aksClusterTreeItem';
 import * as tmpfile from '../../utils/tempfile';
 import { getRenderedContent, getResourceUri } from '../../utils/webviews';
-import { Errorable, failed } from '../../utils/errorable';
+import { combine, Errorable, failed } from '../../utils/errorable';
 import { invokeKubectlCommand } from '../../utils/kubectl';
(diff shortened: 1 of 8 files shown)


src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -287,4 +309,96 @@ function getHostName(output: string): Errorable<string> {
...
+    return { succeeded: true, result: runtimes };
+}
+
+function isWindowsHcpSupported(k8sVersion: string, containerdVersions: string[]): boolean {
+    // Based on https://docs.microsoft.com/en-us/azure/aks/use-windows-hpc#limitations
+    if (semver.lt(k8sVersion, "1.23.0")) {
+        return false;
+    }
+
+    if (containerdVersions.length === 0) {
```
Tatsinnit, line 393: nit - can merge with above `if` as they both return `false`

src/commands/utils/azurestorage.ts
```diff
@@ -9,7 +9,7 @@ export enum LinkDuration {
 function sasDuration(duration: LinkDuration): number {
     switch (duration) {
         case LinkDuration.StartTime: return 2 * 60 * 1000;
-        case LinkDuration.DownloadNow: return 5 * 60 * 1000;
+        case LinkDuration.DownloadNow: return 15 * 60 * 1000; // 15 minutes to allow for Windows image pulls and log export
```
Tatsinnit, line 12: 💡Note: This will not change any behaviour in existing cases because its just to handle windows. Please correct me if I am wrong. 🥷

Tatsinnit (approved):
👍 🧶 🎊

PR #161, opened 2022-08-30
Fix error when running Periscope using a branch release tag

peterbom:
The release tag setting for Periscope allows any GH reference (branch, commit, tag), so this may not be a valid semver value.

Some logic for determining which components to deploy checks the version of Periscope being deployed using the release tag, which results in runtime errors when this is a branch or commit.

This changes that logic to use the image version, which is more likely to be a semver value (it still might not be, so we need to try and parse it first to avoid runtime errors).


Tatsinnit (approved):
🎉☕️❤️

PR #165, opened 2022-09-07
Generate periscope links from run ID

peterbom:
Some UX changes that take advantage of the fact that the Periscope Run ID is now deterministic.
- The 7-day shareable URL can be generated and copied immediately (before blobs are uploaded)
- User no longer has to click a 'refresh' button to see if results are uploaded
- Checking if results are uploaded is much faster (because we don't have to iterate through all blobs in the container)
- Nodes are retrieved in advance rather than inferred from uploaded blobs, meaning:
  - we can display upload progress for all nodes at once
  - we can retrieve and display Periscope pod logs for any node (even if results for those nodes have not been uploaded), which can be used to identify any errors running Periscope 

image

The failure webview now displays the Periscope VSCode config settings that were used to generate the Kustomize template:

image


rzhang628 (approved):

Tatsinnit (approved):
⛩️ 🚢 Thank you for adding me to the review, Loogs good to me, code wise. Thanks.

Issue #173, opened 2022-11-10
Kubectl Don't work for AAD Enabled clusters. 💡☕️


Tatsinnit:
**Describe the bug**
Kubectl Don't work for AAD Enabled clusters.

**Mention the platform you are using**
Mention what platform you are using (oss, vscode) and pick the corresponding label.

**To Reproduce**

* Go to any feature currently using `kubctl` like ` get all pods` for AAD Enabled cluster.
    * The command run will hang as `Longrunning kubectl...`

**Expected behavior**
It should display all the pods under namespaces.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MAC O/S
 - Version - 1.73

**Additional context**
Need to discuss how we can fix the interactive AAD scenario here.

mrafsyam:
Seconded this. 

Running any kubectl command, you'll get something like this for a long time
![image](https://user-images.githubusercontent.com/6120781/201244006-0f4cac70-deaf-4065-8b24-c44abf5fabf2.png)

Eventually, this failed

Tatsinnit:
☕️☕️ The key of this issue is the behind the scenes stuff, for AAD Auth clusters, so if we keep this long running for a while, this happens: (It happening because behind the scene auth has interactive mode and this all appears once the token expires.)

<img width="582" alt="Screenshot 2022-11-11 at 8 29 18 PM" src="https://user-images.githubusercontent.com/6233295/201289378-9d04e5b9-5fbc-40b7-8793-8ed2543feb14.png">

peterbom:
Some notes after a bit of experimentation and discussion with @Tatsinnit...

Some background detail on the existing implementation: we download a temporary `kubeconfig` file for each run of the command. This is why it happens for every invocation of a`kubectl` commands, rather than using any cached tokens that might exist in users' `.kube` directories.

Ideally, we want to avoid forcing the user to perform an interactive login, because that would involve either directly putting the user into a command prompt (unexpected UX), or letting the extension interact with `stdout`/`stdin` of an external process, which is complex, fragile, and fraught with edge cases.

**What is really happening with interactive login?**
Other than local accounts (not applicable to AAD clusters), AKS supports two 'formats' of authentication plugins in `kubeconfig` files: `azure` and `exec`:
[...]

PR #174, opened 2022-11-17
Supply access token to kubelogin to avoid interactive login

peterbom:
~~Partially addresses #173. Currently assumes that `kubelogin` will be on the user's PATH, which is likely not to be the case, so pending some changes to download it.~~

This builds on the work done by @Tatsinnit in #175, which provides a consistent version of `kubelogin` for us to use.

The extension will now update the `exec` section of the `kubeconfig` file for AAD clusters, so that subsequent invocations of `kubectl` using that `kubeconfig` will use a cached AAD access token that we generate. The result of this is that `kubectl` will run non-interactively...**until the access token expires (after about an hour), or the cache in the temp directory is cleaned up by the OS.**

There are two scenarios in this extension where we use this functionality:
1. Running `kubectl` commands
2. The "Save Kubeconfig" and "Merge into Kubeconfig" commands (which is used by the Kubernetes extension's cluster pane nodes)

For (1), we generate a new temporary kubeconfig and access token for every command that's run. That means we have ample time to run the command before any cached data expires or is cleaned up.

For (2), we update the _user's_ `kubeconfig` file, which is long-lived, meaning we need to consider what the behaviour will be when the cached credentials are no longer valid or present. With valid credentials, the cluster pane will expand tree nodes for AAD clusters just as it does for non-AAD clusters. After that, further interaction with the cluster will pop open a browser window for the user to sign in (because `kubectl` invokes `kubelogin` with the `interactive` login option).

I think this is an improvement over the previous behaviour, in which expanding an AAD cluster's tree nodes in the Kubernetes extension will spin for several minutes waiting for a devicecode login that the user never sees, before eventually timing out and displaying an error:
image


Tatsinnit:
❤️☕️ Thank you for adding me to the review, this looks so good and satisfying our work into action, I tested this from the branch - and see how smooth it looks:

I will go through the code soon. Below is the **successful** result we get, as a test I deleted my local `kubelogin` to check if it will download for the users, this also fixes the `kubernetes` extension use case as well for the users using `AKS` extension as a merge to kubeconfig feature. #173  - has rest of detail.

💡❤️☕️ A gentle ping and fyi to @rzhang628, @gambtho , @squillace and @itowlson 

Screenshot 2022-11-27 at 4 28 39 AM

Screenshot 2022-11-27 at 4 28 51 AM

src/commands/utils/clusters.ts
```diff
@@ -73,45 +74,180 @@ export function getAksClusterSubscriptionItem(commandTarget: any, cloudExplorer:
...
+    loginMethod: string
+}
+
+function readExecOptions(execArgs: [string]): ExecOptions {
+    // The 'exec' command args are made up of a subcommand (get-token) and a number of name/value options.
+    const [subcommand, ...options] = execArgs;
+
+    // Extract the options into an object for convenient lookup.
+    const optionLookup: { [name: string]: string } = options.reduce((result: { [name: string]: string }, _arg: string, index: number) => {
+        return index % 2 === 0 ?
```
Tatsinnit, line 172: 💡 I wonder if we add couple of comments around this etc. just like details you had in few functions above.

src/commands/utils/clusters.ts
```diff
@@ -73,45 +74,180 @@ export function getAksClusterSubscriptionItem(commandTarget: any, cloudExplorer:
...
+}
+
+function buildExecOptionsWithCache(execOptions: ExecOptions, cacheDir: string) {
+    return [
+        execOptions.subcommand,
+        "--environment", execOptions.environment,
+        "--server-id", execOptions.serverId,
+        "--client-id", execOptions.clientId,
+        "--tenant-id", execOptions.tenantId,
+        // In a non-interactive environment, 'devicecode' login now returns an error, see https://github.com/Azure/kubelogin/pull/145.
```
Tatsinnit, line 194: 💡☕️ Side not end observation -> This is great information to share, worth for any readers for throwing some light on : `"--login", "interactive",` - Thanks.

Tatsinnit (approved):
❤️☕️🙏 Thank you so much for adding me to this review, this looks good, I have tested the key work around the `kubelogin`.  (Added working screenshot in my previous comments)

Looks good to me. Thank you!!

For extra info - Regarding the `usgov`, `public` login change if the if the object - `cluster.result.subscription.environment.portalUrl` bring right url then it will work. 

Fyi for eye balling or folks who kindly contributed in our discussion here: #173 ❤️☕️💡 @rzhang628 , @squillace , @itowlson and @gambtho

src/commands/utils/clusters.ts
peterbom: Good point, that code isn't obvious at all - hope this explains it a bit better.

PR #175, opened 2022-11-18
Download mechnism for kubelogin.

Tatsinnit:
This PR enables the download mechanism for `kubelogin` which will help to cater AAD Cluster scenarios and the work Pete and I are doing. #173 #174 

Thanks heaps, opening this as a PR to connect the dots. ❤️☕️🙏

Following PR enables download of kibelogin which will be coming handy for the AAD non-interactive work, for AAD Aks cluster scenarios.

Thank you so much all in advance and you kind review. ❤️🙏🎉

Final outcome will be something like this: (macosx output below)

Screenshot 2022-11-22 at 5 32 57 PM


peterbom (approved):

src/commands/utils/helper/kubeloginDownload.ts
```diff
@@ -0,0 +1,137 @@
...
+import * as os from 'os';
+import * as download from '../download/download';
+import * as path from 'path';
+import { getKubeloginConfig } from '../config';
+import { Errorable, failed } from '../errorable';
+import { moveFile } from 'move-file';
+import { longRunning } from '../host';
+
+function getBaseInstallFolder(): string {
+   return path.join(os.homedir(), `.vs-kubernetes/tools`);
```
peterbom, line 12: My bad. Should be
`return path.join(os.homedir(), `.vs-kubernetes/tools/kubelogin`);`

peterbom:
Tested on Windows and WSL - LGTM!

PR #176, opened 2022-11-28
Remove deprecating detector diagnostics for networkconnectivity.

Tatsinnit:
This PR removes the old network connectivity detector diagnostics from vscode.

we will work towards adopting to a new detector but we need to access what we could display and what we can't given is that `aks-category-connectivity` has some graph data which we mightn't be able to display in vscode web views.

Thanks heaps, and fyi @peterbom and @rzhang628 ❤️☕️🙏


peterbom (approved):
Looks great - thanks @Tatsinnit!

I gave the visx a run-through too, and that's and all good.

PR #180, opened 2022-12-22
Enable delete cluster functionality.

Tatsinnit:
This PR is written to enable all the achievable `managed cluster operations` which we can integrate with users for `one-click` - `aks vscode` functionality. 

This PR adds the `submenu` along with one click `delete cluster` api trigger via the TS api: https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/delete?tabs=JavaScript 

🥷💡🙏 Please note: the api return void but it takes time, so I have added appropriate messaging for user to see what is happening. Also, to those who have so far used the `delete cluster operation` from portal have seen little-longer to delete the cluster which is the responsibility of API, hence you will see the `long running` beahviour via vscode as well.

💼☕️ Key advantage of this feature will be that not only in one-click user can delete the cluster but also they can send multiple parallel delete operation at once. 

Thank you so much and fyi to: ❤️💡🥷 @rzhang628, @squillace , @peterbom and @gambtho ☕️❤️🙏

Here are some screenshots: 

Screenshot 2022-12-22 at 10 40 12 AM

Screenshot 2022-12-22 at 10 16 01 AM

Screenshot 2022-12-22 at 10 07 03 AM

Screenshot 2022-12-22 at 10 15 41 AM

### DIFF
--- a/src/commands/aksDeleteCluster/aksDeleteCluster.ts
+++ b/src/commands/aksDeleteCluster/aksDeleteCluster.ts
@@ -1,0 +1,37 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from "@microsoft/vscode-azext-utils";
+import { deleteCluster, getAksClusterTreeItem } from '../utils/clusters';
+import { failed, succeeded } from '../utils/errorable';
+import { longRunning } from '../utils/host';
+
+export default async function aksDeleteCluster(
+    _context: IActionContext,
+    target: any
+): Promise<void> {
+    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+
+    const cluster = getAksClusterTreeItem(target, cloudExplorer);
+    if (failed(cluster)) {
+      vscode.window.showErrorMessage(cluster.error);
+      return;
+    }
+
+    const clusterName = cluster.result.name;
+
+    vscode.window.showInformationMessage(`Do you want to delete cluster ${clusterName}?`, "Yes", "No")
+    .then(async answer => {
+      if (answer === "Yes") {
+
+        const result = await longRunning(`Deleting cluster ${clusterName}.`, async () => { return await deleteCluster(cluster.result, clusterName) });
+
+        if (failed(result)) {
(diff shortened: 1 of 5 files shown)


src/commands/aksDeleteCluster/aksDeleteCluster.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    const cluster = getAksClusterTreeItem(target, cloudExplorer);
+    if (failed(cluster)) {
+      vscode.window.showErrorMessage(cluster.error);
+      return;
+    }
+
+    const clusterName = cluster.result.name;
+
+    vscode.window.showInformationMessage(`Do you want to delete cluster ${clusterName}?`, "Yes", "No")
```
peterbom, line 22: Could we avoid the nested callback? I.e.
```javascript
const answer = await vscode.window.showInformationMessage(`Do you want to delete cluster ${clusterName}?`, "Yes", "No");
if (answer === "Yes") {
...
}
```

src/commands/utils/clusters.ts
```diff
@@ -352,4 +352,18 @@ export async function getWindowsNodePoolKubernetesVersions(
...
+    target: AksClusterTreeItem,
+    clusterName: string
+): Promise<Errorable<string>> {
+    try {
+        const containerClient = getContainerClient(target);
+        await containerClient.managedClusters.beginDeleteAndWait(target.resourceGroupName, clusterName)
+
+        return { succeeded: true, result: "Delete cluster succeeded." };
+    } catch (ex) {
+        return { succeeded: false, error: `Error invoking ${clusterName} managed cluster: ${ex}` };
```
peterbom, line 367: One thing I noticed recently: the `errorable` module exposes a `getErrorMessage` function that can be used for displaying errors in strings. Shall we start making more use of that?
```javascript
return { succeeded: false, error: `Error invoking ${clusterName} managed cluster: ${getErrorMessage(ex)}` };
```

peterbom (approved):
Looks great! Just made a couple of suggestions, nothing major

rzhang628 (approved):

PR #181, opened 2022-12-23
Support ASO for non-Azure clusters and allow subscription selection

peterbom:
There are currently a few limitations in the current ASO installation command, namely:
- The cluster must be in the same subscription as the Azure resources generated by ASO
- The cluster must be an AKS cluster (it's often handy to be able to try this on a `kind` cluster)
- There is no checking that the supplied service principal ID and secret are valid, or have any permissions in the subscription

The changes here address these as follows:
- The command is available from both the AKS cloud explorer (as before) _and_ the Kubernetes cluster explorer (new).
- After entering the SP credentials, they are checked by creating a credential and checking it is valid in the current Azure session(s) by looking it up using the MS Graph API.
- The credential is then used to look up its own role assignments in the filtered list of subscriptions exposed by the Azure extension.
- The user is then prompted to choose the target subscription from any of the subscriptions for which the SP has at least one role assignment.

Side-benefit: if there are any _other_ commands that make sense for both Azure clusters and other clusters in cluster-explorer, we could re-use the code here that retrieves cluster information (name and kubeconfig) for either. The `kubectl` commands would be one such possibility.

Some other changes:
- Some refactoring to reduce the number of files in the ASO command folder: one helper for SP-related functionality, and one for installation.
- Extra checks and timeouts in the installation process to ensure each successive step has time to complete successfully.

image

image

image

image


src/commands/utils/azureAccount.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface AzureSession {
+    readonly environment: Environment;
+    readonly userId: string;
+    readonly tenantId: string;
+    readonly credentials2?: TokenCredential;
+}
+
+export interface AzureSubscription {
+	readonly session: AzureSession;
```
Tatsinnit, line 26: 💡☕️🎊 Extereme minor nit: `spaces`.

src/commands/utils/clusters.ts
```diff
@@ -28,6 +29,63 @@ export enum ClusterStartStopState {
...
+        kubeconfigYaml: fs.readFileSync(kubeconfigPath, 'utf8')
+    };
+
+    return { succeeded: true, result };
+}
+
+function getPath(kubeconfigPath: ConfigurationV1.KubeconfigPath): string {
+    switch (kubeconfigPath.pathType) {
+        case 'host':
+            return kubeconfigPath.hostPath;
```
Tatsinnit, line 83: 💡 Question: This is not changing the existing behaviour? but curious as to how `hostPath` vs `wslPath` was handled before.

src/commands/azureServiceOperators/helpers/servicePrincipalHelper.ts
```diff
@@ -0,0 +1,148 @@
...
+        id: searchResult.id,
+        displayName: searchResult.displayName,
+        session,
+        credential
+    };
+
+    return { succeeded: true, result: spInfo };
+}
+
+function getGraphClientBaseUrl(environment: Environment): string {
```
Tatsinnit, line 108: 💡 Thinking outlaud: I wonder if we rename this to something more contextual? although the api is graph (Just sharing some thought homz)

package.json
```diff
@@ -213,7 +213,7 @@
                 },
                 {
                     "command": "aks.installAzureServiceOperator",
-                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i"
+                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i || view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.\\w*cluster$/i"
```
Tatsinnit, line 216: 💡❤️☕️🎅 For arichve reasons, and please correct me if it's not entirely correct, this work, we will document this as a final release PR for how this enables the ASO for kind cluster et. al. as well and hence a hook to `k8s cncf` extension as well. 

I think this doc doesn't need to be in other extension because its hook which user will only get is the `k8s extension` is downloaded.

Tatsinnit (approved):
❤️☕️🙏🎅 Thank you so much for adding me to this review. Looks good to me.

Very minor comments added. I think you have tested this as well. We can get this tested together with the other work from main branch. 

Thank you for this, and fyi + cc: @rzhang628 🙏❤️☕️

src/commands/utils/clusters.ts
peterbom: This is new, because we were never reading the `kubeconfig` file used by cluster explorer before. Until now, we've only been getting the kubeconfig for AKS clusters, but if the user invokes the command from the cluster explorer, we need a different way of getting the kubeconfig (which I grabbed from the [k8s extension docs](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/master/docs/extending/configuration.md#detecting-the-kubernetes-configuration)). I'll put a comment in.

src/commands/utils/azureAccount.ts
peterbom: Oops - guess I must have copy-pasted something!

src/commands/azureServiceOperators/helpers/servicePrincipalHelper.ts
peterbom: As an aside, I've always found the name 'graph' for the API kind of confusing...

Is `getMicrosoftGraphClientBaseUrl` any better? It's a bit long, but it distinguishes it from Azure AD Graph at least. Any thoughts/suggestions?

package.json
peterbom: I think that's correct. It enables ASO for non-AKS clusters (whatever appears in cluster explorer).

I don't think we need any docs in the k8s extension. This makes use of [an extension point](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/master/docs/extending/commandtargets.md#cluster-explorer-node-right-click) for adding commands to the tree view, but the command itself is here.

PR #182, opened 2022-12-28
Managed cluster rotate certificate feature.

Tatsinnit:
This PR is written to piggy back on all the achievable `managed cluster operations` which we can integrate with users for `one-click` - `aks vscode` functionality. 

This PR adds the one click `rotate cluster certificate` api trigger via the TS api: https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/delete?tabs=JavaScript 

🥷💡🙏 Please note: the api return void but it takes time, so I have added appropriate messaging for user to see what is happening. Also, to those who have so far used the `rotate cluster certificate operation` from portal have seen little-longer to delete the cluster which is the responsibility of API, hence you will see the `long running` beahviour via vscode as well.

💼☕️ Key advantage of this feature will be that not only in one-click user can delete the cluster but also they can send multiple parallel delete operation at once. 

Thank you so much and fyi to: ❤️💡🥷 @rzhang628, @squillace , @peterbom and @gambtho ☕️❤️🙏

Here are some screenshots: 

Screenshot 2022-12-28 at 10 37 00 AM

Screenshot 2022-12-28 at 10 37 36 AM

Screenshot 2022-12-28 at 10 37 43 AM

Screenshot 2022-12-28 at 10 53 56 AM

### DIFF
--- a/src/commands/aksRotateClusterCert/aksRotateClusterCert.ts
+++ b/src/commands/aksRotateClusterCert/aksRotateClusterCert.ts
@@ -21,8 +21,12 @@ export default async function aksRotateClusterCert(
 
   const answer = await vscode.window.showInformationMessage(`Do you want to rotate cluster ${clusterName} certificate?`, "Yes", "No");
 
+  if (answer !== "Yes") {
+    return;
+  }
+
   if (answer === "Yes") {
-    const result = await longRunning(`Rotating cluster certificate for ${clusterName}.`, async () => { return await rotateClusterCert(cluster.result, clusterName) });
+    const result = await longRunning(`Rotating cluster certificate for ${clusterName}.`, async () => rotateClusterCert(cluster.result, clusterName) );
 
     if (failed(result)) {
       vscode.window.showErrorMessage(result.error);
--- a/src/commands/aksRotateClusterCert/aksRotateClusterCert.ts
+++ b/src/commands/aksRotateClusterCert/aksRotateClusterCert.ts
@@ -1,0 +1,35 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from "@microsoft/vscode-azext-utils";
+import { getAksClusterTreeItem, rotateClusterCert } from '../utils/clusters';
+import { failed, succeeded } from '../utils/errorable';
+import { longRunning } from '../utils/host';
+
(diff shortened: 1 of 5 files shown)


src/commands/aksRotateClusterCert/aksRotateClusterCert.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (failed(cluster)) {
+    vscode.window.showErrorMessage(cluster.error);
+    return;
+  }
+
+  const clusterName = cluster.result.name;
+
+  const answer = await vscode.window.showInformationMessage(`Do you want to rotate cluster ${clusterName} certificate?`, "Yes", "No");
+
+  if (answer === "Yes") {
```
peterbom, line 24: This is more of a personal preference, so feel free to disagree, but... I'd rather do:
```javascript
if (answer !== "Yes") {
    return;
}
const result = await longRunning(...);
```

That way, the 'main' code path doesn't get nested inside a conditional block, and it's easy to read where it exits early.

src/commands/aksRotateClusterCert/aksRotateClusterCert.ts
```diff
@@ -0,0 +1,35 @@
...
+    const result = await longRunning(`Rotating cluster certificate for ${clusterName}.`, async () => { return await rotateClusterCert(cluster.result, clusterName) });
```
peterbom, line 25: I think the async/await is not necessary here, and you could instead make it:
```javascript
const result = await longRunning(`Rotating cluster certificate for ${clusterName}.`, () => rotateClusterCert(cluster.result, clusterName));
```

src/commands/utils/clusters.ts
```diff
@@ -426,4 +426,18 @@ export async function deleteCluster(
...
     }
+}
+
+export async function rotateClusterCert(
+    target: AksClusterTreeItem,
+    clusterName: string
+): Promise<Errorable<string>> {
+    try {
+        const containerClient = getContainerClient(target);
+        await containerClient.managedClusters.beginRotateClusterCertificatesAndWait(target.resourceGroupName, clusterName)
```
peterbom, line 437: nit: missing semicolon

peterbom (approved):
Looks great! Just made some minor comments.

rzhang628 (approved):

PR #185, opened 2023-01-04
Move padding for ASO webview to correct css file

peterbom:
I mistakenly [added some CSS padding to the wrong file](https://github.com/Azure/vscode-aks-tools/pull/181/files#diff-1b4ebc75fee5dcd8fb408864476596c83e3fb2ce6e49e11b20d261540e1249d5) when making the changes to support ASO for non-AKS clusters.

This moves it to the right place.


Tatsinnit (approved):
☕️❤️🥷🍿

PR #190, opened 2023-01-18
Feature/deploy undeploy inspektor gadget

Tatsinnit:
This PR enables the Inspector Gadget deploy/undeploy enablement via this extension for vscode audience.

Long term goal will be to expose the inspector gadget command for user.

☕️💡 We have done all the heavy lifting for the user, we download the binary depending  the underlying O/S 
☕️💡 Refactored a little for common download binary.
☕️💡 Implement - the deploy and undeploy.

Many thanks and fyi: @peterbom and @rzhang628 + @gambtho 

Screenshot 2023-01-08 at 1 17 12 PM

Screenshot 2023-01-18 at 6 55 24 PM

Screenshot 2023-01-18 at 6 55 18 PM


package.json
```diff
@@ -263,6 +279,11 @@
                     "submenu": "aks.K8sAPIEndpointHealthCheck",
                     "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i",
                     "group": "9@3"
+                },
+                {
+                    "submenu": "aks.inspektorGadgetSubMenu",
+                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i",
```
peterbom, line 285: To enable it in both the cloud explorer and cluster explorer, I think you can use this (see ASO command):
```json
"when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i || view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.\\w*cluster$/i"
```

package.json
```diff
@@ -431,6 +466,7 @@
         "@microsoft/vscode-azext-utils": "^0.3.8",
         "@sinonjs/fake-timers": "^6.0.1",
         "@types/chai": "^4.2.21",
+        "@types/shelljs": "^0.8.11",
```
peterbom, line 469: Can we remove the package dependencies for `@types/shelljs` and `shelljs`?

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,173 @@
...
+import { getKubernetesClusterInfo, KubernetesClusterInfo } from '../utils/clusters';
+import { getExtensionPath, longRunning } from '../utils/host';
+import { Errorable, failed } from '../utils/errorable';
+import * as tmpfile from '../utils/tempfile';
+import { getKubectlGadgetBinaryPath } from '../utils/helper/kubectlGadgetDownload';
+import { invokeKubectlCommand } from '../utils/kubectl';
+import path = require('path');
+import { createWebView, getRenderedContent, getResourceUri } from '../utils/webviews';
+
+enum InspektorGadget {
```
peterbom, line 13: Could this be named `InspektorGadgetCommand`, or maybe better just `GadgetCommand`?

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,173 @@
...
+    if (failed(clusterInfo)) {
+        vscode.window.showErrorMessage(clusterInfo.error);
+        return undefined;
+    }
+
+    await prepareInspektorGadgetInstall(clusterInfo.result, gadget, clusterInfo.result.kubeconfigYaml, kubectl);
+}
+
+async function prepareInspektorGadgetInstall(
+    cloudTarget: KubernetesClusterInfo,
```
peterbom, line 67: Could this parameter be named `clusterInfo` to match its type? At this stage, the code has no knowledge of vscode UI components.

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,173 @@
...
+    gadget: InspektorGadget,
+    kubeconfig: string,
```
peterbom, line 69: I'd prefer to remove this parameter, because it can be referenced from the `KubernetesClusterInfo`, which is already being passed in.

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,173 @@
...
+    clustername: string,
+    command: string,
+    clusterConfig: string,
+    kubectl: k8s.APIAvailable<k8s.KubectlV1>) {
+
+    const kubectlGadgetPath = await getKubectlGadgetBinaryPath();
+    let kubetlGadgetBinaryPath = "";
+
+    if (failed(kubectlGadgetPath)) {
+        vscode.window.showWarningMessage(`Gadget path is not found ${kubectlGadgetPath.error}`);
```
peterbom, line 115: Shouldn't we `return` immediately after showing the warning message? I don't think we can reliably continue if this fails, which will mean the binary path remains as `""`.

Also, if we bail here, we can delete the `if (kubectlGadgetPath.succeeded)` check below.

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    if (kubectlGadgetPath.succeeded) {
```
peterbom, line 118: Just to reduce the number of path variables in scope, I'd favour something like:
```javascript
const  binaryPathDir = path.dirname(kubectlGadgetPath.result);
```

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,173 @@
...
+      vscode.window.showErrorMessage(extensionPath.error);
+      return;
+    }
+
+    return await longRunning(`Loading ${clustername} kubectl command run.`,
+        async () => {
+            const commandToRun = `gadget ${command}`;
+            console.log(kubetlGadgetBinaryPath);
+            const  binaryPathDir = path.dirname(kubetlGadgetBinaryPath);
+            if (process.env.PATH !== undefined && (process.env.PATH.indexOf(binaryPathDir) < 0)){
```
peterbom, line 134: How about...
```javascript
if (process.env.PATH === undefined) {
    process.env.PATH = binaryPathDir
} else if (process.env.PATH.indexOf(binaryPathDir) < 0) {
    process.env.PATH = binaryPathDir + path.delimiter + process.env.PATH;
}
```

src/commands/utils/config.ts
```diff
@@ -38,21 +38,39 @@ export function getKustomizeConfig(): Errorable<KustomizeConfig> {
...
-    return {succeeded: true, result: config};
- }
- 
+
+    return { succeeded: true, result: config };
+}
+
+export function getKubectlGadgetConfig(): Errorable<KubeloginConfig> {
+    const kubeloginConfig = vscode.workspace.getConfiguration('azure.kubectlgadget');
+    const props = combine([getConfigValue(kubeloginConfig, 'releaseTag')]);
```
peterbom, line 58: Just noticed something I failed to spot in the `getKubeloginConfig` function, that also applies here. It doesn't make sense to me that we're `combine`-ing the results of an array containing a single item...

Would it work if we changed both this and the kubelogin function to:
```javascript
const config = getConfigValue(kubeloginConfig, 'releaseTag');
if (failed(config)) {
return {
    succeeded: false,
    error: `Failed to read azure.kubectlgadget configuration: ${props.error}`
};

return { succeeded: true, result: config.result };
```

src/commands/utils/helper/binaryDownloadHelper.ts
```diff
@@ -0,0 +1,89 @@
+import * as download from '../download/download';
```
peterbom, line 1: I've got more ideas about refactoring this. Can I do a PR to your PR?

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,173 @@
...
+    kubectl: k8s.APIAvailable<k8s.KubectlV1>
+): Promise<void> {
+    const clustername = cloudTarget.name;
+
+    switch (gadget) {
+        case InspektorGadget.Deploy:
+            await deployGadget(clustername, kubeconfig, kubectl);
+            return;
+        case InspektorGadget.Undeploy:
+            const answer = await vscode.window.showInformationMessage(`Do you want to undeploy gadget in selected cluster?`, "Yes", "No");
```
peterbom, line 79: Would be good to specify the cluster name: `Do you want to undeploy gadget in ${clustername}?`

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    const kubectlGadgetPath = await getKubectlGadgetBinaryPath();
+    let kubetlGadgetBinaryPath = "";
+
+    if (failed(kubectlGadgetPath)) {
+        vscode.window.showWarningMessage(`Gadget path is not found ${kubectlGadgetPath.error}`);
+        return;
+    }
+
+    kubetlGadgetBinaryPath = kubectlGadgetPath.result;
```
peterbom, line 119: nit: maybe combine the declaration and assignment of `kubectlGadgetBinaryPath`, and make it a `const`? Could also move it closer to where it's used... Also, if we don't need the `console.log` on line 131, we could probably get away without this identifier, and just use `kubectlGadgetPath.result` instead?

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,175 @@
...
+                process.env.PATH = binaryPathDir + path.delimiter + process.env.PATH;
+            }
+
+            const kubectlresult = await tmpfile.withOptionalTempFile<Errorable<k8s.KubectlV1.ShellResult>>(
+                clusterConfig, "YAML", async (kubeConfigFile) => {
+                    return await invokeKubectlCommand(kubectl, kubeConfigFile, commandToRun);
+                });
+
+            if (failed(kubectlresult)) {
+                vscode.window.showWarningMessage(`Gadget command failed with following error: ${kubectlresult.error}`)
```
peterbom, line 146: `return` here, and skip the following `if (kubectlresult.succeeded)` check?

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            if (kubectlresult.succeeded) {
+                const webview = createWebView('AKS Kubectl Commands', `AKS Kubectl Command view for: ${clustername}`).webview;
+                webview.html = getWebviewContent(kubectlresult.result, command, extensionPath.result, webview);
+            }
+
+            return true;
```
peterbom, line 154: We can skip the return here, can't we? Because the function returns `Promise<void>` (i.e. no return value)

src/commands/utils/config.ts
```diff
@@ -37,22 +37,40 @@ export function getKustomizeConfig(): Errorable<KustomizeConfig> {
...
- 
-    if (failed(props)) {
-       return {
-          succeeded: false,
-          error: `Failed to read azure.kubelogin configuration: ${props.error}`
-       };
+    const config = getConfigValue(kubeloginConfig, 'releaseTag');
+
+    if (failed(config)) {
+    return {
```
peterbom, line 43: Looks as if indentation went awry here?

src/commands/utils/config.ts
```diff
@@ -37,22 +37,40 @@ export function getKustomizeConfig(): Errorable<KustomizeConfig> {
...
+
+    return { succeeded: true, result: configresult };
+}
+
+export function getKubectlGadgetConfig(): Errorable<KubeloginConfig> {
+    const kubectlGadgetConfig = vscode.workspace.getConfiguration('azure.kubectlgadget');
+    const config = getConfigValue(kubectlGadgetConfig, 'releaseTag');
+
+    if (failed(config)) {
+    return {
```
peterbom, line 61: indentation

peterbom (approved):
Thanks Tats - looks great!

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -0,0 +1,168 @@
...
+                });
+
+            if (failed(kubectlresult)) {
+                vscode.window.showWarningMessage(`Gadget command failed with following error: ${kubectlresult.error}`);
+                return;
+            }
+
+            if (kubectlresult.succeeded) {
+                const webview = createWebView('AKS Kubectl Commands', `AKS Kubectl Command view for: ${clustername}`).webview;
+                webview.html = getWebviewContent(kubectlresult.result, command, extensionPath.result, webview);
```
blanquicet, line 147: It should be `commandToRun`. In fact, it is showing `kubectl deploy` as command, instead of ``kubectl gadget deploy`:

```suggestion
                webview.html = getWebviewContent(kubectlresult.result, commandToRun, extensionPath.result, webview);
```

src/commands/utils/helper/kubectlGadgetDownload.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export async function getKubectlGadgetBinaryPath(): Promise<Errorable<string>> {
+    const releaseTag = await getLatestKubectlGadgetReleaseTag();
+    if (!releaseTag) {
+         return {succeeded: false, error: "Could not get latest release tag."};
+    }
+
+    const archiveFilename = getArchiveFilename(releaseTag);
+    const downloadUrl = `https://github.com/inspektor-gadget/inspektor-gadget/releases/download/${releaseTag}/${archiveFilename}`;
+    const pathToBinaryInArchive = getPathToBinaryInArchive();
```
blanquicet, line 25: I would suggest to not use a fixed version here and instead download always the latest. Otherwise, we will have to create a PR every time we release a new version (consider we do it monthly). 

We could use the GitHub API to retrieve the latest version (`tag_name` JSON field): [here](https://api.github.com/repos/inspektor-gadget/inspektor-gadget/releases/latest). And then, compose the URL for the `kubectl-gadget` binary, as you are doing right now. 

Of course, I would suggest to verify if we are using the latest version only for the `Deploy` operation, and not every time we run a new gadget.

blanquicet (commented):
Awesome!!! 🚀 💯 

I added some comments but it looks great. Thanks Tats for working on this!

src/commands/utils/helper/kubectlGadgetDownload.ts
peterbom: @blanquicet: we generally tie the VSCode extension to specific versions of all our dependencies, to make its behaviour as predictable as possible (although we allow users to override this in settings.json, which is mainly helpful for development purposes). We do quite frequent releases of the VSCode extension too, and it's trivial for us to bump the versions of our dependencies. If that does become too onerous, we could consider generating automated PRs to bump the IG version.

On a sort-of related note, this reminds me of another question I had (probably because I haven't read the docs properly, so feel free to point me there): if you run `gadget deploy` with one version, and then `gadget undeploy` with a later version, is that guaranteed to work? Or does it lead to the possibility of errors or orphaned resources?

I wouldn't be surprised if there will be circumstances in the future where we'd want to check whether IG is installed on the cluster, and perhaps the version (maybe we'd want to selectively enable/disable the deploy and undeploy commands, or give users the option to upgrade). Is there a recommended way of checking this?

blanquicet: > We do quite frequent releases of the VSCode extension too, and it's trivial for us to bump the versions of our dependencies. 
>
Ahh ok, great to know.

> If that does become too onerous, we could consider generating automated PRs to bump the IG version.
>
That would be great!

> if you run gadget deploy with one version, and then gadget undeploy with a later version, is that guaranteed to work? Or does it lead to the possibility of errors or orphaned resources? 
>
Yep, it is. There was only one version it wasn't because we changed the namespace where Inspektor Gadget was running. We notified it in the release notes and also in also slack. So, even if I don't expect to have another change that doesn't guarantee this, it will be for sure specified in the release notes, and we can decide how to behave in that case. Is it OK for you?

@mauriciovasquezbernal any comment?

> I wouldn't be surprised if there will be circumstances in the future where we'd want to check whether IG is installed on the cluster, and perhaps the version (maybe we'd want to selectively enable/disable the deploy and undeploy commands, or give users the option to upgrade). Is there a recommended way of checking this?
>
We could use the `version` command (NOTE: Maybe we could add a JSON output mode for that `version` command so that it is easier to parse the server version):

```bash
$ kubectl gadget version
Client version: v0.12.0
Server version: v0.12.0
```

But I would suggest to work on this only if we release a version that doesn't guarantee backward compatibility.

blanquicet (commented):
Few more comments after using it a bit 🙂

mauriciovasquezbernal:
It's fine to use "kubectl gadget undeploy" on an old version. We try to install almost everything to the "gadget" namespace, then it should be fine. As mentioned by Jose, if there's anything that changes we'll announce it on the release notes, but I don't expect this to happen.
________________________________

blanquicet:
I am addressing all my comments/suggestions in https://github.com/Azure/vscode-aks-tools/pull/195.

PR #196, opened 2023-02-14
Enable non-interactive Inspektor Gadget Commands.

Tatsinnit:
Thank you for taking time and reviewing this! This PR enables the few `Non-Interative` **inspector gadget command**. This PR capitalise the ground up work we did under #190

List of commands this drop enables are:

- Inspector gadget Top command
    - TCP
    - eBPF
    - BlockIO
    - File
- IG Snapshot
    - Process
    - Socket
- IG Profile
    -  CPU

Sample:

Screenshot 2023-02-15 at 10 06 39 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,15 @@
         "onCommand:aks.aksKubectlK8sLivezAPIEndpointCommands",
         "onCommand:aks.aksKubectlK8sReadyzAPIEndpointCommands",
         "onCommand:aks.aksInspektorGadgetDeploy",
-        "onCommand:aks.aksInspektorGadgetUnDeploy"
+        "onCommand:aks.aksInspektorGadgetUnDeploy",
+        "onCommand:aks.aksInspektorGadgetTopTCP",
+        "onCommand:aks.aksInspektorGadgetTopEBPF",
+        "onCommand:aks.aksInspektorGadgetTopBlockIO",
+        "onCommand:aks.aksInspektorGadgetTopFile",
+        "onCommand:aks.aksInspektorGadgetProfileBlockIO",
+        "onCommand:aks.aksInspektorGadgetProfileCPU",
+        "onCommand:aks.aksInspektorGadgetSnapshotProcess",
+        "onCommand:aks.aksInspektorGadgetSnapshotSocket"
     ],
     "main": "./dist/extension",
     "contributes": {
@@ -217,6 +225,38 @@
             {
                 "command": "aks.aksInspektorGadgetUnDeploy",
                 "title": "Undeploy"
+            },
+            {
+                "command": "aks.aksInspektorGadgetTopTCP",
+                "title": "Top Tcp"
+            },
+            {
+                "command": "aks.aksInspektorGadgetTopEBPF",
+                "title": "Top Ebpf"
+            },
+            {
+                "command": "aks.aksInspektorGadgetTopBlockIO",
+                "title": "Top BlockIO"
+            },
+            {
+                "command": "aks.aksInspektorGadgetTopFile",
+                "title": "Top File"
+            },
+            {
+                "command": "aks.aksInspektorGadgetProfileBlockIO",
+                "title": "Profile BlockIO"
+            },
+            {
+                "command": "aks.aksInspektorGadgetProfileCPU",
+                "title": "Profile CPU"
+            },
+            {
+                "command": "aks.aksInspektorGadgetSnapshotProcess",
+                "title": "Sanpshot Process"
+            },
+            {
+                "command": "aks.aksInspektorGadgetSnapshotSocket",
+                "title": "Sanpshot Socket"
             }
         ],
         "menus": {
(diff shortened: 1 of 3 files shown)


peterbom (approved):
lgtm!

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -101,6 +177,68 @@ async function unDeployGadget(
...
+    const command = "top block-io 30 --all-namespaces --timeout 30";
+
+    return await runKubectlGadgetCommands(clustername, command, clusterConfig, kubectl);
+}
+
+async function topFile(
+    clustername: string,
+    clusterConfig: string,
+    kubectl: k8s.APIAvailable<k8s.KubectlV1>) {
+    const command = "top file 30 --timeout 30";
```
blanquicet, line 211: Why you didn't add `--all-namespaces` for this gadget?

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -101,6 +177,68 @@ async function unDeployGadget(
...
+
+    return await runKubectlGadgetCommands(clustername, command, clusterConfig, kubectl);
+}
+async function profileCPU(
+    clustername: string,
+    clusterConfig: string,
+    kubectl: k8s.APIAvailable<k8s.KubectlV1>) {
+    const command = "profile cpu --timeout 30";
```
blanquicet, line 219: dito

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -101,6 +177,68 @@ async function unDeployGadget(
...
+
+    return await runKubectlGadgetCommands(clustername, command, clusterConfig, kubectl);
+}
+
+async function snapshotProcess(
+    clustername: string,
+    clusterConfig: string,
+    kubectl: k8s.APIAvailable<k8s.KubectlV1>) {
+    const command = "snapshot process --all-namespaces 30 --timeout 30";
```
blanquicet, line 228: snapshot gadgets don't require a time or timeout. They will last the time they need to take an snapshot of the processes or sockets in the system. So it should be simply:

```suggestion
    const command = "snapshot process --all-namespaces";
```

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -101,6 +177,68 @@ async function unDeployGadget(
...
+
+    return await runKubectlGadgetCommands(clustername, command, clusterConfig, kubectl);
+}
+
+async function snapshotSocket(
+    clustername: string,
+    clusterConfig: string,
+    kubectl: k8s.APIAvailable<k8s.KubectlV1>) {
+    const command = "snapshot socket --all-namespaces 30 --timeout 30";
```
blanquicet, line 237: dito

src/commands/aksInspektorGadget/aksInspektorGadget.ts
```diff
@@ -80,6 +135,27 @@ async function prepareInspektorGadgetInstall(
...
+            return;
+        case Command.ProfileCPU:
+            await profileCPU(clustername, kubeconfig, kubectl);
+            return;
+        case Command.SnapshotProcess:
+            await snapshotProcess(clustername, kubeconfig, kubectl);
+            return;
+        case Command.SnapshotSocket:
+            await snapshotSocket(clustername, kubeconfig, kubectl);
+            return;
```
blanquicet, line 158: Currently, if we try to run one of these gadgets without having deployed Inspektor Gadget before, it will fails with error: 

```
kubectl gadget command failed with following error: Kubectl returned error 1 for gadget snapshot socket --all-namespaces 30 --timeout 30
Error: Error: failed to run gadget: error creating trace: no gadget pods found. Is Inspektor Gadget deployed?
```

What if we verify that Inspektor Gadget is running and in case it is not, we ask the users if they want to deploy it? To verify it is running, you can use the `version` command:

```bash
# Not running
$ ./kubectl-gadget version
Client version: v0.12.1
Server version: not installed

# ./kubectl-gadget deploy
[...]

# Running
$ ./kubectl-gadget version
Client version: v0.12.1
Server version: v0.12.1
```

BTW, we could improve `version` command to allow using JSON format so that it is easier to parse:

```bash
$ ./kubectl-gadget version -o json
{"Client version": "v0.12.1", "Server version": "v0.12.1"}
```

package.json
```diff
@@ -52,7 +52,14 @@
         "onCommand:aks.aksKubectlK8sLivezAPIEndpointCommands",
```
blanquicet, line 52: I noticed that some gadgets as a wide output and it doesn't fit on the screen:

image

I was wondering if we could add a scrollbar because currently it is difficult to see the whole output.

blanquicet (commented):
It is awesome folks! Thanks for working on this. I added some comments/suggestions 🙂.

src/commands/aksInspektorGadget/aksInspektorGadget.ts
peterbom: @blanquicet - that's a great point, and definitely something we should be looking at.

I think a more general point is that the tree-view/command interface doesn't quite capture the state of the various resources deployed to the cluster. In an ideal world, only commands which make sense should be displayed to the users, so it'd be preferable if only 'deploy' were available in the case where IG is not deployed.

I think it'd be _possible_ to selectively display commands based on querying the state of the cluster, but it'd require running background queries to poll all the clusters in the explorer view, which is a performance trade-off we've been very wary of making.

An alternative approach that @Tatsinnit and I have been chatting about is to use a webview. There are possible accessibility concerns here, which will need a bit more investigation. But it could solve a few issues that have been raised so far.

The idea (at least in my head) is, you'd have a single 'Inspektor Gadget' command for the cluster, which would launch a webview containing a single button to deploy it (if it's not already deployed). Once deployed, it could have the options to run the various commands, along with whatever filtering inputs we wanted. I think it could support dynamic commands (like 'top') where the output is periodically updated. And there may be a neat way of managing traces, such that you could create multiple long-running ones, and see them listed and interact with them, even if you close and re-open the webview (because they're actual k8s resources, right?).

blanquicet: > I think a more general point is that the tree-view/command interface doesn't quite capture the state of the various resources deployed to the cluster. In an ideal world, only commands which make sense should be displayed to the users, so it'd be preferable if only 'deploy' were available in the case where IG is not deployed.
> 
I got your point, but instead of not displaying them at all, I would make them "non-clickable" (grayed out). Otherwise, the users will have to deploy IG just to know what gadgets are available. I think that, sometimes, users could be just curious of exploring what options a new feature offers without installing it.

> The idea (at least in my head) is, you'd have a single 'Inspektor Gadget' command for the cluster, which would launch a webview containing a single button to deploy it (if it's not already deployed). Once deployed, it could have the options to run the various commands, along with whatever filtering inputs we wanted. I think it could support dynamic commands (like 'top') where the output is periodically updated.
>
I like this idea. I would give us more control on what the users can and can't do.

> And there may be a neat way of managing traces, such that you could create multiple long-running ones, and see them listed and interact with them, even if you close and re-open the webview (because they're actual k8s resources, right?).
>
We offer this functionality but it is managed by the `kubectl-gadget` CLI. It mainly translate the CLI arguments in a k8s resource (trace CR) to start tracing with the given set of paramters, but it will also delete such resource when you close the CLI. However, you could not use the CLI and manage the trace CR yourself to start/stop tracing, but of course, it is more complex.

peterbom: Thanks @blanquicet! I've tried to capture this [here](#203), because it's a bit of a bigger conversation than this PR. Please add to it if I've missed anything, or you'd like to add your own ideas.

PR #205, opened 2023-02-23
Remove unnecessary kubelogin download

peterbom:
I think we added this for testing purposes when enabling kubelogin for AAD clusters with local users disabled.

It's not needed because the common code for retrieving the kubeconfig takes care of this.


src/commands/aksKubectlCommands/aksKubectlCommands.ts
```diff
@@ -130,10 +129,6 @@ async function loadKubectlCommandRun(
   kubectl: k8s.APIAvailable<k8s.KubectlV1>) {
 
   const clustername = cloudTarget.name;
-  // Download and get Kubelogin path for AAD Enabled AKS Cluster.
-  // Sample use
-  const kubeloginPath = await getKubeloginBinaryPath();
```
Tatsinnit, line 135: 💡Sounds good, Adding for archival reasons: `getAadKubeconfig` is declared under `cluster.ts` based on `aadProfile` property.

Tatsinnit (approved):
🎊☕️ Thank you for adding for review looks good to me, we should get this PM tested, but looks nice, thank you.

PR #206, opened 2023-03-29
Allow development of VS Code-themed webviews using UI toolkit and React framework

peterbom:
**What this is for**

To start addressing the following areas for improvement with Webviews:
- They are cumbersome to develop, involving substantial boilerplate and message plumbing
- It is hard to theme them consistently, both with each other and with the VS Code host environment
- They don't react to theme changes in the host (e.g. switching between dark and light modes)
- We have a lot of work to do to improve accessibility in our Webviews
- It takes a long time to make code changes and see them reflected in the Webviews, making it hard to iterate
- They use our own unique rendering approach and Handlebars helpers which make it harder to port views that we might want to create our own VSCode-themed versions of (e.g. Inspektor Gadget React components in Headlamp)

**How it helps**

This introduces a sub-project that allows development of Webviews as components in a React application. The application can be run in a development web server with continuous update and debugging in a browser, with browser-based mocking of the VS Code backend for message passing.

Some features of this are:
- Reduced boilerplate: not creating an entire HTML document for each Webview, and extracting generic message-passing logic to shared utility types.
- Strong typing of the 'contract' between VS Code and the Webviews, by leaning heavily on TypeScript to check the types of data passed both ways.
- Use of CSS variables to match VS Code [theme colours](https://code.visualstudio.com/api/references/theme-color), allowing instantaneous reflection of colour scheme changes.
- Use of the [VS Code Webview UI Toolkit](https://github.com/microsoft/vscode-webview-ui-toolkit) to display themed and accessible components.
- Near real-time update of rendered views during development.
- Easy creation of manual test scenarios, allowing views to reflect states that might be harder to achieve using a real VS Code backend (e.g. error scenarios).
- A more 'standard' Framework for web UI development, making it easier to port other React components or try out new ideas quickly.

**What is actually changed**

There are a lot of existing Webviews in the extension. As an initial exercise to prove the utility of these changes, this PR includes the conversion one of the more complex ones to use the web application; that is the Periscope deployment result Webview.

This previously rendered something like this:

periscope-old

It is now showing like this (with theme switching between dark and light modes):

periscope-new

There is also a `TestStyleViewer` component, which is not displayed from any VS Code command, but is used in a VS Code test to extract the VS Code CSS variables and default Webview styling so that they can be used to style the React application so that it's consistent with VS Code.


Tatsinnit:
👍 /hallway pass for the CHAT-Gpt stuff B-) because of its token number bug, nothing to do with the PR.

.vscode/launch.json
```diff
@@ -27,12 +27,20 @@
             "runtimeExecutable": "${execPath}",
             "args": [
                 "--extensionDevelopmentPath=${workspaceFolder}",
-                "--extensionTestsPath=${workspaceFolder}/out/test"
+                "--extensionTestsPath=${workspaceFolder}/out/tests/suite/index"
```
peterbom, line 30: This fixes an existing bug when launching tests using VS Code debugging tools (F5), namely the error:
```
Error: Cannot find module 'c:\code\vscode-aks-tools\out\test'
Require stack:
- c:\Users\pebomber\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\loader.js
- c:\Users\pebomber\AppData\Local\Programs\Microsoft VS Code\resources\app\out\bootstrap-amd.js
- c:\Users\pebomber\AppData\Local\Programs\Microsoft VS Code\resources\app\out\bootstrap-fork.js
```

It's kind of nice to get this fixed because, for reasons I don't understand:
- if I run tests from the command line (`npm run test`) VS Code complains when there's another VS Code process running and refuses to launch the test host.
- if I run tests this way (F5), it happily opens a new host regardless of other VS Code processes running.

Plus, this allows us to debug tests through the IDE, with breakpoints and stepping, etc.

.vscode/tasks.json
```diff
@@ -4,19 +4,34 @@
     "version": "2.0.0",
     "tasks": [
         {
+            "label": "dev:webview",
             "type": "npm",
-            "script": "watch",
+            "script": "dev:webview",
```
peterbom, line 9: As far as I could tell, the `watch` task was no longer being used (since the move to webpack), so I figured it was safe to overwrite this rather than creating another task. @Tatsinnit - let me know if I'm wrong about that.

package.json
```diff
@@ -522,13 +522,17 @@
...
+        "dev:webview": "cd webview-ui && npm run dev",
+        "build:webview": "cd webview-ui && npm run build",
+        "vscode:prepublish": "npm run webpack",
+        "webpack": "npm run build:webview && webpack --mode production",
         "webpack-dev": "webpack --mode development --watch --info-verbosity verbose",
-        "test-compile": "tsc -p ./",
+        "test-compile": "npm run build:webview && tsc -p ./",
         "compile": "webpack --mode none",
         "watch": "tsc -watch -p ./",
-        "test": "mocha -V && npm run test-compile && node ./out/tests/runTests.js"
```
peterbom, line 531: I thought `mocha -V` was to check whether `mocha` is installed globally, and that it's not required because it's installed as a development dependency, so I removed it. But actually it worked fine for me before that change, even though I don't have it installed globally. I still _think_ calling `mocha -V` is not necessary, but I can add it back if it's there for a good reason I don't understand.

src/panels/BasePanel.ts
```diff
@@ -0,0 +1,129 @@
...
+        const initialStateJson = initialState ? JSON.stringify(initialState) : "";
+
+        // Tip: Install the es6-string-html VS Code extension to enable code highlighting below
+        return /*html*/ `
+        <!DOCTYPE html>
+        <html lang="en">
+            <head>
+                <meta charset="UTF-8" />
+                <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+                <meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src ${webview.cspSource} https://use.fontawesome.com; script-src 'nonce-${nonce}' https://use.fontawesome.com; font-src 'self' https://use.fontawesome.com">
```
peterbom, line 85: I think we can/should consume the [fontawesome npm package](https://www.npmjs.com/package/font-awesome) in the `webview-ui` project, rather than accessing an external URL here. But perhaps save that for another day.

webview-ui/src/Periscope/ErrorView.tsx
```diff
@@ -0,0 +1,39 @@
...
+export interface ErrorViewProps {
+    clusterName: string
+    message: string
+    config: PeriscopeTypes.KustomizeConfig
+}
+
+export function ErrorView(props: ErrorViewProps) {
+    return (
+        <>
+            <i className={[styles.errorIndicator, "fa", "fa-times-circle"].join(" ")}></i>
```
peterbom, line 14: I don't love this syntax, and I've ended up using it a few times in this PR. It might be worth using the [`classnames` package](https://github.com/JedWatson/classnames) to clean this up.

webview-ui/src/Periscope/ErrorView.tsx
```diff
@@ -0,0 +1,39 @@
...
+            AKS Periscope failed to run on '{props.clusterName}'. Please see the error below for more details.
+
+            <VSCodeDivider />
```
peterbom, line 17: The webviews used to contain collapsible panels. I was too lazy to try and reproduce those, and I'm not sure if people actually use that functionality (I certainly never did). So I opted for a simple divider, which is part of the UI toolkit and looks reasonably clean.

webview-ui/src/Periscope/NodeActions.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const shareableLink = `${props.containerUrl}/${props.runId}/${props.nodeName}/${props.nodeName}.zip${props.shareableSas}`;
+
+    function copyShareLink(e: React.MouseEvent) {
+        e.stopPropagation();
+        navigator.clipboard.writeText(shareableLink);
+    }
+
+    return (
+        <>
+            <VSCodeButton onClick={copyShareLink}>
```
peterbom, line 22: To reduce clutter here, I removed the `Get node logs` button, and instead you just select a row (which corresponds to a node) by clicking on it. Then information about the selected row (which now is just logs, but in theory it could be anything) is displayed below. To me that's intuitive, and gives us more scope to develop UI specific to each node without cluttering the table, but I'm open to further discussion.

webview-ui/package.json
```diff
@@ -0,0 +1,25 @@
+{
```
Tatsinnit, line 1: 💡 Comments please.

webview-ui/src/Periscope/NodeActions.tsx
```diff
@@ -0,0 +1,37 @@
+import { VSCodeButton, VSCodeLink } from "@vscode/webview-ui-toolkit/react";
```
Tatsinnit, line 1: 💡 Comment please for use of `tsx` and relevance.

src/panels/utilities/webview.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A helper function that returns a unique alphanumeric identifier called a nonce.
+ *
+ * @remarks This function is primarily used to help enforce content security
+ * policies for resources/scripts being executed in a webview context.
+ *
+ * @returns A nonce
+ */
+export function getNonce() {
+    const allowedChars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
```
Tatsinnit, line 29: 💡 Question please: Do we need to worry about internationalisation here?

Tatsinnit (approved):
☕ 👍 🥷🚀 Thank you so much for adding me to the review, looks good to me, really appreciate it, they are `50 files`, so lets get some good double checking from @rzhang628 please 🎊

Thank you

webview-ui/src/Periscope/ErrorView.tsx
peterbom: I'll submit a separate PR for this

src/commands/periscope/periscope.ts
```diff
@@ -127,29 +124,37 @@ async function runAKSPeriscope(
...
+        const dataProvider = PeriscopeDataProvider.createForDeploymentError(cluster.name, runId, runCommandResult.error, deploymentParameters);
+        panel.show(dataProvider);
+        return;
     }
 
-    const webview = createWebView('AKS Periscope', `AKS Periscope: ${clusterName}`).webview;
-    webview.html = getSuccessWebviewContent(extensionPath.result, runId, clusterName, clusterStorageInfo.result, nodeNames.result, webview);
-    handleMessages(webview, kubectl, clusterKubeConfig, clusterStorageInfo.result, 'aks-periscope', runId, nodeNames.result);
+    // Show the webview for successful deployment
+    const dataProvider = PeriscopeDataProvider.createForDeploymentSuccess(cluster.name, runId, nodeNames.result, deploymentParameters);
```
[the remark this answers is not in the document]
peterbom: Yep, you've nicely captured the trade-off here. Initially we had one webview and gave it a single data structure that accounted for all possible scenarios. We later split it into three webviews with different state data structures. This meant extra code, but it was easier to reason about each one.

This latest change goes back to the first model (single webview, single data structure), but with a couple of differences:
1. We can split the logic inside the webview into separate React components to handle the different scenarios.
2. We have separate factory functions for creating the state/message handling that are appropriate for each scenario.

src/panels/BasePanel.ts
```diff
@@ -0,0 +1,127 @@
...
+        const scriptUri = getUri(webview, extensionUri, ["assets", "index.js"]);
+
+        // See: https://github.com/microsoft/vscode-webview-ui-toolkit/blob/main/docs/getting-started.md#enable-webview-scripts-and-improve-security
+        const nonce = getNonce();
+
+        const initialStateJson = initialState ? JSON.stringify(initialState) : "";
+
+        // Tip: Install the es6-string-html VS Code extension to enable code highlighting below
+        return /*html*/ `
+        <!DOCTYPE html>
```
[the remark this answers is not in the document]
peterbom: Reasons for the current approach:
- This is how it's done in the webview UI toolkit ['getting started'](https://github.com/microsoft/vscode-webview-ui-toolkit/blob/main/docs/getting-started.md#add-the-button-to-the-webview-markup) guide.
- The [samples](https://github.com/microsoft/vscode-webview-ui-toolkit-samples) all use this approach too.
- Ideally, this string will rarely change (actually, I'm planning to change it in the very next PR, but after _that_ I'm hoping it'll be pretty static).
- This should stay short - it is just the shell of the HTML and all logic should be inside the `webview-ui` project. Putting this into a separate file might suggest it's a bigger thing than it is, and might tempt us to add more stuff into it that doesn't belong.

src/panels/utilities/webview.ts
```diff
@@ -0,0 +1,32 @@
...
+ * given file/resource.
+ *
+ * @param webview A reference to the extension webview
+ * @param extensionUri The URI of the directory containing the extension
+ * @param pathList An array of strings representing the path to a file/resource
+ * @returns A URI pointing to the file/resource
+ */
+export function getUri(webview: Webview, extensionUri: Uri, pathList: string[]) {
+    // Find the path under the 'dist' folder in the 'webview-ui' project.
+    pathList = ["webview-ui", "dist", ...pathList];
```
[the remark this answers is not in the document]
peterbom: Good pointing out stuff that should be in a README! Roughly:
1. Webview development takes place under `/webview-ui/src`.
2. When you build the `webview-ui` project, the bundled+minimized output ends up in `/webview-ui/dist`.
3. When you run the extension in both development and production (extracted from vsix), it loads the webview assets from `/webview-ui/dist`

You're right that this part was not part of the `vsix`. There's a separate change to `.vscodeignore` which ensures that these files _do_ get included.

webview-ui/src/Periscope/NodeActions.tsx
peterbom: @rzhang628 - feel free to weigh in here.

src/panels/utilities/webview.ts
peterbom: I think not, because these are only used to generate a [random string](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) which apparently prevents [replay attacks](https://en.wikipedia.org/wiki/Replay_attack) (I haven't thought through exactly how that exploit would work, but the docs say to add it, so I did). In any case, this is just used to compare against the same set of characters. It's not visible to users, or compared against any user input, and is unaffected by users' locale or language settings.

package.json
```diff
@@ -522,13 +522,16 @@
         ]
     },
     "scripts": {
-        "vscode:prepublish": "webpack --mode production",
-        "webpack": "webpack --mode production",
+        "install:all": "npm install && cd webview-ui && npm install",
+        "dev:webview": "cd webview-ui && npm run dev",
+        "build:webview": "cd webview-ui && npm run build",
+        "vscode:prepublish": "npm run webpack",
+        "webpack": "npm run build:webview && webpack --mode production",
```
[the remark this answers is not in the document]
peterbom: A summary of all npm commands is now under /docs

src/panels/utilities/webview.ts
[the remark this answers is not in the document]
peterbom: Added a comment about this to the `webview-development` readme.

webview-ui/package.json
peterbom: As mentioned somewhere else, you can't add comments to this file, but I think everything is covered in `/docs/package-scripts` and `/docs/webview-development`.

webview-ui/src/Periscope/NodeActions.tsx
peterbom: I won't comment in the code, because these files are all over the place, but it [indicates to the typescript compiler](https://www.typescriptlang.org/docs/handbook/jsx.html) that these files should be compiled in a specific way, to handle the inline `<...>` elements.

.vscode/tasks.json
peterbom: Actually it's still there - I was confused by the diff!

webview-ui/src/Periscope/NodeActions.tsx
rzhang628: I like the new UI, much more convenient for a lazy me lol. I am good with approving this one :)

PR #207, opened 2023-04-17
Add webview component for detectors

peterbom:
This re-implements all the detectors using the `webview-ui` React project.

Previously, each detector was in its own directory under the root. Now, because the extension code for launching a detector is simpler, the extra folder structure isn't really needed. So instead, all detectors have been moved to a single file, `src/commands/detectors/detectors.ts`.

All 'special case' logic for detectors has also been removed (including Handlebars helpers of course, because they're being rendered in React now). This fixes some display issues, like missing data in `SNAT Connection and Port Allocation`.

There are a *lot* of new files, but the majority are `.json` files representing ARM responses, which are used for manually testing during development, allowing us to quickly see the output for every single detector. These are not included in the built output or packaged in the extension.

Below are some 'before and after' screenshots, showing slightly improved styling, as well as the light- and dark-mode appearances.

**Before**
image
image
image

**After**
image
image
image
image


Tatsinnit:
💡☕️ Nice and decent size improvement, I wonder if we can quickly archive the features which will get directly and indirectly gets impacted by this, so that we can test them via vsix.  FYI to @rzhang628 as well. ☕️🙏❤️

Tatsinnit (approved):
🎊☕️

PR #208, opened 2023-04-26
Update Periscope deployment to latest version

peterbom:
This accounts for a change in the way Periscope is deployed in the latest version. Because the image tag refers to a manifest containing images for each OS, the tag is now the same for both Linux and Windows deployments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,13 +80,13 @@
                 },
                 "aks.periscope.releaseTag": {
                     "type": "string",
-                    "default": "0.0.11",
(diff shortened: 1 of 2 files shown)


Tatsinnit (approved):
❤️🎉☕️ thank you for this, looks good to me, shall we kindly get @rzhang628 eyes for quick test. Thanks all. 🙏

rzhang628:
Tested and it looks good to me :)

PR #217, opened 2023-07-17
Feature/create cluster

Tatsinnit:
This PR is to enable 2 things, which will be amazing for next level of AKS Developer Experience we could aim for. ❤️☕️

* Multi Step Wizard Mechanism.
* Minimalistic `Create Standard AKS Cluster` 

Details: (Please also refer to screenshots below)

The `Create Cluster` will appear as a subscription level `subMenu`, the key thing about this feature drop is that this feature is short, needs least input form the user and all user need to add is the **valid cluster name** and in step-2 select the **Predefined resource group name** form the drop down.

With this experience we are enhancing the developer experience and now fully one-click functionality to create standard cluster with just 2 steps from VsCode.

User will `right-click` on their subscription in the tree note, select `Create Standard Cluster` --> which will initiate a 2 step wizard - Ste-1 : Will accept valid cluster name and Step-2 : Select Resrouce Group Name - as a location the location of the resource group name will be taken as the location of the cluster and all user need to do is now setback and let VsCode experience do its thing call the aks managed operation and in the end user gets served with message box with click -> `nav to the portal cluster page` which will take user to the exact cluster which is created.

Thanks all, @peterbom , @rzhang628 , @qpetraroia and fyi : @gambtho ❤️☕️ 🎊

This kind of originated from this discussion here: https://github.com/Azure/vscode-aks-tools/discussions/204 

workflow:

Screenshot 2023-07-18 at 11 09 34 AM

Screenshot 2023-07-18 at 11 10 24 AM

Screenshot 2023-07-18 at 11 10 41 AM

Screenshot 2023-07-18 at 12 32 44 PM

Screenshot 2023-07-06 at 4 58 27 PM

### DIFF
--- a/src/commands/aksCreateCluster/aksCreateCluster.ts
+++ b/src/commands/aksCreateCluster/aksCreateCluster.ts
@@ -1,25 +1,25 @@
 import { QuickPickItem } from 'vscode';
-import { MultiStepInput } from '../../multistep-helper/multistep-helper';
+import { createInputBoxStep, createQuickPickStep, runMultiStepInput } from '../../multistep-helper/multistep-helper';
 import { IActionContext } from "@microsoft/vscode-azext-utils";
 import { getAksClusterSubscriptionItem, getContainerClientFromSubTreeNode, getResourceGroupList } from '../utils/clusters';
-import { failed } from '../utils/errorable';
+import { Errorable, failed } from '../utils/errorable';
 import * as vscode from 'vscode';
 import * as k8s from 'vscode-kubernetes-tools-api';
 import SubscriptionTreeItem from '../../tree/subscriptionTreeItem';
 import { ResourceIdentityType } from '@azure/arm-containerservice';
 import { longRunning } from '../utils/host';
-import { ResourceGroupsListNextResponse } from '@azure/arm-resources/esm/models';
+import { ResourceGroup } from '@azure/arm-resources/esm/models';
 const meta = require('../../../package.json');
 
-
 interface State {
-    title: string;
-    step: number;
-    totalSteps: number;
-    resourceGroup: QuickPickItem;
-    name: string;
+    resourceGroup: ResourceGroupItem;
     clustername: string;
     subid: string | undefined;
+}
+
+interface ResourceGroupItem extends QuickPickItem {
+    name: string;
+    location: string;
 }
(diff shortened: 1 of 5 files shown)


qpetraroia:
Looks great @Tatsinnit, is there a vsix I can test?

qpetraroia (approved):

gambtho (approved):

src/commands/aksCreateCluster/aksCreateCluster.ts
```diff
@@ -0,0 +1,177 @@
...
+
+interface State {
+    title: string;
+    step: number;
+    totalSteps: number;
+    resourceGroup: QuickPickItem;
+    name: string;
+    clustername: string;
+    subid: string | undefined;
+    runtime: QuickPickItem;
```
peterbom, line 23: I think we can remove this now, can't we?

src/commands/aksCreateCluster/aksCreateCluster.ts
```diff
@@ -0,0 +1,176 @@
...
+        });
+
+        state.resourceGroup = pick;
+    }
+
+
+    function shouldResume() {
+        // Could show a notification with the option to resume.
+        return new Promise<boolean>((resolve, reject) => {
+            // noop
```
peterbom, line 90: I think this will never complete. This gets called when an input (text/dropdown) is cancelled (`onDidHide`). So what actually happens is the main Promise in `showInputBox` and `showQuickPick` never completes, and the `finally` block never gets hit, and the disposables are never cleaned up.

We _could_ change it to `return Promise.resolve(false);`... But then you run into another problem. `MultiStepInput.run()` completes without an error, even when the user escapes out, and we go on to try and create the cluster, even though the user didn't want to.

I can see this code came directly from the Microsoft [samples](https://github.com/Microsoft/vscode-extension-samples/blob/main/quickinput-sample/src/multiStepInput.ts), but I think that's where the real problem is. Let me make some tweaks...

Tatsinnit: Perfect!! Pete and I had conversation already and this is something we will rearrange code for, Thanks heaps Pete.

peterbom (approved):

PR #221, opened 2023-07-26
Upgrade ARM resources library

peterbom:
This upgrades the `@azure/arm-resources` library to the latest version, which has support for creating a client with a specified ARM endpoint (as we do for the Container Services client).

This cleans up some client code because the newer libraries have better support for paged enumeration of resources. It also identified places where we were creating these clients directly rather than using our library function, so we are no always creating them in the same way.

The `clusters.ts` file is also cleaned up a little, so that the same functions can be called to create both `ResourceManagerClient` and `ContainerServicesClient` objects, without needing separate functions depending on whether you're starting from a subscription or a cluster.


Tatsinnit (approved):
🎊❤️☕️💡

PR #225, opened 2023-08-31
Feature welcome workflow

Tatsinnit:
This PR is intending to introduce a very early version of AKS Walkthrough in VsCode.

Once, Mark done, user can reaches these by `shift + cmd + p` and using `welcome: walkthrough` as selection to go through list of walkthroughs available.

The mov below might be able to emphasise on the behaviour.

This will be a great fyi to many users who wonder about the list of features this extension provide which we have documented well in gh docs but its outside the vscode experience, hence bridging that gap.

Thank you so much in advance. ❤️🙏☕️

https://github.com/Azure/vscode-aks-tools/assets/6233295/564b6891-4f69-405d-bb3b-44d78eb3e423

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
         "walkthroughs": [
             {
               "id": "walkthrough",
-              "title": "Welcome to AKS Extesion",
-              "description": "Azure Kubernetes Service (AKS) Extension for Visual Studio Code helps enable AKS developers with one-click quick to deploy features with in vscode eco-system.",
+              "title": "Welcome to AKS Extension",
+              "description": "Azure Kubernetes Service (AKS) Extension for Visual Studio Code helps enable AKS developers with single click feature deployment within VS Code ecosystem.",
               "media": {
                 "image": "./docs/book/src/features/features.md",
                 "altText": "AKS Extension"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -36,34 +36,51 @@
...
+                      "markdown": "./docs/book/src/features/features.md"
+                    }
+                },
                 {
                   "id": "runcommand",
                   "title": "Cluster Operation Command",
-                  "description": "This step will run a command and check off once it has been run.\n[Run Command](command:getting-started-sample.runCommand)",
-                  "media": { "image": "media/image.png", "altText": "Empty image" },
-                  "completionEvents": ["onCommand:getting-started-sample.runCommand"]
+                  "description": "AKS Managed Cluster Operations supported by this extension, like create cluster, delete, start or stop cluster from show properties page, rotate cluster certificate, etc.",
```
qpetraroia, line 55: Instead of the word "like", can we use "such as" 

"such as create/delete cluster, start/stop cluster, show cluster properties, rotate cluster certificates, etc"

package.json
```diff
@@ -36,34 +36,51 @@
...
-                  "media": { "image": "media/image.png", "altText": "Empty image" },
-                  "completionEvents": ["onCommand:getting-started-sample.runCommand"]
-                },
-                {
-                  "id": "changesetting",
-                  "title": "Change Setting",
-                  "description": "This step will change a setting and check off when the setting has changed\n[Change Setting](command:getting-started-sample.changeSetting)",
-                  "media": { "markdown": "media/markdown.md" },
-                  "completionEvents": ["onSettingChanged:getting-started-sample.sampleSetting"]
+                  "description": "This extension provides a set of commands to help developers to troubleshoot AKS clusters. Like quick run AKD Diagnostics using detectors, run AKS Periscope, run Inspektor Gadget, etc.",
```
qpetraroia, line 71: Remove "like" for "such as"

package.json
```diff
@@ -36,34 +36,51 @@
...
+                    "markdown": "./docs/book/src/features/manage-cluster-operations.md"
+                  }
                 },
                 {
                   "id": "developercommand",
                   "title": "Developer Command",
-                  "description": "This step will run a command and check off once it has been run.\n[Run Command](command:getting-started-sample.runCommand)",
-                  "media": { "image": "media/image.png", "altText": "Empty image" },
-                  "completionEvents": ["onCommand:getting-started-sample.runCommand"]
+                  "description": "This extension provides a set of commands to help developers to run kubectl commands against users AKS clusters.",
```
qpetraroia, line 63: Maybe remove "users" 

"This extension provides a set of commands to help developers run kubectl commands against AKS clusters" 

I think this sounds more flush

src/extension.ts
```diff
@@ -48,7 +48,10 @@ export async function activate(context: vscode.ExtensionContext) {
         context.subscriptions.push(uiExtensionVariables.outputChannel);
 
         registerAzureUtilsExtensionVariables(uiExtensionVariables);
-
+        vscode.commands.executeCommand(
+            'workbench.action.openWalkthrough', 
+            { category: 'ms-kubernetes-tools.vscode-aks-tools#walkthrough'}
```
hsubramanianaks, line 53: Tats, everything looks good. Instead of keeping this as generic name "walk through" do we want to name it as "aks tools : walk through" or "aks tools: welcome page" something tied to "aks" ? Please let me know if this make sense. Thanks.

hsubramanianaks (approved):
@Tatsinnit looks good. Only kind suggestion would be to have id: walkthrough as id: aks-tools.walkthrough something specific to aks. Thanks for doing this.

hsubramanianaks (approved):

src/extension.ts
Tatsinnit: Thank you for this, I see what you mean I could simplify to make it clear essentially the way this thing gets mapped it’s the combination of “publishername.extensionid#id-of-thewalkthrough-tag” but instead of string I could read this through meta tag and make that more clear. So, essentially in this the initial objects are not changeable in name but I will play around and I’ll see if I can simplify those sentiments using meta information ℹ️☕️🥷

hsubramanianaks: As discussed, let's keep it as ```walkthrough``` as id, my intention was to make this searchable but seems like it wouldn't help. Thanks for the discussion.

package.json
```diff
@@ -31,6 +31,60 @@
     ],
     "main": "./dist/extension",
     "contributes": {
+        "walkthroughs": [
+            {
+              "id": "walkthrough",
+              "title": "Welcome to AKS Extesion",
```
peterbom, line 37: `Extesion` => `Extension`

package.json
```diff
@@ -31,6 +31,60 @@
...
+              "description": "Azure Kubernetes Service (AKS) Extension for Visual Studio Code helps enable AKS developers with one-click quick to deploy features with in vscode eco-system.",
```
peterbom, line 38: Maybe something like: "with single click feature deployment within VS Code ecosystem." ?

package.json
```diff
@@ -31,6 +31,60 @@
...
+                    "id": "summarycommand",
+                    "title": "All Features",
+                    "description": "Once you successfully log in with your Azure Account, you can view all AKS clusters in your Azure subscriptions(s) under the section named **Azure**. You can right click on your AKS cluster and click a menu item to perform following actions.",
+                    "media": { 
+                      "markdown": "./docs/book/src/features/features.md"
+                    }
+                },
+                {
+                  "id": "runcommand",
+                  "title": "Cluster Operation Command",
```
peterbom, line 54: Were we going to change these to "Commands" rather than "Command"?

package.json
```diff
@@ -31,6 +31,60 @@
...
+                  }
+                },
+                {
+                  "id": "quickdeploymentcommand",
+                  "title": "Quick deployments",
+                  "description": "This extension also provides quick deployment features to help developers to quick deploy applications to AKS clusters such as inspektor gadget, azure service operator, etc.",
+                  "media": {
+                    "markdown": "./docs/book/src/features/install-azureserviceoperator.md" 
+                     
+                }
```
peterbom, line 83: super minor: closing brace indentation

peterbom (approved):

qpetraroia (approved):

Issue #222, opened 2023-08-23
Ability to create a shared dotfile of custom commands for collaboration


rahulrai-in:
**Is your feature request related to a problem? Please describe.**
Currently, the tool is opinionated about the commands that developers want to execute. As a developer, I would like to maintain a library of commands that I can easily execute from the extension. Also, the default commands should be configurable. For example, I prefer to include the "-o wide" parameter when using the `kubectl get` commands.

**Mention what platform you want to support the new feature**
VSCode

**Describe the solution you'd like**
A dotfile, preferably named `.aks` that I can commit to the code repo with a dictionary of custom commands and options to modify the behavior of existing ones.

**Describe alternatives you've considered**
This is a common method used by many tools to customize their behavior and hence I did not consider other alternatives.

**Additional context**
An advantage of using dotfiles is that they can be easily distributed to the wider community through popular repositories: https://github.com/webpro/awesome-dotfiles

rahulrai-in:
Tagging @nellysattari for additional recommendations.

Tatsinnit:
Hiya, I am moving this to the discussion forum now, given this is now U/X'd well with new design (#226 ), which is well improved, lets get the feelers for how many use-cases will need specific online file support.

#226 - was the drop of the pull request which did re-design for the this `kubectl` commands for helpful developer experience.

Thank you so much.

PR #226, opened 2023-09-04
Feature customise kubectl command

Tatsinnit:
This is moving forward PR which is written with an intent to make part way progress on #222 

As part 1 - of this - the PR enables any user input kubectl command against their `aks` cluster. 

As a build up work, next we can do is to build a `kubectlfile.aks` file where user can add their list of `kubectl` commands and we can add a parser to hook so that we can read the input of that file and execute and record result for the user.

Current custom behaviour will look like this. 🙏☕️❤️ thank you so much as always.

Screenshot 2023-09-04 at 11 38 35 AM

Screenshot 2023-09-04 at 11 43 44 AM

Screenshot 2023-09-04 at 1 42 54 PM

Screenshot 2023-09-04 at 11 44 16 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -267,6 +267,10 @@
             {
                 "command": "aks.createCluster",
                 "title": "Create Standard Cluster"
+            },
+            {
+                "command": "aks.aksCustomKubectlCommand",
+                "title": "Run Custom Kubectl Command"
             }
         ],
         "menus": {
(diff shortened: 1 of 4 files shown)


peterbom (approved):
I like how this is done with minimal changes by reusing the existing code for showing the results.

In the future I think it'd make sense to build the custom commands into the existing webview...but that'd require more changes so let's keep it simple for now.

PR #227, opened 2023-09-05
Run Inspektor Gadget commands from webview

peterbom:
The outcome of discussion #203.

Replaces all the previous Inspektor Gadget commands in the nested context menu with a single 'Show Inspektor Gadget' command, which launches a webview providing all the functionality previously supplied by the individual commands.

Example view:
ig-webview-4

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
                 "js-yaml": "^3.14.0",
                 "move-file": "^3.0.0",
                 "neverthrow": "^4.3.0",
+                "rxjs": "^7.8.0",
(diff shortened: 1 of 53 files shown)


Tatsinnit (approved):

Tatsinnit:
❤️🙏 thank you so much for this, bunch of very cool changes, specially the new look and feel.

for archiving sake, we chatted regularly and I see the changes are pretty good to land in terms of telemetry instead of fine grained we will achieve more cumulative approach which is great because there was no advantage we have previously to actually finegrain. ☕️🎉 let’s get all this tested.

PR #235, opened 2023-09-29
Add common OpenAI code with example usage in kubectl webview

peterbom:
Not to be merged, for now.

Opening this as an example of how we might integrate OpenAI logic into our webviews.

This splits the hackathon work into common files/components that could be reused in different scenarios, as well as an example of how it could be incorporated into the `kubectl` webview.


hsubramanianaks:
Closing this PR as the person left.

PR #239, opened 2023-10-03
Fix comon id issue.

### LINKED ISSUE #236: Investigate why the welcome page disappears.

Tatsinnit:
This PR fixes #236 thank you so much all, cc: @qpetraroia and @peterbom 

As you can see, it was common `id` which was getting over written, also the media tag is not correct.

Thanks heaps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,11 +33,11 @@
     "contributes": {
         "walkthroughs": [
             {
-                "id": "walkthrough",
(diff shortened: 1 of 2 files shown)


hsubramanianaks (approved):
looks good to me.

peterbom (approved):

PR #240, opened 2023-10-04
Update ASO version and move to webview-ui project

peterbom:
This updates the version of ASO we install to the latest (v2.0).

The bulk of the changes are to move the UI to the React `webview-ui` project, which makes further changes for user interaction much easier.

See example screenshots below (the error is due to throttling - it did install successfully on a previous attempt):

Screenshot 2023-10-04 162743
Screenshot 2023-10-04 162836
Screenshot 2023-10-04 163030


src/panels/AzureServiceOperatorPanel.ts
```diff
@@ -0,0 +1,264 @@
...
+                commandResults: [],
+                cloudName: servicePrincipalAccess.result.cloudName as AzureCloudName,
+                subscriptions: servicePrincipalAccess.result.subscriptions,
+                tenantId: servicePrincipalAccess.result.tenantId
+            }
+        });
+    }
+
+    private async _handleInstallCertManagerRequest(webview: MessageSink<ToWebViewMsgDef>): Promise<void> {
+        const asoCrdYamlFile = "https://github.com/jetstack/cert-manager/releases/download/v1.12.1/cert-manager.yaml";
```
Tatsinnit, line 80: 💡 Observation ☕️☕️🥷 - I wonder what is the difference and if raw version of this file is there via `raw.gh url`  https://github.com/cert-manager/cert-manager/releases/download/v1.12.1/cert-manager.yaml

Tatsinnit (approved):
<3

src/panels/AzureServiceOperatorPanel.ts
peterbom: This URL is taken from the [ASO installation instructions](https://azure.github.io/azure-service-operator/#installation).

I _think_ the `raw.githubusercontent` URLs are for files which exist within the repo. The `yaml` file referred to here is a published asset (so it's built from the repo but may not be present in there).

Interesting that there is a redirect from `https://github.com/jetstack/cert-manager` to `https://github.com/cert-manager/cert-manager`, though. Maybe the ASO docs should be updated, but for now I think we can stick as closely as possible with the published docs; we can easily update this later if necessary.

Tatsinnit: Sounds good to me🍿 might be add the official docs link here and then resolve conversation for archival sake.

peterbom: Added

PR #241, opened 2023-10-05
Implement Create Cluster as WebView

peterbom:
This replaces the multi-step wizard with a webview.

The command name is also changed from 'Create Standard Cluster' to 'Create Cluster from VS Code'. I think the original naming was to support different cluster 'presets', but those can be implemented within the webview itself so there's no need to add extra context menu items.

It also supports creating the cluster in a resource group that doesn't yet exist (which would previously have necessitated a trip to the portal).

Screenshot 2023-10-05 164540<br/>
Screenshot 2023-10-05 164621<br/>
Screenshot 2023-10-05 164720<br/>
Screenshot 2023-10-05 164800<br/>
Screenshot 2023-10-05 164809<br/>
Screenshot 2023-10-05 164523<br/>


package.json
```diff
@@ -214,6 +214,10 @@
                 "command": "aks.createClusterNavToAzurePortal",
                 "title": "Create Cluster From Azure Portal"
             },
+            {
+                "command": "aks.createCluster",
+                "title": "Create Cluster From VS Code"
```
Tatsinnit, line 219: 💡 purely for idea sake, shall we rename this to whatever the default configuration and sky it uses to create this test cluster. (AKA dev/test - or something ) I am not sure what is the default configuration eludes to.

src/multistep-helper/multistep-helper.ts
```diff
@@ -1,187 +0,0 @@
-import * as vscode from 'vscode';
```
Tatsinnit, line 1: 👍 fyi @qpetraroia  this part of the multi-step will be gone after this change.

Tatsinnit (approved):
I quite like this given the longe term objective this tends to solve, aka, more custom based create cluster options, less of leaving the vscode idea.

package.json
hsubramanianaks: I like Create Cluster from VS Code :)

Tatsinnit: Totally sounds good as name - I think I was keen to point out if either on webview or somehow we can be clear to state that - hey it creates some baseline as default - if and only if there is a way to summarise that in 3 to 4 words in a sentence. 💡 (so the key action is clearly conveyed as to what kind of default create it will do) 

will leave that as chain of thought here.

qpetraroia:
Can you share a vsix? @peterbom ?

package.json
peterbom: I think adding something to the webview would be best, since the idea is that if we create more types/presets they will be selectable on the webview. I'll try to add something.

peterbom: image

peterbom:
@qpetraroia:
> Can you share a vsix? @peterbom ?

Can you try this? Delete the `.zip` extension (don't unzip).
[vscode-aks-tools-1.3.14-20231006-01.vsix.zip](https://github.com/Azure/vscode-aks-tools/files/12825028/vscode-aks-tools-1.3.14-20231006-01.vsix.zip)

PR #243, opened 2023-10-05
Abort last operation on cluster.

Tatsinnit:
This PR enables the quick to `Abort Last Operation` managed cluster API via Vscode, bringing user experience feature closer to their favourite editor.

https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/abort-latest-operation?tabs=HTTP

Than you so much, fyi @peterbom, @gambtho, @hsubramanianaks  and @qpetraroia 

how it will work et. al. very handy command and as an upgrade in next drop we can integrate this in the `ClusterProperty` page.

Screenshot 2023-10-06 at 10 48 42 AM

Screenshot 2023-10-06 at 10 47 13 AM

Screenshot 2023-10-06 at 10 34 24 AM
Screenshot 2023-10-06 at 10 46 33 AM
 
Screenshot 2023-10-06 at 10 35 27 AM

Screenshot 2023-10-06 at 10 34 02 AM

### DIFF
--- a/src/commands/utils/clusters.ts
+++ b/src/commands/utils/clusters.ts
@@ -447,6 +447,20 @@ export async function deleteCluster(
     }
 }
 
+export async function abortLastOperationInCluster(
+    target: AksClusterTreeItem,
+    clusterName: string
+): Promise<Errorable<string>> {
+    try {
+        const containerClient = getContainerClient(target);
+        await containerClient.managedClusters.beginAbortLatestOperationAndWait(target.resourceGroupName, clusterName)
+
+        return { succeeded: true, result: "Abort last operation in cluster succeeded." };
+    } catch (ex) {
+        return { succeeded: false, error: `Error invoking ${clusterName} managed cluster: ${getErrorMessage(ex)}` };
+    }
+}
+
 export async function rotateClusterCert(
     target: AksClusterTreeItem,
     clusterName: string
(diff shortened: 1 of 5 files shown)


src/commands/utils/clusters.ts
```diff
@@ -447,6 +447,20 @@ export async function deleteCluster(
     }
 }
 
+export async function abortLastOperationInCluster(
+    target: AksClusterTreeItem,
+    clusterName: string
+): Promise<Errorable<string>> {
+    try {
+        const containerClient = getContainerClient(target);
+        await containerClient.managedClusters.beginAbortLatestOperationAndWait(target.resourceGroupName, clusterName)
```
peterbom, line 456: What happens here if there's no operation in progress? Is it a no-op, or an error?

And is there a way to check if an operation is in progress? (Noting that attempting to check in advance might lead to race conditions.)

Tatsinnit: yeah this happens :)  basically fairly detail in the error. 

I tried this in order and vice-versa - but I am guessing just like stop/start if mis-used then it will cause issue in cluster :P  you know what I mean.

Screenshot 2023-10-06 at 10 46 33 AM

peterbom: I wonder if we can check the cluster provisioning state before asking the user "Do you want to abort last operation in cluster ${clusterName}?"...

If it's not in a suitable state, we could just pop up an info message saying that?

Tatsinnit: Nice idea lemme update that :) thank you.

peterbom (approved):

PR #254, opened 2023-10-06
Add state management to webviews

peterbom:
Courtesy of @hsubramanianaks: this avoids webviews losing their state when they are hidden.

See: https://code.visualstudio.com/api/extension-guides/webview#retaincontextwhenhidden

The next step will be to work on an alternative approach to address the documented concern that "this has high memory overhead and should only be used when other persistence techniques will not work".


hsubramanianaks (approved):
Thanks @peterbom for kind words. Without talking to you we wouldn't have found this state management issue. Looks good.

PR #257, opened 2023-10-09
fix build.

Tatsinnit:
This PR fixes the ripple effect of updating `"@azure/arm-containerservice": "^19.3.0",` from `17.*` which changed the way status is passed. hence there was a build failure on create cluster.

FYi and cc: @peterbom, thanks heaps.

https://learn.microsoft.com/en-us/javascript/api/%40azure/arm-containerservice/managedcluster?view=azure-node-latest

Screenshot 2023-10-09 at 10 06 33 PM


peterbom (approved):
Thanks for spotting this! LGTM

PR #258, opened 2023-10-10
Avoid lengthy kubectl output changing the layout of containing elements

peterbom:
Fixes a display issue where wide output of a `kubectl` command would increase the width of containing elements, which could push the 'Run' button out of the visible area of the screen.

This ensures that the `pre` element containing the output remains at the size dictated by its parent containers, and handles any scrolling itself.


Tatsinnit (approved):
🎉🎉

PR #279, opened 2023-10-16
Simplify state management for new webviews

peterbom:
There is currently more boilerplate code than necessary required for creating new webviews with state, and handling the state updates based on events within the webviews and messages from VS Code.

This attempts to simplify the code and tidy up the naming so that it's clearer which handlers are for events within the webview and which are for messages from VS Code.

**Example Changes**

```typescript
// Before
export function createState(): ASOState { ... }
export const vscodeMessageHandler: StateMessageHandler<ToWebViewMsgDef, ASOState> = { ... };
export const userMessageHandler: StateMessageHandler<UserMsgDef, ASOState> = { ... };
export const updateState = chainStateUpdaters(
    toStateUpdater(vscodeMessageHandler),
    toStateUpdater(userMessageHandler));

// After
export const stateUpdater: WebviewStateUpdater<"aso", EventDef, ASOState> = {
    createState: initialState => ({ ... }),
    vscodeMessageHandler: { ... },
    eventHandler: { ... }
};
```

```typescript
// Before
const [state, dispatch] = useReducer(updateState, createState());
const userMessageEventHandlers = getEventHandlers<UserMsgDef>(dispatch, userMessageHandler);
useEffect(() => {
    const msgHandler = getMessageHandler<ToWebViewMsgDef>(dispatch, vscodeMessageHandler);
    vscode.subscribeToMessages(msgHandler);
}, []);

// After
const {state, eventHandlers, vsCodeMessageHandlers} = getStateManagement(stateUpdater, initialState);
useEffect(() => {
    vscode.subscribeToMessages(vsCodeMessageHandlers);
}, []);
```

```typescript
// Before
webview.postMessage({
    command: "installCertManagerResponse",
    parameters: {
        succeeded,
        errorMessage,
        commandResults: [{ command, stdout, stderr }]
    }
});

// After
webview.postInstallCertManagerResponse({
    succeeded,
    errorMessage,
    commandResults: [{ command, stdout, stderr }]
});
```

The penalty for adding these `post` functions is the need to supply an object containing all the keys of the message definition when creating a message sink. The values can be anything, so it feels a bit arbitrary, but you get type checking and auto-complete so it's not too painful. Example:
```typescript
export const vscode = getWebviewMessageContext<"aso">({
    checkSPRequest: null,
    installCertManagerRequest: null,
    installOperatorRequest: null,
    installOperatorSettingsRequest: null,
    waitForCertManagerRequest: null,
    waitForControllerManagerRequest: null
});
```


src/panels/DetectorPanel.ts
```diff
@@ -6,7 +6,7 @@ const meta = require('../../package.json');
 
 export class DetectorPanel extends BasePanel<"detector"> {
     constructor(extensionUri: Uri) {
-        super(extensionUri, "detector");
+        super(extensionUri, "detector", {});
```
hsubramanianaks, line 9: this panel doesn't have webview post messages ?

peterbom: That's right - and if it did, the above would cause a TypeScript compilation error

src/panels/BasePanel.ts
```diff
@@ -86,36 +87,29 @@ export abstract class BasePanel<TContent extends ContentId> {
...
-                } else {
-                    window.showErrorMessage(`No handler found for command ${message.command}`);
-                }
-            },
-            undefined,
-            this._disposables
-        );
-    }
-}
+function getMessageContext<TContent extends ContentId>(webview: Webview, webviewCommandKeys: CommandKeys<ToWebviewMsgDef<TContent>>, disposables: Disposable[]): VsCodeMessageContext<TContent> {
```
peterbom, line 90: We are no longer defining a class or implementing an interface here because the `MessageContext` types have a type-dependent number of members (`postXxx(...)` for each command).

src/webview-contract/messaging.ts
```diff
@@ -6,12 +6,30 @@ export type MessageDefinition = {
...
+    [P in Command<TMsgDef>]: any
+};
+
 /**
  * A component that messages can be sent to. Both Webviews and the
  * VS Code extension can act as `MessageSink` instances.
  */
-export interface MessageSink<TPostMsgDef extends MessageDefinition> {
-    postMessage(message: Message<TPostMsgDef>): void
+export type MessageSink<TMsgDef extends MessageDefinition> = {
```
peterbom, line 20: This is the change that enables posting messages with simpler syntax:
- Before: `postMessage({ command: "customMessage", parameters: args })`
- After: `postCustomMessage(args)`

The new `asMessageSink` function converts implementations of the former to the latter.

webview-ui/src/AzureServiceOperator/AzureServiceOperator.tsx
```diff
@@ -1,50 +1,43 @@
...
 import { Inputs } from "./Inputs";
 import { getRequiredInputs } from "./helpers/inputs";
 
-export function AzureServiceOperator(props: InitialState) {
-    const vscode = getWebviewMessageContext<"aso">();
-
-    const [state, dispatch] = useReducer(updateState, createState());
-
-    const userMessageEventHandlers = getEventHandlers<UserMsgDef>(dispatch, userMessageHandler);
+export function AzureServiceOperator(initialState: InitialState) {
```
peterbom, line 11: Note: this tries to establish a convention of explicitly naming the props as `initialState` so that it's clear it should only be used to set up state management (`getStateManagement(stateUpdater, initialState);`).

It is not intended to be referred to later in the component (`state` should be used instead).

webview-ui/src/AzureServiceOperator/helpers/state.ts
```diff
@@ -41,8 +41,21 @@ function completedStep(succeeded: boolean, result: InstallStepResult | null): In
     return {status: succeeded ? InstallStepStatus.Succeeded : InstallStepStatus.Failed, result};
 }
 
-export function createState(): ASOState {
-    return {
+export type EventDef = {
```
peterbom, line 44: Note: This is removed from `userCommands.ts` and renamed from `UserMsgDef`. The original naming was intended to convey the fact that these events are typically user-initiated, in contrast to messages from VS Code. But really they are any webview events, and can be triggered programmatically, e.g. during initialization. So the name `EventDef` is more appropriately generic, and ties in more obviously with `eventHandler` on the `WebviewStateUpdater` type.

webview-ui/src/Detector/state.ts
```diff
@@ -0,0 +1,20 @@
+import { InitialState } from "../../../src/webview-contract/webviewDefinitions/detector";
```
peterbom, line 1: Added for consistency with other webviews, even though it adds some extra code.

Tatsinnit (approved):
🎉 since it is progressive PR for over all state change as long as we have this tested and all seems in order we house go ahead and might be incrementally refactor. Longer run minor documentation at the repo level will be super.

❤️🙏

src/panels/BasePanel.ts
Tatsinnit: Sounds good, can you resolve the discussions and we can merge this. Thank you, WDYT?

PR #280, opened 2023-10-18
Feature/reconcile cluster using update

Tatsinnit:
This PR is draft and it enables - Reconcile cluster operation which is an empty update on the cluster.

### DIFF
--- a/package.json
+++ b/package.json
@@ -233,6 +233,10 @@
                 "title": "Abort Last Operation"
             },
             {
+                "command": "aks.aksReconcileCluster",
+                "title": "Reconcile Operation"
+            },
+            {
                 "command": "aks.aksInspektorGadgetShow",
                 "title": "Show Inspektor Gadget"
             },
@@ -368,6 +372,10 @@
                 }, 
                 {
                     "command": "aks.aksAbortLastOperationInCluster",
+                    "group": "navigation"
+                }, 
+                {
+                    "command": "aks.aksReconcileCluster",
                     "group": "navigation"
                 }
             ]
--- a/package.json
+++ b/package.json
@@ -234,7 +234,7 @@
             },
             {
                 "command": "aks.aksReconcileCluster",
-                "title": "Reconcile Operation"
+                "title": "Reconcile Cluster"
             },
             {
                 "command": "aks.aksInspektorGadgetShow",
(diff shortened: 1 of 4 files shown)


src/commands/aksReconcileCluster/aksReconcileCluster.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const cluster = getAksClusterTreeItem(target, cloudExplorer);
+  if (failed(cluster)) {
+    vscode.window.showErrorMessage(cluster.error);
+    return;
+  }
+
+  const clusterName = cluster.result.name;
+
+  const clusterProvisioingState = await determineProvisioningState(cluster.result, clusterName);
```
peterbom, line 22: Looks like this is unused and could be removed...unless we can figure out the circumstances under which reconciliation is appropriate, and this is relevant?

src/commands/aksReconcileCluster/aksReconcileCluster.ts
```diff
@@ -0,0 +1,39 @@
+import * as vscode from 'vscode';
+import * as k8s from 'vscode-kubernetes-tools-api';
+import { IActionContext } from "@microsoft/vscode-azext-utils";
+import { reconcileUsingUpdateInCluster, determineProvisioningState, getAksClusterTreeItem } from '../utils/clusters';
```
hsubramanianaks, line 4: I think determineProvisioningState needs to be removed from the imports as well.

sprab:
Tested this with multiple scenarios and works as expected.

sprab (approved):
Tested with multiple scenarios and works as expected.

sprab (commented):
Tested this with various scenarios that includes changes made to MC resources outside of AKS RP and the reconcile works as expected.

package.json
```diff
@@ -232,6 +232,10 @@
                 "command": "aks.aksAbortLastOperationInCluster",
                 "title": "Abort Last Operation"
             },
+            {
+                "command": "aks.aksReconcileCluster",
+                "title": "Reconcile Operation"
```
peterbom, line 237: Could we call this just 'Reconcile', or maybe 'Reconcile Cluster'?

As it is, it sounds a bit like 'reconcile the operation', which doesn't make sense.

hsubramanianaks (approved):
@Tatsinnit  looks good, tested with my own clusters and it works fine. Thanks for this feature, it will solve multiple user AKS cluster problems.

peterbom (approved):
Looks good - thanks @Tatsinnit! I'd like to do some more work in future around when it's appropriate to perform this action, and guide users accordingly, but that'll require a lot more analysis. For now, let's make this functionality available.

PR #283, opened 2023-11-01
Clean up unused or unnecessary dependencies

peterbom:
The first part of #286.

Removes most of the dependencies identified by [depcheck](https://www.npmjs.com/package/depcheck) as being unused.

The `depcheck` tool also identified some _missing_ dependencies, including `tslint`. Tslint is now [deprecated](https://github.com/palantir/tslint/issues/4534) in favour of `eslint`, so this also converts the project to `eslint`. No linting is done as part of any build scripts or pipelines, but there is now an npm script to run it manually. I'll create a separate issue to enable automated linting.

The `utils` dependency was there to support the 'promisification' of node's `fs` library. This is now unnecessary thanks to the [inbuilt `fs/promises`](https://nodejs.org/api/fs.html#file-system) library.

This also removes some redundant processing from the webpack configuration.


src/commands/utils/download/download.ts
```diff
@@ -1,16 +1,8 @@
 import * as path from 'path';
-import * as stream from 'stream';
 import { succeeded, Errorable, getErrorMessage} from '../errorable';
 import { Dictionary } from '../dictionary';
 import { sleep } from '../sleep';
-
```
peterbom, line 6: Note: this hack was originally added to avoid the `download` import overwriting the user's `HOME` environment variable. This happened because of a [transitive](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/302#issuecomment-404678781) dependency on `npm-conf` via the [`caw`](https://www.npmjs.com/package/caw) package, which has since [been removed](https://github.com/kevva/download/commit/a16ba04b30dafbe7d9246db93f1534320d8e0dd3) as a dependency of `download`. I've verified that the `HOME` value is not being overwritten, even without this workaround.

src/tests/suite/index.ts
```diff
@@ -1,41 +1,31 @@
 import * as path from 'path';
 import * as Mocha from 'mocha';
-import * as glob from 'glob';
+import { glob } from 'glob';
```
peterbom, line 3: `glob` was previously an indirect dependency, even though we're using it explicitly.

These changes result from taking a dependency on the latest version, which provides a Promise-based async API.

Tatsinnit:
💡 This looks great to me, I can surely take a look bit more, but just for idea sake, shall we merge this after tcpDump work - and cut a separate release branch and then merge this.

src/tests/runTests.ts
```diff
@@ -1,22 +1,26 @@
...
- 		await runTests({ extensionDevelopmentPath, extensionTestsPath });
- 	} catch (err) {
- 		console.error('Failed to run tests');
- 		process.exit(1);
- 	}
- }
+        // Download VS Code, unzip it and run the integration test
+        await runTests({ extensionDevelopmentPath, extensionTestsPath });
+    } catch (err) {
+        console.error(`Failed to run tests:\n${err}`);
```
peterbom, line 17: Added this extra logging because swallowing the error here caused hours of head-scratching today.

webpack.config.js
```diff
@@ -29,26 +28,12 @@ const config = {
...
   },
   plugins: [
-      new webpack.IgnorePlugin({ resourceRegExp: /^\.\/locale$/, contextRegExp: /^electron$/}),
-      new webpack.IgnorePlugin({ resourceRegExp: /^\.\/locale$/, contextRegExp: /handlebars$/ }),
-      new fileManagerPlugin({
-        events: {
-          onEnd: {
-              copy: [
-                  {
-                      source: path.join(__dirname, 'node_modules', '@microsoft', 'vscode-azext-azureutils', 'resources', '**'),
```
hsubramanianaks, line 39: this is removed due to webpack update ? Just trying to understand.

peterbom: No, I think it's actually due to [an update](https://github.com/Azure/vscode-aks-tools/pull/150/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) a while back where we upgraded some vscode libraries that previously needed this copying, but now apparently don't. At least I've built and run it, and everything seems to be working.

This is worth double-checking though, because I'm not totally sure why this was needed in the first place. @Tatsinnit - do you remember?

Tatsinnit: ❤️❤️🙏 @peterbom, thank you for the kind ping for this, I think we have the changes coming into this PR soon, rest looks sweet, if you share VSIX we all can test

peterbom: Actually I tested this and it is needed (for treeview icons to appear), so I've added it back.

Tatsinnit (approved):
Looks good, thank you!  worth test which we all can chip in, fyi @hsubramanianaks for viz

webpack.config.js
peterbom: Icons are back:
image

PR #289, opened 2023-11-14
Feature - agent pool abort last operation

Tatsinnit:
This PR enables the `Agent Pool Last Operation` via show properties page.

The PR has rest of the implementation details and cc: @sprab + @peterbom + @hsubramanianaks  🎊☕️ for a Quick Look.

image

Screenshot 2023-11-14 at 2 45 47 PM


webview-ui/src/ClusterProperties/AgentPoolDisplay.tsx
```diff
@@ -1,22 +1,54 @@
...
-    profileInfo: AgentPoolProfileInfo
+    clusterInfo: ClusterInfo;
+    profileInfo: AgentPoolProfileInfo;
+    eventHandlers: EventHandlers<EventDef>;
+    clusterOperationRequested: boolean;
 }
 
+// Abort on anything except: Canceled, Failed, Succeeded
+// Check about deleting
+type ProvisionState = "CapturingSecurityVHDSnapshot" | "Creating" | "Deleting" | "RefreshingServicePrincipalProfile"
```
peterbom, line 16: This list is probably not exhaustive, and I'm afraid that readers may think it's supposed to be. We're only using a subset of these, which we _do_ hope to be exhaustive, i.e. the ones that indicate an operation is not in progress (I'm wanting to call these "terminal" states, although I'm not confident that's correct terminology).

Instead, we could declare a `const` collection of these terminal states, and use those instead, like:
```typescript
const terminalProvisioningStates = ["Canceled", "Failed", "Succeeded"] as const;
```

If we wanted to create a discriminated union type out of these (which isn't needed yet) we could use:
```typescript
type TerminalProvisioningState = typeof terminalProvisioningStates[number];
```

webview-ui/src/ClusterProperties/AgentPoolDisplay.tsx
```diff
@@ -1,22 +1,54 @@
...
+                    | "RotatingClusterCertificates" | "Scaling" | "Starting" | "Stopping" | "Succeeded" | "Updating" 
+                    | "Upgrading" | "UpgradingNodeImageVersion";
+
 export function AgentPoolDisplay(props: AgentPoolDisplayProps) {
+    const isProvisioningStateFromCluster = props.clusterInfo.provisioningState === props.profileInfo.provisioningState;
+    const isOperationInProgress = !["Canceled", "Failed", "Succeeded"].includes(props.profileInfo.provisioningState);
```
peterbom, line 22: If we replace the `ProvisionState` type with `terminalProvisioningStates`, we can change this to:
```typescript
const isOperationInProgress = !terminalProvisioningStates.includes(props.profileInfo.provisioningState);
```

webview-ui/src/manualTest/clusterPropertiesTests.tsx
```diff
@@ -115,6 +116,22 @@ export function getClusterPropertiesScenarios() {
...
+
+        async function handleAbortAgentPoolOperation(agentPoolName: string, hasError: boolean) {
+            await new Promise(resolve => setTimeout(resolve, 1000));
+            if (hasError) {
+                webview.postErrorNotification("Sorry, I wasn't quite able to abort the operation.");
+                return;
+            }
+
+            const agentPoolProfile = testClusterInfo.agentPoolProfiles.find(p => p.name === agentPoolName)!;
+            agentPoolProfile.provisioningState = "Canceling";
```
peterbom, line 128: In reality, it looks as if provisioning state doesn't change after invoking an 'abort' request, so we can remove this line that changes the state to "Canceling".

webview-ui/src/manualTest/clusterPropertiesTests.tsx
```diff
@@ -115,6 +116,21 @@ export function getClusterPropertiesScenarios() {
             });
             webview.postGetPropertiesResponse(testClusterInfo);
         }
+
+        async function handleAbortAgentPoolOperation(agentPoolName: string, hasError: boolean) {
+            await new Promise(resolve => setTimeout(resolve, 1000));
+            if (hasError) {
+                webview.postErrorNotification("Sorry, I wasn't quite able to abort the operation.");
```
hsubramanianaks, line 123: Can we finetune this error msg ? ex: ```Abort agent pool operation failed due to error``` ?

Tatsinnit: Sure, whatever here is fine **because** this is **test** error so any error here is just for testing sake no end to end functionality is in direct impact. Do we need consensus - if no replies I could just change it to whatever recommended. thanks.

hsubramanianaks: Sorry didn't realize this is test file. I will resolve this comment.

webview-ui/src/ClusterProperties/AgentPoolDisplay.tsx
```diff
@@ -1,22 +1,51 @@
...
+
     return (
         <dl className={styles.propertyList}>
+            <dt>Provisioning State</dt>
+            <dd>
+                {props.profileInfo.provisioningState}
+                {showAbortButton &&
+                <>
+                    &nbsp;
+                    <VSCodeButton disabled={props.clusterOperationRequested} onClick={() => handleAbortClick(props.profileInfo.name)}>Abort</VSCodeButton>
```
peterbom, line 35: Sorry, I completely forgot about this, but I think this button should have the attribute `appearance="secondary"`. That'll change the colour from that really bright blue to something a bit more subtle.

While we're at it, I think the same attribute should be applied to the Start and Stop buttons (in `ClusterDisplay.tsx`).

Tatsinnit: The `Secondary` gives this and primary gives this:

I almost feel `Primary` is better :) see how abort looks almost camouflaged, but I am not good in colour choice, let me know and I can make primary or secondary.

image

Tatsinnit: sounds good, lets go with `primary` made the change and pushed.

peterbom (approved):

Issue #291, opened 2023-11-15
Add refresh ability to cluster properties page


peterbom:
This could be a refresh button, or periodic polling, or a combination of both.

PR #292, opened 2023-11-16
Feature: abort/reconcile cluster operation in show properties page.

Tatsinnit:
This PR adds / move the `Abort Last Operation` and `Reconcile` at cluster level at the `show Properties` page level and remove the command run from the package.json.

Idea/Thought process is to accumulate Clos knot feature of `start. Stop, Abort last node and Abost Last Cluster Operation` in one location which is show properties page.

Thanks and cc: @peterbom please.

Screenshot 2023-11-16 at 2 33 35 PM

Screenshot 2023-11-16 at 5 52 35 PM


src/panels/ClusterPropertiesPanel.ts
```diff
@@ -64,6 +65,29 @@ export class ClusterPropertiesDataProvider implements PanelDataProvider<"cluster
         }
     }
 
+    private async _handleAbortClusterOperation(webview: MessageSink<ToWebViewMsgDef>) {
+        try {
+            await this._readAndPostClusterProperties(webview);
```
peterbom, line 70: I see this line is in the agent pool abort method, too, but I can't remember why we added it, and can't see a reason for having it. Do you remember?

Tatsinnit: Yeah, I think it was their for status update, but let me see removing will fix the enable/disable button thing. 🤷🏽‍♂️☕️☕️

webview-ui/src/ClusterProperties/ClusterDisplay.tsx
```diff
@@ -14,6 +14,8 @@ export interface ClusterDisplayProps {
 
 type StartStopState = "Started" | "Starting" | "Stopped" | "Stopping";
 
+const terminalProvisioningStates = ["Canceled", "Failed", "Succeeded", "Starting", "Stopping"];
```
peterbom, line 17: I think this means we will never be able to abort start/stop operations - is that right? Are we sure about that restriction?

If that's what we want, I'd prefer to make it explicit in the code and keep `terminalProvisioningStates` consistent with the agent pool level ones (i.e. just `["Canceled", "Failed", "Succeeded"]`).

If we want to prevent users aborting start/stop operations, could we have a separate variable for this, something like:
```typescript
const unabortableProvisioningStates = [...terminalProvisioningStates, "Starting", "Stopping"];
```

And then later:
```typescript
const showAbortButton = !unabortableProvisioningStates.includes(props.clusterInfo.provisioningState);
```

webview-ui/src/manualTest/clusterPropertiesTests.tsx
```diff
@@ -131,6 +133,31 @@ export function getClusterPropertiesScenarios() {
...
         }
+
+        async function handleAbortClusterOperation(hasError: boolean) {
+            await new Promise(resolve => setTimeout(resolve, 1000));
+            if (hasError) {
+                webview.postErrorNotification("Sorry, I wasn't quite able to abort the operation.");
+                return;
+            }
+
+            testClusterInfo.provisioningState = "Succeeded";
```
peterbom, line 144: Can we check what actually happens to the provisioning and power states when we abort a cluster operation, and make this consistent?

I _think_ that nothing changes immediately after sending the abort request, and then a little while later the cluster-level provisioning state changes to "Canceled". I'm not sure what happens at the agent pool level, but hopefully something similar?

It might make sense to avoid changing the power state at all here?

src/panels/ClusterPropertiesPanel.ts
```diff
@@ -64,6 +65,53 @@ export class ClusterPropertiesDataProvider implements PanelDataProvider<"cluster
...
+            await poller.pollUntilDone();
+        } catch (ex) {
+            const errorMessage = getErrorMessage(ex);
+            webview.postErrorNotification(errorMessage);
+        }
+    }
+
+    private async _handleReconcileClusterOperation(webview: MessageSink<ToWebViewMsgDef>) {
+        try {
+            const getClusterInfo = await this.client.managedClusters.get(this.resourceGroup, this.clusterName);
```
peterbom, line 92: Just a note for future reference because I don't think it's worth changing this right now, but ideally we'd pass `location` from the webview rather than performing an extra `get` here.

Maybe we can look at doing that as part of #291.

src/panels/ClusterPropertiesPanel.ts
```diff
@@ -64,6 +65,53 @@ export class ClusterPropertiesDataProvider implements PanelDataProvider<"cluster
...
+            const poller = await this.client.managedClusters.beginCreateOrUpdate(this.resourceGroup, this.clusterName, {
+                location: getClusterInfo.location
+            });
+
+            poller.onProgress((state) => {
```
peterbom, line 97: I think we can have an extra progress check for "cancelled" here (as we do for start/stop).

The reason we skipped that for Abort was because Abort effectively _is_ a cancel operation, so I don't think it makes sense to cancel it. But Reconcile is an update, which I think _could_ be cancelled?

I guess we could test this by kicking off a long-running reconcile and trying to abort it from the CLI.

webview-ui/src/manualTest/clusterPropertiesTests.tsx
```diff
@@ -59,7 +59,9 @@ export function getClusterPropertiesScenarios() {
             getPropertiesRequest: () => handleGetPropertiesRequest(),
             stopClusterRequest: () => handleStopClusterRequest(withErrors && sometimes()),
             startClusterRequest: () => handleStartClusterRequest(withErrors && sometimes()),
-            abortAgentPoolOperation: (agentPoolName: string) => handleAbortAgentPoolOperation(agentPoolName, withErrors && sometimes())
+            abortAgentPoolOperation: (agentPoolName: string) => handleAbortAgentPoolOperation(agentPoolName, withErrors && sometimes()),
+            abortClusterOperation: () => handleAbortClusterOperation(withErrors && sometimes()),
+            reconcileClusterRequest: () => { }
```
peterbom, line 64: I missed this earlier - we should add a handler here for Reconcile

peterbom (approved):

PR #333, opened 2023-11-22
Eslint and Prettier integration

### LINKED ISSUE #284: Better linting configuration and build integration
#283 updates the project to use `eslint` instead of `tslint`. There are a [few more steps](https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration) required for effective linting:
- [Extend the configuration](https://github.com/typescript-eslint/tslint-to-eslint-config/blob/main/docs/FAQs.md#should-i-use-tslint-to-eslint-config) from `typescript-eslint`'s recommended rulesets.
  ```json
    "extends": [
        "eslint:recommended",
        "plugin:@typescript-eslint/recommended"
    ],
  ```
  - This will involve cleaning up the `.eslintrc.js` config file to avoid re-specifying rules.
- [Use Prettier](https://github.com/typescript-eslint/tslint-to-eslint-config/blob/main/docs/FAQs.md#should-i-use-prettier) rather than the linter to handle formatting concerns.
  - We might use [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) as part of this.
- Add linting to the `webview-ui` project. The lint config currently points to `tsconfig.json` as the project root, and `webview-ui` has its own TS configuration.
- Consider [typed linting](https://typescript-eslint.io/linting/typed-linting/).
- Fix up all linting warnings/errors.
- Convert warnings to errors and run linting as part of build scripts/pipelines for CI.

peterbom:
Resolves #284 
- Adds eslint to `webview-ui` project
- Makes rules consistent between projects
- Ensures linting of both projects is done as part of build
- Delegates formatting responsibilities to Prettier
- Runs Prettier on save in VS Code


src/commands/aksInspektorGadget/traceItems.ts
```diff
@@ -34,18 +37,10 @@ export function parseOutputLine(line: string): Errorable<object[]> {
  */
 export function asFlatItems(item: object): TraceOutputItem[] {
     const allObjectEntries = Object.entries(item).reduce(addFlatEntries, [[]]);
-    return allObjectEntries.map(objectEntries => Object.fromEntries(objectEntries));
+    return allObjectEntries.map((objectEntries) => Object.fromEntries(objectEntries) as TraceOutputItem);
 }
 
-export function isObject(value: any) {
```
peterbom, line 40: These are moved to `utils/runtimeTypes.ts`

src/commands/periscope/helpers/periscopehelper.ts
```diff
@@ -65,36 +70,51 @@ export async function chooseStorageAccount(diagnosticSettings: DiagnosticSetting
...
 
         // Get keys from storage client.
-        const storageClient = new StorageManagementClient(cluster.subscription.credentials, cluster.subscription.subscriptionId);
+        const storageClient = new StorageManagementClient(
+            cluster.subscription.credentials,
+            cluster.subscription.subscriptionId,
+        );
         const storageAccKeyList = await storageClient.storageAccounts.listKeys(resourceGroupName, accountName);
-        const storageKey = storageAccKeyList.keys?.find((it) => it.keyName === "key1")?.value!;
+        if (storageAccKeyList.keys === undefined) {
```
peterbom, line 105: Apparently the previous code was "unsafe and wrong", which made me feel very ashamed of myself, so it's replaced by a bunch of very explicit checks.

src/commands/utils/azureAccount.ts
```diff
@@ -144,39 +184,43 @@ function getMicrosoftGraphClientBaseUrl(environment: Environment): string {
...
     // The service principal needs *some* permissions in the subscription, but Contributor is not
     // necessarily required. See: https://azure.github.io/azure-service-operator/#installation
-    return { succeeded: true, result: {subscription, hasRoleAssignment: roleAssignments.length > 0}};
+    return { succeeded: true, result: { subscription, hasRoleAssignment: roleAssignments.length > 0 } };
 }
 
-function isUnauthorizedError(e: any): boolean {
-    return e.code === 'AuthorizationFailed' && e.statusCode === 403;
+function isUnauthorizedError(e: unknown): boolean {
+    return isObject(e) && "code" in e && "statusCode" in e && e.code === "AuthorizationFailed" && e.statusCode === 403;
```
peterbom, line 225: I don't think this will behave differently in practice, because errors are _usually_ objects (especially from trusted libraries), but this adds and extra check just to make sure.

src/commands/utils/config.ts
```diff
@@ -38,93 +39,99 @@ export function getKustomizeConfig(): Errorable<KustomizeConfig> {
...
         return [];
     }
 
-    return value.filter(isCommand).map(item => ({...item, category: CommandCategory.Custom}));
+    return value.filter(isCommand).map((item) => ({ ...item, category: CommandCategory.Custom }));
 
-    function isCommand(value: any): value is PresetCommand {
-        return (value.constructor.name === 'Object') && (value as PresetCommand).command && (value as PresetCommand).name ? true : false;
+    function isCommand(value: unknown): value is PresetCommand {
+        return isObject(value) && "command" in value && "name" in value;
```
peterbom, line 117: This should be equivalent to the previous code, except the old code wouldn't handle nulls, whereas `isObject` does.

src/commands/utils/detectors.ts
```diff
@@ -1,60 +1,66 @@
...
+import {
+    ARMResponse,
+    CategoryDetectorARMResponse,
+    SingleDetectorARMResponse,
+    isCategoryDataset,
+} from "../../webview-contract/webviewDefinitions/detector";
 import { getResourceManagementClient } from "./clusters";
-const tmp = require('tmp');
-const meta = require('../../../package.json');
+import { dirSync } from "tmp";
```
peterbom, line 12: Not sure why we were using `require()` for importing `tmp` functions, so I thought I'd better call this out in case there's a reason I don't understand. It seems to build and run fine, so I _think_ this change is fine.

src/commands/utils/detectors.ts
```diff
@@ -1,60 +1,66 @@
...
-) {
-    const outputDirObj = tmp.dirSync();
-
-    function saveDetector(detector: ARMResponse<any>) {
-        const detectorFilePath = path.join(outputDirObj.name, `${detector.name}.json`);
-        // Anonymize the data.
-        detector.id = "/subscriptions/12345678-1234-1234-1234-1234567890ab/resourcegroups/test-rg/providers/Microsoft.ContainerService/managedClusters/test-cluster/detectors/" + detector.name;
-        fs.writeFileSync(detectorFilePath, JSON.stringify(detector, null, 2));
-    }
+export async function saveAllDetectorResponses(target: AksClusterTreeItem, categoryDetectorIds: string[]) {
```
peterbom, line 18: Just some minor refactoring to move the nested function out.

Note, this code is not run as part of any commands, and was put in for generating realistic test data from ARM calls.

src/commands/utils/download/download.ts
```diff
@@ -1,56 +1,50 @@
-import * as path from 'path';
-import { succeeded, Errorable, getErrorMessage} from '../errorable';
-import { Dictionary } from '../dictionary';
-import { sleep } from '../sleep';
-import * as download from 'download';
+import * as path from "path";
+import { succeeded, Errorable, getErrorMessage } from "../errorable";
+import { Dictionary } from "../dictionary";
+import { sleep } from "../sleep";
+import download from "download";
```
peterbom, line 5: I actually think this shouldn't be equivalent, but `eslint` complained that `download` was the wrong type in the previous code. I tested downloading files after the update and it still works.

src/commands/utils/kubectl.ts
```diff
@@ -107,53 +134,68 @@ export async function getResources<T>(
...
 }
 
-export async function streamKubectlOutput(kubectl: APIAvailable<KubectlV1>, kubeConfigFile: string, kubectlArgs: string[]): Promise<Errorable<OutputStream>> {
-    const kubectlInternal = <KubectlInternal>(<any>kubectl.api).kubectl;
+export async function streamKubectlOutput(
+    kubectl: APIAvailable<KubectlV1>,
+    kubeConfigFile: string,
+    kubectlArgs: string[],
+): Promise<Errorable<OutputStream>> {
+    const kubectlInternal = asInternal(kubectl.api);
```
peterbom, line 158: Replaced `any` cast with `asInternal` function that returns an `errorable` if the underlying implementation changes.

src/commands/utils/reporter.ts
```diff
@@ -19,10 +20,10 @@ interface IPackageInfo {
     aiKey: string;
 }
 
-function getPackageInfo(context: vscode.ExtensionContext): IPackageInfo | undefined {
-    const extensionPackage = require('../../../package.json');
+function getPackageInfo(): IPackageInfo | undefined {
+    const extensionPackage = meta;
```
peterbom, line 24: moved this to a top-level import - not sure if there was a reason it was previously loaded dynamically?

src/extension.ts
```diff
@@ -1,93 +1,105 @@
...
+import aksRotateClusterCert from "./commands/aksRotateClusterCert/aksRotateClusterCert";
+import { aksInspektorGadgetShow } from "./commands/aksInspektorGadget/aksInspektorGadget";
+import aksCreateCluster from "./commands/aksCreateCluster/aksCreateCluster";
+import aksReconcileCluster from "./commands/aksReconcileCluster/aksReconcileCluster";
+import { aksTCPDump } from "./commands/aksTCPCollection/tcpDumpCollection";
 
 export async function activate(context: vscode.ExtensionContext) {
     const cloudExplorer = await k8s.extension.cloudExplorer.v1;
-    context.subscriptions.push(new Reporter(context));
+    context.subscriptions.push(new Reporter());
```
peterbom, line 48: `context` was never used here.

src/panels/TcpDumpPanel.ts
```diff
@@ -1,21 +1,26 @@
...
+} from "../webview-contract/webviewDefinitions/tcpDump";
 import { withOptionalTempFile } from "../commands/utils/tempfile";
 
 const debugPodNamespace = "default";
 const tcpDumpCommandBase = "tcpdump --snapshot-length=0 -vvv";
 const captureDir = "/tmp";
 const captureFilePrefix = "vscodenodecap_";
 const captureFileBasePath = `${captureDir}/${captureFilePrefix}`;
-const captureFilePathRegex = `${captureFileBasePath.replace(/\//g, '\\$&')}(.*)\.cap`; // Matches the part of the filename after the prefix
+const captureFilePathRegex = `${captureFileBasePath.replace(/\//g, "\\$&")}(.*)\\.cap`; // Matches the part of the filename after the prefix
```
peterbom, line 23: Added a second `\` character to escape the other one, which would previously have been ignored, meaning the regex would match _any_ character before the extension, not just a literal `.`. So I think the linter caught a genuine bug here.

src/panels/TcpDumpPanel.ts
```diff
@@ -28,7 +33,7 @@ function getTcpDumpCommand(capture: string): string {
 function getCaptureFromCommand(command: string, commandWithArgs: string): string | null {
     if (command !== "tcpdump") return null;
     if (!commandWithArgs.startsWith(tcpDumpCommandBase)) return null;
-    const fileMatch = commandWithArgs.match(new RegExp(`\-w ${captureFilePathRegex}`));
+    const fileMatch = commandWithArgs.match(new RegExp(`\\-w ${captureFilePathRegex}`));
```
peterbom, line 36: Escaping the `\` so that a `\` character appears in the regex string to capture a literal `-` in the command. I haven't quite thought through how this could have been correct before, so it's worth doing a bit more testing (although it does seem to work now).

src/tests/runTests.ts
```diff
@@ -1,26 +1,25 @@
...
-import { runTests } from '@vscode/test-electron';
+import * as path from "path";
+import { runTests } from "@vscode/test-electron";
 
 async function main() {
     try {
         // The folder containing the Extension Manifest package.json
         // Passed to `--extensionDevelopmentPath`
-        const extensionDevelopmentPath = path.resolve(__dirname, '../../');
+        const extensionDevelopmentPath = path.resolve(__dirname, "../../..");
```
peterbom, line 8: Had to change the typescript root folder from `src` to the project root, because some `.ts` files import `package.json` from the root, and so we want the compiler to be aware of that file. That means that the `out` folder now contains `src` and everything is nested one level deeper, hence the change in the relative path here.

Note, the `out` folder is only used for running tests, not building the vsix, and tests are now running successfully.

webview-ui/src/AzureServiceOperator/AzureServiceOperator.tsx
```diff
@@ -4,67 +4,76 @@ import { InitialState } from "../../../src/webview-contract/webviewDefinitions/a
 import { useEffect } from "react";
 import { InstallStepStatus, stateUpdater, vscode } from "./helpers/state";
 import { Progress, StepWithDescription } from "./Progress";
-import { getStateManagement } from "../utilities/state";
+import { useStateManagement } from "../utilities/state";
```
peterbom, line 7: This is changed to use a [custom hook](https://react.dev/learn/reusing-logic-with-custom-hooks). Some of the logic (like `useState`) that was in this function was not valid for an ordinary library function, but if you name the function `useXxxx` it's interpreted as a custom hook and that stuff is allowed. It also allows `useEffect`, meaning we could move the following three lines of repeated code into the function, further cleaning up the React components.

webview-ui/src/InspektorGadget/GadgetSelector.tsx
```diff
@@ -4,11 +4,11 @@ import { GadgetCategory } from "./helpers/gadgets/types";
...
 
 export interface GadgetSelectorProps {
-    category: GadgetCategory
-    id: string
-    className: React.HTMLAttributes<any>['className']
-    required?: boolean
-    onResourceChanged: (resource: string | null) => void
+    category: GadgetCategory;
+    id: string;
+    className?: string;
```
peterbom, line 9: The old type `React.HTMLAttributes<any>['className']` evaluates to `string | undefined`, so the simpler type is equivalent. (I guess the complex type name could be used by some tooling for intellisense but I don't think it provides any value for us.)

webview-ui/src/InspektorGadget/helpers/gadgets.ts
```diff
@@ -100,18 +122,26 @@ function negateSortFunction(fn: SortFunction): SortFunction {
...
-    let key = specifier.property.key;
-    let descending = specifier.direction === SortDirection.Descending;
+function asSortFunction(specifier: SortSpecifier<string>): SortFunction {
+    const key = specifier.property.key;
+    const descending = specifier.direction === SortDirection.Descending;
     const fn: SortFunction = (a, b) => {
-        if (a[key] > b[key]) {
+        const aValue = a[key];
+        const bValue = b[key];
+        if (aValue === null && bValue === null) {
```
peterbom, line 131: I'm not sure whether values can ever be `null`. The previous code didn't handle `null`s, so added these extra checks to be safe.

webview-ui/src/Periscope/Periscope.tsx
```diff
@@ -4,39 +4,44 @@ import { InitialState } from "../../../src/webview-contract/webviewDefinitions/p
...
 
 export function Periscope(initialState: InitialState) {
-    const {state, eventHandlers, vsCodeMessageHandlers} = getStateManagement(stateUpdater, initialState);
+    const { state, eventHandlers } = useStateManagement(stateUpdater, initialState, vscode);
 
     useEffect(() => {
-        vscode.subscribeToMessages(vsCodeMessageHandlers);
         sendUploadStatusRequest();
-    }, []); // Empty list of dependencies to run only once: https://react.dev/reference/react/useEffect#useeffect
+    });
```
peterbom, line 15: This removal of the empty list of dependencies (done because there were hidden dependencies) will result in more frequent upload status requests. It's worth testing this and seeing if it's too chatty.

webview-ui/src/utilities/runtimeTypes.ts
```diff
@@ -0,0 +1,7 @@
+export function isObject(value: unknown): value is object {
```
peterbom, line 1: Note - this file is duplicated between the webview and extension projects. In the future it might be good to create a library of files that can be shared between the two (those that have no dependency on vscode or react). A discussion for later...

webview-ui/src/utilities/state.ts
```diff
@@ -55,51 +76,68 @@ type StateUpdater<TState> = React.Reducer<TState, StateMessage>;
...
  * `useReducer` function.
  */
-function toStateUpdater<TMsgDef extends MessageDefinition, TState>(handler: StateMessageHandler<TMsgDef, TState>): StateUpdater<TState> {
+function toStateUpdater<TMsgDef extends MessageDefinition, TState>(
+    handler: StateMessageHandler<TMsgDef, TState>,
+): StateUpdater<TState> {
     return (state, msg) => {
         if (msg.command in handler) {
-            return handler[msg.command](state, msg.args);
+            return handler[msg.command](state, msg.args as TMsgDef[string]); // TODO: Check this cast
```
peterbom, line 84: This is just a type annotation and won't affect the compiled JavaScript, so there's no urgent need to resolve the TODO, but I want to keep it there as a reminder that there's something not quite right with the types here.

src/commands/utils/reporter.ts
peterbom: Used by telemetry, verified that reporting still works correctly after the change.

src/extension.ts
peterbom: Used by telemetry, verified that reporting still works correctly after the change.

src/panels/TcpDumpPanel.ts
peterbom: Turns out that the `-` character [doesn't always need escaping](https://stackoverflow.com/a/5484178). In this case it wasn't being escaped (because the escape character `\` itself needed escaping to appear in the regex string), but it still worked because this was one of the cases where it didn't need escaping (it was correctly interpreted as a literal).

This change escapes it, which is also valid.

src/tests/runTests.ts
peterbom: Tats asked the question of whether the `tsconfig` change would increase the vsix side (because it's looking at more than just the `src` folder). It doesn't increase the size because webpack only pulls in what is needed starting from `/src/extension.ts`.

However, for performance reasons it's probably best to ensure that we're not looking at unnecessary folders for TypeScript compilation, linting, and prettifying. I'll add another update for this.

peterbom: Updated. `tsconfig.json` now uses `includes` (specifying `src`) instead of `excludes`. This still avoids the error when files are pulled in from outside the `src` folder.

webview-ui/src/Periscope/Periscope.tsx
peterbom: Manually tested, and not noticed any different behaviour.

.vscode/launch.json
```diff
@@ -27,10 +27,10 @@
             "runtimeExecutable": "${execPath}",
             "args": [
                 "--extensionDevelopmentPath=${workspaceFolder}",
-                "--extensionTestsPath=${workspaceFolder}/out/tests/suite/index"
+                "--extensionTestsPath=${workspaceFolder}/out/src/tests/suite/index"
             ],
             "outFiles": [
-                "${workspaceFolder}/out/tests/**/*.js"
+                "${workspaceFolder}/out/src/tests/**/*.js"
```
Tatsinnit, line 33: 💡Please Note: I think we chatted about this, for wider readers - this change don't really make any functional problem with the code.

src/commands/aksTCPCollection/tcpDumpCollection.ts
```diff
@@ -1,14 +1,14 @@
...
-import { getVersion, invokeKubectlCommand } from '../utils/kubectl';
+import { getKubernetesClusterInfo } from "../utils/clusters";
+import { getExtension } from "../utils/host";
+import { Errorable, failed, map as errmap } from "../utils/errorable";
+import * as tmpfile from "../utils/tempfile";
+import { TcpDumpDataProvider, TcpDumpPanel } from "../../panels/TcpDumpPanel";
+import { getVersion, invokeKubectlCommand } from "../utils/kubectl";
 
-export async function aksTCPDump(_context: IActionContext, target: any) {
+export async function aksTCPDump(_context: IActionContext, target: unknown) {
```
Tatsinnit, line 11: 💡 Please note: for wider readers: `unknown is the type-safe counterpart of any`

webview-ui/package.json
```diff
@@ -5,6 +5,8 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
+    "lint": "eslint -c .eslintrc.cjs ./src",
```
Tatsinnit, line 8: 💡 I think this is probably part of `nom` regular build.

Tatsinnit (approved):
❤️☕️🎊 Lots of progressive minor changes which were necessity, give int HUGE I see most of them are the `nit` yet - great evangelistic - slinging changes much needed for repo - awesomeness.

I have added just comment as future reader of these changes.

We should merge this, happy to do it in few hours after this approval which gives more time for anyone else to present their thoughts.

Issue #183, opened 2022-12-30
💡 Fix Refresh Icon Via VsTool API.  (Behaviour and detail)


Tatsinnit:
This work item / GitHub behaviour correction is open to keep a note and find more permanent solution, essentially the issue is that refresh icon do not really refresh the node. I once opened this issue: #40 but at that point of time it was not much of a bigger deal as all features were on current existing feature functionality but now since we will be enabling other `Managed cluster functionality` we should cope a right fix for this. Things like `Create, delete` et. al. will need this for smoother transition.

**Current workaround**:

* Close the vscode and reopen it that way **vscode** will reload the cluster tree.

**Problem**:

* When I add new cluster under existing sub I donot have any way to refresh my sub for new cluster to appear I.e. Refresh icon in cloud explorer will resolve this issue.

**Steps to produce**:

* I added new cluster and had my explorer window open in extension.
* Once new AKS cluster is added, I tried to refresh the current sub in my extension.
* Result: Sub is not refreshed with new cluster.

**Possible solution**

* If I remember correctly the correct solution will reside in `tool-api` project ->  https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools-api/blob/master/ts/cloudexplorer/v1.ts 
* Or we can discuss suggestion here: https://github.com/Azure/vscode-aks-tools/pull/41#discussion_r490639105 

Thanks heaps cc and fyi ❤️☕️ @peterbom and @rzhang628 🎄☕️

**Screenshot below**: 

![image](https://user-images.githubusercontent.com/6233295/93402731-a51db280-f8d9-11ea-9ad1-3b1f3a37884b.png)

Thanks,

PR #335, opened 2023-11-22
Create cluster ux

hsubramanianaks:
This PR introduces UI/UX changes to create cluster experience along with presets similar to azure portal. Currently only Dev/Test preset is introduced in this PR and incrementally will introduce more presets based on user feedback and learnings. Thanks @peterbom and @Tatsinnit for tagging along with me for this feature. Thanks Ivelisse Capellan Heyer and Bonnie Jiang for the awesome UX designs. FYIP @qpetraroia @gambtho 

image


src/commands/aksCreateCluster/aksCreateCluster.ts
```diff
@@ -40,6 +40,7 @@ export default async function aksCreateCluster(_context: IActionContext, target:
         portalUrl,
         subscriptionTreeItem.subscription.subscriptionId,
```
peterbom, line 41: Rather than dereferencing the `subscription` object three times, I think it might be preferable to pass the object itself (i.e. passing `subscriptionTreeItem.subscription` directly).

src/panels/CreateClusterPanel.ts
```diff
@@ -31,7 +26,8 @@ export class CreateClusterDataProvider implements PanelDataProvider<"createClust
         readonly portalUrl: string,
         readonly subscriptionId: string,
```
peterbom, line 27: As suggested in an earlier comment, maybe replace these three subscription-related parameters with `readonly subscriptionContext: ISubscriptionContext,`

src/commands/utils/errorable.ts
```diff
@@ -89,6 +89,16 @@ export function getErrorMessage(error: unknown) {
     return String(error);
 }
 
+export function getInnerErrorMessage(error: unknown) {
```
peterbom, line 92: I'm not sure we should be disabling the linting. Let me have a play around and try to make sense of this error structure. Unless it's documented somewhere?

src/panels/templates/DevTestCreateCluster.json
peterbom, line 1: Resources like this should go in the `resources` folder. Partly for consistency, and partly because (I think) the webpack copying mechanism means that files under `src` get excluded from the published vsix.

webview-ui/src/CreateCluster/CreateClusterInput.tsx
```diff
@@ -137,21 +139,21 @@ export function CreateClusterInput(props: CreateClusterInputProps) {
...
+                                value={`(New) ${newResourceGroup.name} (${newResourceGroup.location})`}
                             ></VSCodeTextField>
                             <VSCodeButton className={styles.sideControl} onClick={() => setNewResourceGroup(null)}>
                                 Clear
                             </VSCodeButton>
                         </>
-                    )}
+                    )} */}
 
                     <label htmlFor="name-input" className={styles.label}>
```
peterbom, line 155: After discussion with the Portal team, I think we agreed to provide a Location drop-down, didn't we? Is this planned for a separate PR? If so, we should probably create a work item for it.

src/panels/utilities/ClusterSpecCreationBuilder.ts
```diff
@@ -0,0 +1,247 @@
+import { Deployment } from "@azure/arm-resources";
+
+import fs from 'fs';
```
peterbom, line 3:
```typescript
import devTestTemplate from "../templates/DevTestCreateCluster.json";
```

(and we can remove the 'fs' and 'path' imports)

src/panels/utilities/ClusterSpecCreationBuilder.ts
```diff
@@ -0,0 +1,247 @@
...
+            "userEmailAddress": {
+                value: clusterSpec.username
+            }
+        }
+
+
+        const deploymentParameters: Deployment = {
+            "properties": {
+                "parameters": parameters,
+                "template": loadTemplate("DevTestCreateCluster.json"),
```
peterbom, line 65:
```typescript
"template": devTestTemplate
```
(because we imported `devTestTemplate` at the top of the file)

src/panels/CreateClusterPanel.ts
```diff
@@ -162,11 +164,32 @@ async function createCluster(
...
+        kubernetesVersion: kubernetesVersion.values[0].version, // selecting the latest version since versions come in descending order
+        username: username
+    };
+
+    const deploymentSpec = getManagedClusterSpec(clusterSpec, preset);
 
     try {
-        const poller = await containerServiceClient.managedClusters.beginCreateOrUpdate(group.name, name, clusterSpec);
-        poller.onProgress((state) => {
+        const poller = await resourceManagementClient.deployments.beginCreateOrUpdate(group.name, name, deploymentSpec);
```
peterbom, line 191: I was thinking it'd be a good idea to call the deployment validation API first - what do you think?

src/commands/utils/errorable.ts
peterbom: Okay, error types seem pretty poorly documented. The best I could find was [this](https://learn.microsoft.com/en-us/javascript/api/@azure/core-http/resterror?view=azure-node-latest#@azure-core-http-resterror-details), which simply lists the `details` property as being of type `unknown`.

Here is how I'd code it. It's a bit lengthy, but type-safe and descriptive:

```typescript
    } catch (ex) {
        const errorMessage = isInvalidTemplateDeploymentError(ex)
            ? getInvalidTemplateErrorMessage(ex)
            : getErrorMessage(ex);
        // ...
    }

function getInvalidTemplateErrorMessage(ex: InvalidTemplateDeploymentRestError): string {
    const innerDetails = ex.details.error?.details || [];
    if (innerDetails.length > 0) {
        const details = innerDetails.map((d) => `${d.code}: ${d.message}`).join("\n");
        return `Invalid template:\n${details}`;
    }

    const innerError = ex.details.error?.message || "";
    if (innerError) {
        return `Invalid template:\n${innerError}`;
    }

    return `Invalid template: ${getErrorMessage(ex)}`;
}

type InvalidTemplateDeploymentRestError = RestError & {
    details: {
        error?: {
            code: "InvalidTemplateDeployment";
            message?: string;
            details?: {
                code?: string;
                message?: string;
            }[];
        };
    };
};

function isInvalidTemplateDeploymentError(ex: unknown): ex is InvalidTemplateDeploymentRestError {
    return isRestError(ex) && ex.code === "InvalidTemplateDeployment";
}

function isRestError(ex: unknown): ex is RestError {
    return isObject(ex) && ex.constructor.name === "RestError";
}
```

peterbom: I'd put all that in `CreateClusterPanel.ts` and remove it from `errorable.ts`, because it's very specific to a particular function in one library, and is not generally applicable to other errors.

src/panels/CreateClusterPanel.ts
```diff
@@ -190,10 +213,12 @@ async function createCluster(
                 });
             }
         });
-
         await poller.pollUntilDone();
     } catch (ex) {
-        const errorMessage = getErrorMessage(ex);
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
peterbom, line 218: See comment in `errorable.ts`

src/panels/utilities/ClusterSpecCreationBuilder.ts
```diff
@@ -0,0 +1,247 @@
...
+                "mode": "Incremental"
+            }
+        };
+        return deploymentParameters;
+    }
+
+}
+
+//eslint-disable-next-line @typescript-eslint/no-explicit-any
```
peterbom, line 206: If we import the json directly, we can delete this function.

webview-ui/src/CreateCluster/CreateCluster.module.css
```diff
@@ -2,13 +2,85 @@
...
+    height: 18px;
+    margin-bottom: 10px;
+}
+
+.presetContainer {
+    width: 217px;
+    height: 166px;
+    flex-shrink: 0;
+    border-radius: 8px;
+    border: 1px solid var(--base-base-10, #808080);
```
peterbom, line 18: We should avoid hard-coding colours. There are [lots of variables](https://github.com/Azure/vscode-aks-tools/blob/main/webview-ui/src/manualTest/vars.css) we can choose from for this.

src/commands/aksCreateCluster/aksCreateCluster.ts
hsubramanianaks: Done, updated.

src/commands/utils/errorable.ts
hsubramanianaks: @peterbom I expected this comments :),  I looked for the documentation for the types in the officials MS wiki's couldn't find one. I will look into your suggestion.

src/panels/samleDeploymentAks.json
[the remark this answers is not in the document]
hsubramanianaks: @peterbom I kept for some testing references. I will delete them. Thanks.

webview-ui/src/CreateCluster/CreateClusterInput.tsx
hsubramanianaks: @peterbom Yes even @qpetraroia brought this up. But if you see the resource groups are tied to location and cluster would be created in that location. Do we need separate location drop down again? Portal has them all in one single page both creation/selection of resource groups. In vscode we have them as separate dialog. Please let me know. 

image

src/panels/CreateClusterPanel.ts
hsubramanianaks: updated this. Thank you.

src/panels/utilities/ClusterSpecCreationBuilder.ts
```diff
@@ -0,0 +1,247 @@
...
+function generateDnsPrefix(clusterName: string): string {
+    // this replaces all white spaces in the string
+    clusterName = clusterName.replace(/\s+/g, '');
+    return validateDnsPrefix(clusterName);
+}
+
+function validateDnsPrefix(dnsPrefix: string): string {
+    // please refer https://aka.ms/aks-naming-rules
+    dnsPrefix.length > 54 ? dnsPrefix.substring(0, 54) : dnsPrefix;
+    // eslint-disable-next-line no-useless-escape
```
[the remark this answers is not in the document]
hsubramanianaks: Took care of it.

src/commands/utils/errorable.ts
hsubramanianaks: @peterbom Thanks for suggestion, now the eslint warnings are done. I will test this and resolve this conversation.

Tatsinnit:
💡🪄☕️ know issue - https://github.com/Azure/vscode-aks-tools/issues/183 (relevance et al explained)

src/panels/CreateClusterPanel.ts
```diff
@@ -203,25 +227,52 @@ async function createCluster(
...
-                count: 3,
-                enableNodePublicIP: true,
-                mode: "System",
-                osSKU: "AzureLinux",
-                osType: "Linux",
-                vmSize: "Standard_DS2_v2",
-            },
-        ],
-        dnsPrefix: `${name}-dns`,
+function getManagedClusterSpec(clusterSpec: ClusterSpec, preset: string): Deployment {
```
Tatsinnit, line 230: 💡☕️ Thoughts 💭 if this sounds super cool, shall we heavily document this function and  add some clarity as to even this says the aspects enterprise and various other versions it is not the full blown “prod”, “enterprise” et al aka portal obviously has dense options to make it more u/x driven.

This is super cool and what personally I think this does is to provide that seem less experience which leads to the IDE to stay as it is and not scattered which leads to the context changes which I think leads to break for the focused workflow. ❤️🙏

webview-ui/src/CreateCluster/CreateCluster.module.css
```diff
@@ -2,13 +2,85 @@
     display: grid;
     grid-template-columns: 8rem 20rem 8rem;
     grid-gap: 1rem;
+    margin-top: 18px;
+}
```
Tatsinnit, line 6: 💭🙈🙏 If we can find some resource really chill to kindly do css help or see what else better we can do will be awesome.

src/panels/CreateClusterPanel.ts
```diff
@@ -162,11 +164,32 @@ async function createCluster(
...
+        kubernetesVersion: kubernetesVersion.values[0].version, // selecting the latest version since versions come in descending order
+        username: subscriptionContext.userId
+    };
+
+    const deploymentSpec = getManagedClusterSpec(clusterSpec, preset);
 
     try {
-        const poller = await containerServiceClient.managedClusters.beginCreateOrUpdate(group.name, name, clusterSpec);
-        poller.onProgress((state) => {
+        const poller = await resourceManagementClient.deployments.beginCreateOrUpdate(group.name, name, deploymentSpec);
```
peterbom, line 191: We should probably [generate a unique name](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/deploy-cli#azure-deployment-template-name) for the deployment, rather than using the cluster name, to avoid earlier deployments getting overwritten by later ones.

This will probably only be a problem when things go wrong and users try to create the same cluster a second time...but those are precisely the times when users are likely to want a consistent deployment history.

src/panels/CreateClusterPanel.ts
peterbom: I can think of two reasons for performing validation first:
1. I don't believe ARM template deployment is atomic. I.e. if there's an error during deployment your resources are not rolled back to their previous state. So, performing validation first should reduce the risk of users ending up with a partially deployed set of resources.
2. A good validation API should return any validation errors in a well-documented, strongly-typed object that's returned directly (as the return type) from the validation function.

However... Here's what I've learned today:

The validation API does _not_ work as I expected it to. It _looks_ as if it returns a [`DeploymentValidateResult`](https://learn.microsoft.com/en-us/javascript/api/@azure/arm-resources/deploymentvalidateresult?view=azure-node-latest) containing an `error` property if it encounters validation errors. But actually it throws an error, which you have to drill into to extract the real error, just like we have to for the `beginCreateOrUpdate` `catch` block.

Given the identical error thrown between `beginValidate` and `beginCreateOrUpdate`, it appears there's some overlap between the validation that happens in those functions. I guess ARM delegates to each service's RP to perform validation, and the AKS RP does the same validation in both cases. And, for the dev/test cluster at least, all the resources depend on the cluster, so if the cluster creation fails then other resources won't be created anyway. So I think my reason (1) for performing the validation first is moot.

@Tatsinnit - any thoughts about this? I _think_ all this means that an extra validation step is not strictly necessary, but maybe at least a comment in the code saying why we're not doing that would be good.

src/panels/utilities/ClusterSpecCreationBuilder.ts
```diff
@@ -0,0 +1,236 @@
...
+                "template": devTestTemplate, //TODO - change this to the correct template
+                "mode": "Incremental"
+            }
+        };
+        return deploymentParameters;
+    }
+
+}
+
+function generateDnsPrefix(clusterName: string): string {
```
peterbom, line 204: I would avoid calling a validate function that throws an error, because it won't be handled anywhere. We've made a decision not to do explicit validation, and instead allow the deployment to fail if anything is wrong.

We can ensure the DNS prefex adheres to naming rules with a function like this:
```typescript
function generateDnsPrefix(clusterName: string): string {
    return clusterName
        .replaceAll(/[^a-z0-9-\\s]/gi, "")
        .replace(/^-/, "")
        .replace(/-$/, "")
        .substring(0, 54);
}
```

src/panels/CreateClusterPanel.ts
```diff
@@ -203,25 +227,52 @@ async function createCluster(
...
-                vmSize: "Standard_DS2_v2",
-            },
-        ],
-        dnsPrefix: `${name}-dns`,
+function getManagedClusterSpec(clusterSpec: ClusterSpec, preset: string): Deployment {
+    const specBuilder: ClusterSpecBuilder = new ClusterSpecBuilder();
+    switch (preset) {
+        case "dev":
+            return specBuilder.buildDevTestClusterSpec(clusterSpec);
+        case "economy":
```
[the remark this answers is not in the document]
Tatsinnit: Totally awesomeness to use `union` 🥷🎊 nice!!

src/panels/CreateClusterPanel.ts
```diff
@@ -162,11 +164,32 @@ async function createCluster(
         errorMessage: null,
     });
 
-    const clusterSpec = getManagedClusterSpec(location, name);
+    // kubernetes version is required to create a cluster via deployments
+    const kubernetesVersion = await containerServiceClient.managedClusters.listKubernetesVersions(location);
+    if (!kubernetesVersion || !kubernetesVersion.values || kubernetesVersion.values.length === 0 || !kubernetesVersion.values[0].version) {
```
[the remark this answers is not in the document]
Tatsinnit: Thanks heaps. just a note: this `isDefault` if we can add note or comment in this PR for doc reference, in case this is changed in coming future et. al. we don't need to search olde reference, while we already know where the docs reside.

src/commands/utils/errorable.ts
hsubramanianaks: This was tested and now when the error happens, webview displays the correct inner exception details to the user. Thank you.

src/panels/CreateClusterPanel.ts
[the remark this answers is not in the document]
hsubramanianaks: @Tatsinnit Yes correct, it is equivalent to creating dev/test preset from portal. To keep both clients vscode and portal in par we moved towards deployments. Plus, it creates extra resources like action group, alerts for the user whoever is creating the cluster.

src/panels/CreateClusterPanel.ts
hsubramanianaks: I agree, right now we have only one preset dev/test, I think we can document heavily once we introduce the other presets in the future releases.

src/panels/CreateClusterPanel.ts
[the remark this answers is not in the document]
hsubramanianaks: @peterbom Thanks for the PR. this is updated.

src/panels/CreateClusterPanel.ts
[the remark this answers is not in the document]
hsubramanianaks: Updated the PR.

webview-ui/src/CreateCluster/CreateCluster.module.css
hsubramanianaks: Thanks for the suggestions, I will look more into this.

hsubramanianaks: This is take care in the PR - https://github.com/hsubramanianaks/vscode-aks-tools/pull/4

webview-ui/src/CreateCluster/CreateCluster.module.css
hsubramanianaks: PR https://github.com/hsubramanianaks/vscode-aks-tools/pull/4 take care of this.

src/panels/CreateClusterPanel.ts
hsubramanianaks: I think we can agree we don't want validation api which is of not much use for this enhancement ? do you guys agree @Tatsinnit @peterbom ? Please confirm.

src/panels/CreateClusterPanel.ts
hsubramanianaks: @Tatsinnit  agree ?

Tatsinnit: 👍 absolutely with clarity that this feature is quick create cluster feature from users favourite editor aka vscode and simplifies the experience. (Which means portal will have full blown functionality available if anyone keen)

src/panels/CreateClusterPanel.ts
peterbom: I agree (for now).

Just want to note, though, in case we ever revisit this discussion: I certainly would consider adding it in the future, especially for the other presets where we're deploying more resources.

Tatsinnit (approved):
❤️🎊🙏 Congrats @hsubramanianaks and @peterbom for the effort, looks good to me, the final go-ahead is something I will leave that on @peterbom.

ivelisseca:
[heart]         Ivelisse Capellan Heyer reacted to your message:
________________________________

PR #346, opened 2023-11-28
fix dependabot types/node update.

Tatsinnit:
Fix for #340 thanks , cc: @peterbom 

Please note: Node version equivalent to https://github.com/microsoft/vscode/blob/b34399960a13aa0a980aba8d3e16d57c013fa23c/remote/.yarnrc#L2 

Thanks.


package.json
```diff
@@ -428,7 +428,7 @@
         "@types/glob": "^8.1.0",
         "@types/js-yaml": "^4.0.9",
         "@types/mocha": "^10.0.6",
-        "@types/node": "^16.11.7",
+        "@types/node": "^20.10.0",
```
peterbom, line 431: This should probably be in sync with the node version we're actually using. Otherwise we risk using node features that look like they're available from the types, but aren't actually, leading to runtime errors.

To enforce that, I think we should:
- Specify the node version in the [`engines` section](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines).
- Specify [`engine-strict: true`](https://docs.npmjs.com/cli/v10/using-npm/config#engine-strict) in an [`.npmrc` file](https://docs.npmjs.com/cli/v10/configuring-npm/npmrc#per-project-config-file).
- Specify `node-version: 20` in `build.yml` and `publish.yml`.

peterbom (approved):

PR #358, opened 2023-12-05
Move portal URL generation into a common function

peterbom:
We have a lot of duplicate code for generating URLs representing Azure resources in the Portal. There is some complexity to this, around ensuring we use the appropriate domain name for the user's cloud, and setting the correct query strings for identifying the referrer.

This moves that logic into a single function, which I thought would be really useful for #335.

**EDIT**: This has now been rebased on top of `main` after merging #335. It also includes a small change to display a link to the Portal for the deployment while the cluster creation is in progress.

~~Unfortunately, this had a far larger blast radius than I anticipated. Even more unfortunately, it conflicts heavily with #335 which is otherwise almost ready to merge.~~

~~So, I'm opening this as a draft to make clear the _kind_ of changes we'd need to make, and @hsubramanianaks can make use of any of these changes that look useful for displaying the Portal link in Create Cluster. I can tidy this up and add the remaining bits (mostly around Detectors) after #335 is merged.~~

~~@hsubramanianaks, @Tatsinnit: what do you think?~~

### DIFF
--- a/src/commands/aksCreateCluster/aksCreateCluster.ts
+++ b/src/commands/aksCreateCluster/aksCreateCluster.ts
@@ -33,12 +33,10 @@ export default async function aksCreateCluster(_context: IActionContext, target:
 
     const resourceManagementClient = getResourceManagementClient(subscriptionTreeItem);
     const containerServiceClient = getContainerClient(subscriptionTreeItem);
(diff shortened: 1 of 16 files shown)


Tatsinnit (approved):
🎊☕️ Looks good to me, it will be worth to get quick test for most of these links. Rest looks pretty solid, I tried my best to follow the `armid` et. al. changes.

Issue #347, opened 2023-12-01
Compare two or more aks clusters or any resources for their metadata and properties.


sprab:
**Is your feature request related to a problem? Please describe.**
New feature.

**Mention what platform you want to support the new feature**
VScode, portal and/or external.

**Describe the solution you'd like**
A feature to compare two or more clusters for their metadata, properties/attributes that will allow us to draw conclusions while troubleshooting issues as what is different between two clusters (working or non-working scenarios).

**Describe alternatives you've considered**
There are no alternate solutions apart from being manually compare them side by side on two different windows.

**Additional context**

Tatsinnit:
**Docs**: Get Jason and then use `vscode` - `diff`.

* https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/get?view=rest-aks-2023-08-01&tabs=HTTP

Implementation level details: (High level) 

####  Drop -1: Compare 2 cluster - aka - 2 cluster compare.

* U/X of some Sort (I can add pic latter of a sketch)
* On hit homage, we do `get` or `rest operation` on the both the clusters like: https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/get?view=rest-aks-2023-08-01&tabs=HTTP 
    * Then formulate the `temp` files for both cluster.
* Then use `code.diff` docs here : https://code.visualstudio.com/api/references/commands

Thanks heaps,  fyi => @peterbom and @sprab 

#### Sample only below.

<img width="1238" alt="Screenshot 2023-12-04 at 2 00 25 PM" src="https://github.com/Azure/vscode-aks-tools/assets/6233295/7f318727-3e87-4c2d-a54b-7f351b8f5c8d">

<img width="1770" alt="Screenshot 2023-12-04 at 1 56 05 PM" src="https://github.com/Azure/vscode-aks-tools/assets/6233295/32b6ddc4-3fbd-4a65-81e6-5c28228f5adc">

Tatsinnit:
Thank you @sprab , I Think we can move this to discussion for the further enhancement aspects you have and the internal work which shows the next level of this enhancement to be discussed along with wider folks like @peterbom. 

I thought I will drop this fyi comment as #363 is now merged as phase-0 to attain the quick way to compare 2 clusters.

PR #363, opened 2023-12-11
Feature: compare 2 aks cluster

Tatsinnit:
## ☕️🔔 Phase 0 of compare cluster - Simplistic Approach.

This PR is to take care the initial mindset towards #347 , the key aspect of this feature is to take care of a simple scenario which at the moment needs more then few steps to compare 2 cluster configuration.

💡 Scenario: At present if I have 2 cluster and I think they both are of exactly same configuration and one stop working then as a very initial step, I would love to compare the 2 first that my claim actually is true or not, for that one known route to do this comparison is that as a user I will first fetch the json configuration of both cluster save in respective files and then use some third party tool to compare those 2 file.

💡🎊 **_One-Click Solution:_** This is first phase of that compare cluster which enables user to simply select their cluster to compare `with` and compare `from` and sit-back and relax for true potential of the user's fav editor vscode to show the diff.

Thank you so much @sprab , @peterbom - cc: @gambtho as fyi ❤️🙏

_Screenshot to explain this is display below._

Screenshot 2023-12-07 at 10 01 13 AM

Screenshot 2023-12-08 at 10 27 40 PM

Screenshot 2023-12-08 at 10 27 49 PM

Screenshot 2023-12-08 at 10 28 02 PM


.vscode/settings.json
```diff
@@ -10,5 +10,8 @@
     "eslint.workingDirectories": [".", "./webview-ui"],
     "editor.defaultFormatter": "esbenp.prettier-vscode",
     "editor.formatOnSave": true,
-    "cloudcode.azureExplorerVisible": true
+    "cloudcode.azureExplorerVisible": true,
+    "[typescript]": {
```
peterbom, line 14: I think this might conflict with `editor.defaultFormatter` above. Was the prettier formatter not working correctly?

src/commands/utils/detectors.ts
```diff
@@ -97,7 +97,6 @@ export async function getDetectorInfo(
 }
 
 export function getPortalUrl(clusterdata: ARMResponse<unknown>) {
```
peterbom, line 99: This looks like just a formatting change, and I think it'll cause a merge conflict with #358 - shall we skip this change?

src/commands/aksCompareCluster/aksCompareCluster.ts
```diff
@@ -0,0 +1,123 @@
...
+        return;
+    }
+    const containerServiceClient = getContainerClient(subscriptionTreeItem);
+    const clusterList: string[] = [];
+    const clusterGroupList: Dictionary<string> = {};
+
+    await longRunning(`Getting AKS Cluster list for ${subscriptionTreeItem.name}`, async () => {
+        const iterator = containerServiceClient.managedClusters.list();
+        for await (const cluster of iterator.byPage()) {
+            clusterList.push(...cluster.map(c => c.name ?? ''));
```
peterbom, line 55: I don't think I actually knew about the `??` operator!

There is a slight inconsistency here, though: on line 55 you handle the `name` property being null or undefined, but on line 59 you assert it's non-null with the `!` operator.

Minor thing: I prefer to use `Array.forEach` instead of `map` for operations that perform mutations. It doesn't really matter, but I think `map` is more typically used for immutable operations that return a value.

Oh, and I prefer the name `clusterResourceDictionary` that you used below.

We could do something like:
```typescript
const clusterResourceDictionary: Dictionary<string> = {};
for await (const clusters of iterator.byPage()) {
    const validClusters = clusters.filter(c => c.id && c.name);
    clusterList.push(...validClusters.map(c => c.name!);
    validClusters.forEach(c => {
        clusterResourceDictionary[c.name!] = c.id!;
    });
}
```

src/commands/aksCompareCluster/aksCompareCluster.ts
```diff
@@ -0,0 +1,123 @@
...
+    target: unknown
+): Promise<void> {
+    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
+
+    const cluster = getAksClusterSubscriptionItem(target, cloudExplorer);
+    if (failed(cluster)) {
+        vscode.window.showErrorMessage(cluster.error);
+        return;
+    }
+    const subscriptionTreeItem = <SubscriptionTreeItem>cluster.result;
```
peterbom, line 41: I've not noticed this before, and I know there's a lot of code that already uses this, so we probably shouldn't change it here...but ideally we wouldn't need a cast here. We can revisit this later.

src/commands/aksCompareCluster/aksCompareCluster.ts
```diff
@@ -0,0 +1,123 @@
...
+    };
+
+    const state = await runMultiStepInput('Compare AKS Cluster', initialState, clusterGroupCompareWithStep, clusterGroupCompareFromStep);
+    if (!state) {
+        // Cancelled
+        return;
+    }
+
+    // Call compare cluster at this instance
+    await compareManagedCluster(state, clusterGroupList, <SubscriptionTreeItem>cluster.result)
```
peterbom, line 96: I think we can use `subscriptionTreeItem` here, to avoid another cast - can we?

src/commands/aksCompareCluster/aksCompareCluster.ts
Tatsinnit: Sounds awesome, thank you! ☕️🥷 I have made rest of the recommended changes.

peterbom (approved):
Looks great!

PR #375, opened 2023-12-13
Add filters for TCPDump

peterbom:
Capture files can become huge very quickly. We need to allow users to filter output so that file sizes are manageable.

This is achieved with a new `Filters` option:
image

Users can filter on network interface, as well as build a [pcap filter string](https://www.tcpdump.org/manpages/pcap-filter.7.html) using protocol/port fields, or editing it manually:
image

The number of options here could grow quite large and become mutually inconsistent, depending on which capture scenarios we want to support in future. For example, collecting outgoing traffic from a _single_ pod requires different filter options from traffic _between_ two known pods. Conceivably we might want to support other scenarios in the future, perhaps involving services or ingresses.

For this reason, we have also added a `Capture Scenario` drop-down. Choosing different values here gives you different additional filter options, and allows us to add future filters in a consistent manner. For now we support "Specific Pod" and "Two Pods":
image
image

Note that the `Pcap Filter` string gets updated with the correct IP addresses of the selected pods. `tcpdump` behaviour in response to these filters can be verified from the downloaded capture file:
image

**Known limitation**: The `Pcap Filter` string allows freeform text because the grammar is too complex for us to validate. If the user enters an invalid filter string there will be an error running the `tcpdump` command. Unfortunately, we run that command in the background (because it's long-running and we need to return control to the user and allow them to stop the capture), meaning we don't catch any non-zero exit code, and the capture appears to be running when it's not. The user will not notice until they try to _stop_ the capture, and there will be an error about being unable to find it. This is solvable, for example we can poll for running `tcpdump` processes and detect if it has exited early, but I'd prefer to make that change in a separate PR.

### DIFF
--- a/src/panels/TcpDumpPanel.ts
+++ b/src/panels/TcpDumpPanel.ts
@@ -8,8 +8,11 @@ import { MessageHandler, MessageSink } from "../webview-contract/messaging";
 import { BasePanel, PanelDataProvider } from "./BasePanel";
 import { KubectlVersion, getExecOutput, invokeKubectlCommand } from "../commands/utils/kubectl";
 import {
+    CaptureFilters,
     CompletedCapture,
+    FilterPod,
     InitialState,
+    NodeName,
     ToVsCodeMsgDef,
     ToWebViewMsgDef,
 } from "../webview-contract/webviewDefinitions/tcpDump";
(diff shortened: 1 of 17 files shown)


Tatsinnit (approved):
❤️🥷 Looks SOLID :) let's get this in. I have tested this as well, with a bad cluster scenario et. al.

Issue #376, opened 2023-12-13
Telemetry for some webview commands


peterbom:
As we've moved commands out of the context menu and into webviews, we've lost a lot of granularity from our telemetry. For example, we no longer know _which_ `kubectl` or Inspektor Gadget commands are being run, or whether users are invoking cluster operations from webviews. Instead we can only see that a webview has been launched, even though users may not have taken any action from it.

It would be really useful to log telemetry about actions taken from webviews. This can be determined from the messages sent from webviews to VS Code.

**However**... We want to be _very_ careful about what data we're collecting here. For privacy reasons we _cannot_ simply collect the contents of each message sent to VS Code. Defining what gets collected needs to be clear, explicit, and opt-in only, so that we can easily see and review what is being collected.

**Proposal**: We extend the `PanelDataProvider` interface so that each implementor (one per webview) has to implement a function that specifies what telemetry is collected (if any) for each message to VS Code in that webview, e.g.:
```diff
export interface PanelDataProvider<TContent extends ContentId> {
    getTitle(): string;
    getInitialState(): InitialState<TContent>;
+    getTelemetryDefinition(): TelemetryDefinition<TContent>;
    getMessageHandler(webview: ToWebviewMessageSink<TContent>): ToVsCodeMessageHandler<TContent>;
}
```

An implementation of this might look like (for 'Create Cluster'):
```typescript
    getTelemetryDefinition(): TelemetryDefinition<"createCluster"> {
        return {
            getResourceGroupsRequest: false,  // No telemetry for this message
            getLocationsRequest: false,  // No telemetry for this message
            cancelDeploymentRequest: true,  // Log that the command was run, but no other data
            createClusterRequest: (args) => ({
                location: args.location,  // Illustration purposes only: in practice we probably wouldn't log `location`
                preset: args.preset,
            }),
        };
    }
```

[Edited above to allow `boolean` values to indicate whether to log whether a command is run, with no other metadata]

peterbom:
It seems like we can't specify _any_ properties that are "unknown". Only `command` seems to be permitted. The others get stripped out:
```
"DataHandlingTags": unknown_props;dropped_unclassified_props;ip2city;dropped_null_props;delinked_macaddresshash;dropped_suppressed_props;,
```

This means it's easy to see, say, an Inspektor Gadget command was run:
```
"command": "gadget.runStreamingTraceRequest"
```

[...]

PR #377, opened 2023-12-14
Support telemetry for commands from webviews

peterbom:
Addresses #376 

~~We need further discussion about how best to structure the properties that we emit: @Tatsinnit, @qpetraroia, @hsubramanianaks so opening this as draft for now.~~

Some background:
- Commands from webviews are invoked by sending a message from the webview to VS Code.
- These messages are well-known and explicitly defined for each webview. For example, the `createCluster` webview uses three messages:
  - `getResourceGroupsRequest`
  - `getLocationsRequest`
  - `createClusterRequest`
- We probably **don't** want to emit telemetry for all of these events. In the above case we'd probably want an event for `createClusterRequest` only.
- We have some messages where the name itself is probably not sufficiently granular, and we'd need to extract some content from the message in order for the telemetry to be useful. This is the case for Inspektor Gadget commands, in which all we can tell from the message name is whether it was a 'streaming' or 'blocking' command. What we actually care about is the _category_ (e.g. `trace`, `snapshot`, `top`, etc.) and the _resource_ (`dns`, `tcp`, etc.).
- The changes here will allow us to define ~~which properties will be emitted~~ _custom command names_ for each message. See comment in `webviewTypes.ts`:
```
 * Possible values are:
 * - false: No telemetry will be emitted
 * - true: A telemetry event will be emitted containing the property "command" with value "<webview>.<messagetype>"
 * - (args) => string: A telemetry event will be emitted containing the property "command" with value "<webview>.<returnValue>"
 *                     where `returnValue` is the command name returned from the specified function.
```
- So, a `trace dns` Inspektor Gadget command will result in an event like:
```
EventName: "vscode-aks-tools/command"
Properties: {
   command: "gadget.streamTrace_trace_dns",
   ...commonProperties
}
```

**Earlier discussion below**:

I think the main thing that warrants further discussion is the structure of the properties. Do we want different property names that are specific to certain messages, or do we want a well-known property name, like `commandDetail`, which would squash all relevant properties into a single string, e.g. `commandDetail: "trace dns"`?

There are trade-offs here between maintenance and ease of querying, and I don't think the answer is obvious.

**EDIT**: Another option here is to _only_ allow the telemetry definition to override the `command` value, returning just a string, not an object. E.g. for the IG scenario, the definition could be something like:
```typescript
    startStreamingTraceRequest: (args) => `runCommand_${args.category}_${args.resource}`,
```
And the event would look like:
```
EventName: "vscode-aks-tools/command"
Properties: {
   command: "gadget.runCommand_trace_dns"
}
```
This would also save us doing _any_ GDPR classification.

### DIFF
--- a/src/panels/BasePanel.ts
+++ b/src/panels/BasePanel.ts
@@ -1,6 +1,7 @@
 import { Disposable, Webview, window, Uri, ViewColumn } from "vscode";
 import {
     CommandKeys,
+    Message,
     MessageDefinition,
     MessageHandler,
     MessageSource,
@@ -139,14 +140,8 @@ function getMessageContext<TContent extends ContentId>(
                         throw new Error(`Invalid message to VsCode: ${JSON.stringify(message)}`);
                     }
 
-                    const getTelemetryData = telemetryDefinition[message.command];
-                    if (getTelemetryData !== false) {
-                        const commandTelemetryData =
-                            getTelemetryData === true ? {} : getTelemetryData(message.parameters);
-                        const telemetryData = {
-                            ...commandTelemetryData,
-                            command: `${contentId}.${message.command}`,
-                        };
+                    const telemetryData = getTelemetryData(contentId, telemetryDefinition, message);
+                    if (telemetryData !== null) {
                         reporter.sendTelemetryEvent("command", telemetryData);
                     }
 
@@ -164,4 +159,22 @@ function getMessageContext<TContent extends ContentId>(
     };
 
     return { ...sink, ...source };
+}
+
+function getTelemetryData<TContent extends ContentId>(
+    contentId: TContent,
+    telemetryDefinition: TelemetryDefinition<TContent>,
+    message: Message<ToVsCodeMsgDef<TContent>>,
+): { [key: string]: string } | null {
+    const getTelemetryData = telemetryDefinition[message.command];
+
+    // getTelemetryData is either `true` or a function returning a properties object.
+    if (getTelemetryData === false) return null;
+
+    // The `command` value we emit will combine the webview identifier (contentId), e.g. `createCluster`
+    // with the command in the message, e.g. `createClusterRequest`.
+    const data = { command: `${contentId}.${message.command}` };
+    if (getTelemetryData === true) return data;
+
+    return { ...data, ...getTelemetryData(message.parameters) };
 }
--- a/src/panels/BasePanel.ts
+++ b/src/panels/BasePanel.ts
@@ -13,13 +13,14 @@ import { encodeState } from "../webview-contract/initialState";
 import {
     ContentId,
     InitialState,
-    ToVsCodeMessage,
+    TelemetryDefinition,
     ToVsCodeMessageHandler,
     ToVsCodeMsgDef,
     ToWebviewMessageSink,
     ToWebviewMsgDef,
     VsCodeMessageContext,
 } from "../webview-contract/webviewTypes";
+import { reporter } from "../commands/utils/reporter";
 
 const viewType = "aksVsCodeTools";
 
@@ -31,6 +32,7 @@ const viewType = "aksVsCodeTools";
 export interface PanelDataProvider<TContent extends ContentId> {
     getTitle(): string;
     getInitialState(): InitialState<TContent>;
+    getTelemetryDefinition(): TelemetryDefinition<TContent>;
     getMessageHandler(webview: ToWebviewMessageSink<TContent>): ToVsCodeMessageHandler<TContent>;
 }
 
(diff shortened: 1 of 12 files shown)


src/panels/BasePanel.ts
```diff
@@ -124,10 +135,21 @@ function getMessageContext<TContent extends ContentId>(
...
                         throw new Error(`Invalid message to VsCode: ${JSON.stringify(message)}`);
                     }
 
+                    const getTelemetryData = telemetryDefinition[message.command];
+                    if (getTelemetryData !== false) {
+                        const commandTelemetryData =
+                            getTelemetryData === true ? {} : getTelemetryData(message.parameters);
+                        const telemetryData = {
+                            ...commandTelemetryData,
+                            command: `${contentId}.${message.command}`,
```
hsubramanianaks, line 148: @peterbom It would be helpful if you can give a sample value or example of contentId and message.command just for others to know what will be logged/captured as part of telemetry.

src/panels/BasePanel.ts
```diff
@@ -124,10 +135,21 @@ function getMessageContext<TContent extends ContentId>(
         subscribeToMessages: (handler) => {
             webview.onDidReceiveMessage(
                 (message: object) => {
-                    if (!isValidMessage<ToVsCodeMessage<TContent>>(message)) {
+                    if (!isValidMessage<ToVsCodeMsgDef<TContent>>(message)) {
                         throw new Error(`Invalid message to VsCode: ${JSON.stringify(message)}`);
                     }
 
+                    const getTelemetryData = telemetryDefinition[message.command];
+                    if (getTelemetryData !== false) {
```
hsubramanianaks, line 143: @peterbom I am confused here , you are checking for !== false and then inside you check for === true which is like conflicting. Can you elaborate more on this ?

src/panels/CreateClusterPanel.ts
```diff
@@ -46,6 +47,14 @@ export class CreateClusterDataProvider implements PanelDataProvider<"createClust
         };
     }
 
+    getTelemetryDefinition(): TelemetryDefinition<"createCluster"> {
+        return {
+            getResourceGroupsRequest: false,
+            getLocationsRequest: false,
+            createClusterRequest: true,
```
hsubramanianaks, line 54: @peterbom I am assuming you are still working on this PR so that we are NOT logging whole createClusterRequest here correct ? we might need just args like preset selected, command was run. nothing else.

hsubramanianaks: @peterbom never mind, got it. if it just true, it would log just the command is ran. make sense. Thanks.

src/panels/BasePanel.ts
hsubramanianaks: @peterbom I got it now, for a second it was confusing. But still adding some comments here would be helpful for the readers. :)

src/panels/BasePanel.ts
peterbom: I've updated the PR description, but I'm guessing you mean in the code? Yes, that makes sense. Let me add...

src/panels/BasePanel.ts
peterbom: I agree, that doesn't read well. I'll see if I can write it more clearly!

Just to explain quickly for now, though: the value of `getTelemetryData` is of type `boolean | `(args: MsgArgs) => { [key: string]: string }`

We only emit telemetry if it's `true` or a function is specified, in other words if `getTelemetryData !== false`. Once inside that condition, it must be either `true` or be a function.

src/panels/CreateClusterPanel.ts
peterbom: Absolutely! I realise my original PR description wasn't clear, so I updated that. In short, the `createClusterRequest: true` here means an event will be emitted containing:
```
EventName: "vscode-aks-tools/command"
Properties: {
   command: "createCluster.createClusterRequest"
}
```
There won't be any other data in there.

src/panels/BasePanel.ts
peterbom: Updated

src/panels/BasePanel.ts
peterbom: Extracted this logic into a separate function with separate `return` statements for each condition. I _think_ this makes it clearer.

src/panels/BasePanel.ts
hsubramanianaks: @peterbom Thank you looks great !!

src/panels/BasePanel.ts
hsubramanianaks: looks good to me. Thank you.

Tatsinnit (approved):

PR #378, opened 2023-12-14
Fix occasional error message that npm is missing

peterbom:
I have been getting occasional errors launching debug profiles, that I can't reproduce reliably.

The error was `could not find the task "npm: webpack"`. Searching for that error, I found several _similar_ sounding problems, like [this](https://github.com/microsoft/vscode/issues/147193), but in all the examples I could see, the `preLaunchTask` value referred to an entry in `tasks.json`, which appears to be the recommended way to do it.

Creating this PR because making the change myself appeared to fix the issue I was getting, and it seems to be the "right" approach.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,7 +18,7 @@
             "outFiles": [
                 "${workspaceFolder}/dist/**/*.js"
             ],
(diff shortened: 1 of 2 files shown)


Tatsinnit (approved):
🎊

PR #379, opened 2023-12-15
Use new telemetry library

peterbom:
We are currently relying on a [deprecated telemetry library](https://www.npmjs.com/package/vscode-extension-telemetry). This is now preventing us from using the latest version of other libraries which have dependencies on the recommended telemetry libraries.

These changes move us onto the latest library.

One slight complication here is that the new library no longer supports specifying the extension name. Instead, it works it out itself, constructing it from `{publisher}.{name}`, which in our case is `ms-kubernetes-tools.vscode-aks-tools`. In the current code, we're specifying it explicitly, without the publisher component. That means that after this change we'll be publishing events with a different name, since the full event name is `{extensionName}/{event}`.

In other words:
- current event name: `vscode-aks-tools/command`
- new event name: `ms-kubernetes-tools.vscode-aks-tools/command`

I thought maybe we could use the [Replacements API](https://github.com/microsoft/vscode-extension-telemetry/blob/4361d93aebb8fed61cc4a1657a91d05443fc27c0/dist/telemetryReporter.d.ts#L16-L26) to update the `common.extname` property, and that would update the event name accordingly, but that doesn't seem to be the way it works. So, I think we're stuck with the name change.


Tatsinnit (dismissed):
🥂🎉🙏❤️

Tatsinnit (approved):
❤️ Before release we can test this again as part of final `vsix` looks good to me, thank you so much.

PR #392, opened 2023-12-19
Update vscode-azext-utils package

peterbom:
We previously had to revert this in #374 because it depended on a newer telemetry library. Now we've updated our telemetry library, we can make this change.


hsubramanianaks (approved):
look good to me. I hope we have done testing to make sure everything works. Thank you.

Tatsinnit (approved):

PR #393, opened 2023-12-21
Refresh subscription

peterbom:
Addresses the comments in #41 to support refreshing subscriptions in Cloud Explorer without reloading the whole window.

Opening as Draft because I'm not sure why the earlier PR was closed. @Tatsinnit - do you remember?

This partially addresses #183, although that issue appears to refer to the refresh icon at the _root_ level of Cloud Explorer. The root-level refresh button only refreshes the list of providers, not the contents of each provider (subscriptions/clusters). This is [apparently by design](https://github.com/Azure/vscode-aks-tools/issues/40#issuecomment-693765433) and not addressed here.

Instead, this takes the approach used in the earlier PR, allowing users to refresh the clusters for individual subscriptions.

It also allows us to programmatically invoke the subscription-level `refresh` command after creating/deleting clusters, so that the treeview will automatically refresh. This can be done in a separate PR.


src/tree/subscriptionTreeItem.ts
```diff
@@ -37,6 +49,15 @@ export default class SubscriptionTreeItem extends SubscriptionTreeItemBase imple
         return aksClusterResources.map((aksClusterResource) => new AksClusterTreeItem(this, aksClusterResource));
     }
 
+    public async refreshImpl(): Promise<void> {
```
Tatsinnit, line 52: 💡🥷🎅🛺  yeah this probably is more sane than this: https://github.com/Azure/vscode-aks-tools/pull/41#discussion_r490639105 .I think it is more favourable to get this in for total experience for the end-users.

At some stage we should really plan to take over the over-all API around this and advancement we could do, just as spitballing ideas, the cross subscription stuff could be the line of thinking we could keep when we take a look at this API for sub tree.

Tatsinnit (approved):
❤️ Thank you so much for this. Looks favourable and less buggy, worth to get in, user always has option of closing and reopening the editor, just to note as archive this still will have those situations where is cluster is deleted or freshly created the ARM Call itself will take time to update the data so might take solid few mins to appear as refreshed feature into the tree.

Could make this part of release note, or might be mention in the comment somewhere. Feel free to convert this into the `ready for review` state. 

We can share VSIX to wider folks for test if they want to.

src/tree/subscriptionTreeItem.ts
peterbom: Yeah, I think that if we wanted the _cloud_-level refresh button to refresh the subscriptions, we'd need some change to the API, but we'd certainly need to think very carefully about what that change should be.

PR #415, opened 2024-01-03
Refresh subscriptions after adding and deleting clusters

peterbom:
Programmatically refresh the Cloud Explorer treeview after adding and deleting clusters.

This builds on the command to refresh the clusters under a subscription in the treeview (#393), and makes use of the slightly more constrained interface for interacting with TreeView items (#406).

### DIFF
--- a/src/commands/aksDeleteCluster/aksDeleteCluster.ts
+++ b/src/commands/aksDeleteCluster/aksDeleteCluster.ts
@@ -5,6 +5,8 @@ import { deleteCluster, getAksClusterTreeNode } from "../utils/clusters";
 import { failed, succeeded } from "../utils/errorable";
 import { longRunning } from "../utils/host";
 
+const refreshIntervals = [1, 2, 5, 10, 30, 60, 120];
+
 export default async function aksDeleteCluster(_context: IActionContext, target: unknown): Promise<void> {
     const cloudExplorer = await k8s.extension.cloudExplorer.v1;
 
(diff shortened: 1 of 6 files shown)


Tatsinnit (approved):
🎊❤️🙏☕️ Thank you for this. Looks good to me, for archive only, folks can argue about the intervals, I at this stage has no preference so looks good to me, if in future anyone wants to start a discussion, I think comment is in place for discussion.

peterbom:
Thanks @Tatsinnit!

> for archive only, folks can argue about the intervals, I at this stage has no preference so looks good to me, if in future anyone wants to start a discussion, I think comment is in place for discussion.

Good point. You're referring to the refresh intervals for refreshing after cluster deletion, right? Just for the record, when I tried it, the updated list came back after the 30s interval. I have no idea what kind of caching is going on internally, so it could conceivably take significantly less or more time. I think the current list avoids hammering the API too much, while still refreshing reasonably responsively. Worst case, the user can refresh manually with the refresh button.

Tatsinnit:
> Thanks @Tatsinnit!
> 
> > for archive only, folks can argue about the intervals, I at this stage has no preference so looks good to me, if in future anyone wants to start a discussion, I think comment is in place for discussion.
> 
> Good point. You're referring to the refresh intervals for refreshing after cluster deletion, right? Just for the record, when I tried it, the updated list came back after the 30s interval. I have no idea what kind of caching is going on internally, so it could conceivably take significantly less or more time. I think the current list avoids hammering the API too much, while still refreshing reasonably responsively. Worst case, the user can refresh manually with the refresh button.

src/commands/aksDeleteCluster/aksDeleteCluster.ts
```diff
@@ -33,6 +35,14 @@ export default async function aksDeleteCluster(_context: IActionContext, target:
 
         if (succeeded(result)) {
             vscode.window.showInformationMessage(result.result);
+
+            // Periodically refresh the subscription treeview, because the list-clusters API
+            // call still includes the cluster for a while after it's been deleted.
+            refreshIntervals.forEach((interval) => {
+                setTimeout(() => {
+                    vscode.commands.executeCommand("aks.refreshSubscription", clusterNode.result.subscriptionTreeNode);
+                }, interval * 1000);
```
hsubramanianaks, line 44: do we really need this periodic refresh instead user can refresh himself right ? just thinking,,

Issue #441, opened 2024-01-21
Azure Kubernetes Service extension can't connect to AKS clusters with same name but are in different resource groups 


SOFSPEEL:
**Describe the bug**
Have 2 clusters with the **same name** in the same subscription but in different resource groups
This works in version 1.3.18 and earlier version of **Azure Kubernetes Service** Extension, so this was broken in 1.4.0 version
e.g. 
Subscription: ACS-MGMT-DEV01

cluster 1: mobileservice
resource group: **eu1d**-mobileservice-aks

cluster 2: mobileservice
resource group: **eu1t**-mobileservice-aks

Note that the tool does not connect successfully to the second cluster and shows "invalid"

**Mention the platform you are using**
vscode
Version: 1.85.1 (Universal)
Commit: 0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2
Date: 2023-12-13T09:48:06.308Z
Electron: 25.9.7
ElectronBuildId: 25551756
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Darwin arm64 23.2.0

**To Reproduce**
1. Go to 'Kubernetes' CLOUDS/Azure section in vscode
2. Expand to  an Azure subscription that has multiples AKS with same name but are in different resource groups
3. Results in the cluster name 'invalid' for all but the first AKS cluster.

**Expected behavior**
Should connect to both clusters.

**Screenshots**
<img width="218" alt="image" src="https://github.com/Azure/vscode-aks-tools/assets/1612646/260a3f04-1757-4a20-a101-a6eb63102a10">

**Desktop (please complete the following information):**
```
  Model Name:	MacBook Pro
  Model Identifier:	Mac14,5
  Model Number:	MPHG3LL/A
  Chip:	Apple M2 Max
  Total Number of Cores:	12 (8 performance and 4 efficiency)
  Memory:	32 GB
  System Firmware Version:	10151.61.4
  OS Loader Version:	10151.61.4
```

Tatsinnit:
Thank you so much @SOFSPEEL for initiating this. ☕️❤️  I can readily replicate the steps provided, and I am exploring potential solutions. We will carefully examine the issue and strive to devise the best possible solution to address it effortlessly. Thank you for sharing; I'll be in touch soon.

[...]

PR #442, opened 2024-01-21
Fix for making instance when cluster name is same but RG are different.

Tatsinnit:
This Pull Request intend to fix the issue #441 with `AZTreeItem` object which is working within the `azext-utils` given earlier we were supplying `ClusterName` only as the unique identity I think it was filtering that as the duplicated hence the failure during the TreeNode Creation.

Below is the **before** and **after** of this issue after the fix, I played around, and ~~will share the `vsix` for quick play around~~. Here is the link for the `vsix` to test if this fixes the issue for you. (Please rename file after download this `zip` file and then rename to `vsix` and then try to install it manually. 

Here are help how to install:

- [Helpful guidance how to install vsix manually](https://code.visualstudio.com/docs/editor/extension-marketplace#_manage-extensions) OR [guidance 2 how to install manually in vscode](https://stackoverflow.com/questions/42017617/how-to-install-vs-code-extension-manually)

- **Actual VSIX** which fixes the issue (please download and remove *zip from this file before installation steps above) -> [vscode-aks-tools-1.4.0-uniquename-test1.vsix.zip](https://github.com/Azure/vscode-aks-tools/files/14031679/vscode-aks-tools-1.4.0-uniquename-test1.vsix.zip)

Thank you so much all. 

**Before**

image

**After**

image

### DIFF
--- a/src/commands/aksCompareCluster/aksCompareCluster.ts
+++ b/src/commands/aksCompareCluster/aksCompareCluster.ts
@@ -7,19 +7,21 @@ import * as vscode from "vscode";
 import * as k8s from "vscode-kubernetes-tools-api";
 import { SubscriptionTreeNode } from "../../tree/subscriptionTreeItem";
 import { getExtension } from "../utils/host";
-import { Dictionary } from "../utils/dictionary";
 import * as tmpfile from "../utils/tempfile";
 import { longRunning } from "../utils/host";
+import { ManagedCluster } from "@azure/arm-containerservice";
 
 interface State {
     clusterGroupCompareFrom: ClusterNameItem;
     clusterGroupCompareWith: ClusterNameItem;
     clustername: string;
+    armid: string;
     subid: string | undefined;
 }
 
 interface ClusterNameItem extends QuickPickItem {
     name: string;
+    armid: string;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/commands/aksCompareCluster/aksCompareCluster.ts
```diff
@@ -42,25 +44,21 @@ export default async function aksCompareCluster(_context: IActionContext, target
         return;
     }
     const containerServiceClient = getContainerClient(subscriptionNode.result);
-    const clusterList: string[] = [];
-    const clusterResourceDictionary: Dictionary<string> = {};
```
hsubramanianaks, line 46: @Tatsinnit you can just a leave a comment to mention why this file was changed, at first glance I thought this file had the bug then realized it was included as part of this PR to make the code simple and readable. Thanks.

hsubramanianaks (approved):
looks good to me. Thank you.

src/commands/aksCompareCluster/aksCompareCluster.ts
Tatsinnit: This is relating to kind testing from @sprab who actually find a small thing which needs adjustment.

Hence it will make this PR more simple and readable especially the `az compare` feature :) we have simplified the whole logic, but if you are wanting to understand the feature, let me know I can help to do that as well.

Hope this helps, thanks heaps for comment. fyi:  @hsubramanianaks, please let me know if you have any further query or want me to detail the eventual Simplification in detail for az compare feature.

hsubramanianaks: @Tatsinnit Thank you for the detailed explanation, looks clear to me.

sprab (approved):
The compare feature now works as expected. This is awesome.  We do surface the resourceId which against which the operation is done.  This helps clear confusion when there are clusters with same name across resource groups and clusters with same RG across subscriptions. Customers can now safely initiate these operations.

Issue #524, opened 2024-03-01
Update Inspektor Gadget version


blanquicet:
The Inspektor Gadget version used by default is 0.19.0, but next week Inspektor Gadget will already release v0.26.0. I suggest updating it to provide the new features released in all those versions.

Tatsinnit:
Nice! Thank you so much @blanquicet , ❤️🙏☕️  Updated to the latest available but regularly we will keep eye on latest versions and update accordingly, since Ita taken care by #525 will close it for now, but please do let us know if there is anything else.

PR #525, opened 2024-03-03
Upgrade the inspektor gadget to latest stable version.

Tatsinnit:
This specific PR caters #524 chelatest version of Inspector gadget from gh is `v0.25.1` hence using that. I have tested this and things are all behaving as normal.

https://github.com/inspektor-gadget/inspektor-gadget/releases

Thanks heaps, cc: @peterbom ☕️🥷

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
                 },
                 "azure.kubectlgadget.releaseTag": {
                     "type": "string",
-                    "default": "v0.19.0",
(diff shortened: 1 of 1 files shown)


peterbom (approved):
I'm guessing this newer version also includes some new gadgets or properties of existing ones. If we want to expose any of this new functionality we can include that in a separate PR.

PR #538, opened 2024-03-07
Feature/k8s version deprecated available

Tatsinnit:
This came out of an enhancement request @sprab please check this out, it helps resolving the whole is my closer version deprecated and if so show me quick info for what is available thing.

thanks heaps as always and fyi @sprab, @peterbom + @gambtho  ❤️🥷☕️

Screenshot 2024-03-07 at 6 25 45 PM

[vscode-aks-tools-1.4.1-test copy.vsix.zip](https://github.com/Azure/vscode-aks-tools/files/14521692/vscode-aks-tools-1.4.1-test.copy.vsix.zip)

### DIFF
--- a/src/commands/utils/clusters.ts
+++ b/src/commands/utils/clusters.ts
@@ -378,13 +378,12 @@ export async function getManagedCluster(
 export async function getKubernetesVersionInfo(
     client: azcs.ContainerServiceClient,
     location: string,
-    clusterName: string,
 ): Promise<Errorable<azcs.KubernetesVersionListResult>> {
     try {
         const managedCluster = await client.managedClusters.listKubernetesVersions(location);
         return { succeeded: true, result: managedCluster };
     } catch (e) {
-        return { succeeded: false, error: `Failed to retrieve cluster ${clusterName}: ${e}` };
+        return { succeeded: false, error: `Failed to list Kubernetes versions in ${location}: ${getErrorMessage(e)}` };
     }
 }
 
--- a/src/commands/utils/clusters.ts
+++ b/src/commands/utils/clusters.ts
@@ -375,6 +375,19 @@ export async function getManagedCluster(
     }
 }
 
+export async function getKubernetesVersionInfo(
+    client: azcs.ContainerServiceClient,
+    location: string,
+    clusterName: string,
+): Promise<Errorable<azcs.KubernetesVersionListResult>> {
+    try {
+        const managedCluster = await client.managedClusters.listKubernetesVersions(location);
+        return { succeeded: true, result: managedCluster };
+    } catch (e) {
+        return { succeeded: false, error: `Failed to retrieve cluster ${clusterName}: ${e}` };
+    }
+}
+
 export async function determineProvisioningState(
     clusterNode: AksClusterTreeNode,
     clusterName: string,
--- a/src/panels/ClusterPropertiesPanel.ts
+++ b/src/panels/ClusterPropertiesPanel.ts
@@ -225,7 +225,7 @@ function asClusterInfo(cluster: ManagedCluster, kubernetesVersionList: Kubernete
     return {
         provisioningState: cluster.provisioningState!,
         fqdn: cluster.fqdn!,
-        kubernetesVersion: cluster.kubernetesVersion!,
+        kubernetesVersion: cluster.currentKubernetesVersion!,
         powerStateCode: cluster.powerState!.code!,
         agentPoolProfiles: (cluster.agentPoolProfiles || []).map(asPoolProfileInfo),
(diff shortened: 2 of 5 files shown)


src/commands/utils/clusters.ts
```diff
@@ -375,6 +375,19 @@ export async function getManagedCluster(
     }
 }
 
+export async function getKubernetesVersionInfo(
+    client: azcs.ContainerServiceClient,
+    location: string,
+    clusterName: string,
```
peterbom, line 381: Can we remove the `clusterName` parameter here? It's only used in the error message, which should say something like `Failed to retrieve versions for ${location}`

src/webview-contract/webviewDefinitions/clusterProperties.ts
```diff
@@ -1,3 +1,4 @@
+import { KubernetesVersionListResult } from "@azure/arm-containerservice";
```
peterbom, line 1: I'm actually surprised this works... The `webview-contract` files are referenced by both the main extension project and the `webview-ui` project. The `webview-ui` project does not have any Azure SDK libraries as dependencies, so I'd expect the `import` to fail (maybe it works because it's just a type annotation?). Shall we see if we can come up with a plain object-based structure that doesn't refer to any Azure libraries?

peterbom (dismissed):

webview-ui/src/ClusterProperties/ClusterDisplay.tsx
```diff
@@ -62,6 +62,9 @@ export function ClusterDisplay(props: ClusterDisplayProps) {
     const showReconcileButton =
         props.clusterInfo.provisioningState === "Canceled" && props.clusterInfo.powerStateCode === "Running";
 
+    const supportedPatchVersions = props.clusterInfo.supportedVersions.flatMap((v) => v.patchVersions);
```
peterbom, line 65: @sprab, @Tatsinnit: I changed this logic slightly - can you please check it? Previously it was checking whether the version was supported by seeing whether the cluster version was a substring of one of the supported minor versions. This is now a stricter check, which looks at whether the cluster version matches one of the supported patch versions. Is this right?

src/commands/utils/clusters.ts
peterbom: Added a commit to address this

src/webview-contract/webviewDefinitions/clusterProperties.ts
peterbom: Added a commit to address this.

webview-ui/src/ClusterProperties/ClusterDisplay.tsx
Tatsinnit: 💡 Thanks, we just need to think of this might need @sprab , Pete and I might need some ideas/clarifications for this 

image

peterbom: @sprab: A specific question that might help us answer this: what does it mean for the user if the patch version is not in the list of supported versions, but the minor version is?

I agree with Tats that "out of support" is misleading in this case, because the minor version is still in support. But is there something "wrong" with the cluster if the patch version is not in the list?

src/panels/ClusterPropertiesPanel.ts
```diff
@@ -204,17 +211,24 @@ export class ClusterPropertiesDataProvider implements PanelDataProvider<"cluster
...
+        webview.postGetPropertiesResponse(asClusterInfo(cluster.result, kubernetesVersion.result));
     }
 }
 
-function asClusterInfo(cluster: ManagedCluster): ClusterInfo {
+function asClusterInfo(cluster: ManagedCluster, kubernetesVersionList: KubernetesVersionListResult): ClusterInfo {
     return {
         provisioningState: cluster.provisioningState!,
         fqdn: cluster.fqdn!,
         kubernetesVersion: cluster.kubernetesVersion!,
```
peterbom, line 228: I just figured out something: this should be `cluster.currentKubernetesVersion`, not `cluster.kubernetesVersion`.

`kubernetesVersion` may not include the patch version (as is the case for the cluster you were looking at):

```sh
az aks show -n tats-jose-inspektorgadget-demo-2 -g tats-demo | grep -i kubernetesVersion
  "currentKubernetesVersion": "1.27.9",
  "kubernetesVersion": "1.27",
```

webview-ui/src/ClusterProperties/ClusterDisplay.tsx
Tatsinnit: So, th fix for this fixed  the above 🥷 https://github.com/Azure/vscode-aks-tools/pull/538#discussion_r1516995823


Screenshot 2024-03-08 at 3 04 21 PM

peterbom (approved):

PR #551, opened 2024-03-12
Remove dependency on Azure Account extension

peterbom:
Addresses #566 

This removes our dependence on the [deprecated](https://github.com/microsoft/vscode-azure-account?tab=readme-ov-file#azure-account-and-sign-in) Azure Account extension.

The following commands were originally only in the Azure Account extension, and so have been re-implemented in the AKS extension:
- Filter selected subscriptions
- Sign-in to Azure

This required a new configuration setting in the AKS extension: `aks.selectedSubscriptions`. This behaves identically to the equivalent setting in the Azure Account extension (`azure.resourceFilter`). If the Azure Account setting exists, our extension will still use that, until the selected subscriptions for the AKS extension are changed.

The recommended alternative to the Azure Account extension is the [@microsoft/vscode-azext-azureauth](https://www.npmjs.com/package/@microsoft/vscode-azext-azureauth) package. However, this does not provide good support for getting credentials that we can use to create clients for other Azure SDKs. It was necessary to use the VS Code authentication API directly, in addition to the Azure Auth API. This duplication made the session state confusing and potentially inconsistent. To simplify, I removed the dependency on the Azure Authentication package in favour of consuming just the VS Code API (see `azureAuth.ts`).

Until now we have also been using the [@microsoft/vscode-azext-azureutils](https://www.npmjs.com/package/@microsoft/vscode-azext-azureutils) package which provides base class implementations of Account and Subscription nodes for our Cloud Explorer treeview. Unfortunately these classes are very tightly coupled to the Azure Account extension. Since our code base makes heavy use of credentials retrieved from tree nodes (via `ISubscriptionContext`), using these base classes means we are still using the Azure Account extension. So, this PR removes that package dependency too, and now implements the Account and Subscription tree nodes directly (based on classes in the [@microsoft/vscode-azext-utils](https://www.npmjs.com/package/@microsoft/vscode-azext-utils) library, which is not tied to the Azure Account extension). This change means the Account and Subscription icons are now in our own repository, and so the webpack configuration for copying these files has also been removed.

For AAD clusters, we were previously using the low-level [MSAL library](https://www.npmjs.com/package/@azure/msal-node) to get an access token using the cluster's tenant and server scopes. This made use of a refresh token retrieved from the `ISubscriptionContext` credentials. With the authentication changes, this is now no longer available. Besides, the refresh token flow was [not recommended](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/6bd770d16fa11688ad3ea2a42f5ed82c67632553/lib/msal-node/src/client/ClientApplication.ts#L201-L207) anyway. Fortunately, the VS Code authentication API handles this scenario, without us having to resort to lower-level libraries, and so that dependency is also removed.

Other notes:
- The `clusters.ts` contains functions for creating ARM clients for other resource types (not just clusters). I'd like to move these to the new `arm.ts` file so they're in one place, and so it's easier to verify that we're using credentials in a consistent manner. We can do that in a separate PR.
- To avoid the temptation to add more ARM functionality that's not related to clusters into `clusters.ts`, I created `subscriptions.ts`. I think we can follow a similar pattern of having separate files for other ARM resources.
- The `azureAccount.ts` file needs more work. It was previously used to get the Azure Account extension API, plus some miscellaneous ASO functionality (which is now broken on the MS tenant, see #552). As part of fixing the ASO functionality, we should probably delete this file entirely.


src/auth/azureAuth.ts
```diff
@@ -0,0 +1,246 @@
...
+
+        const newSignInStatus = "SigningIn";
+        if (newSignInStatus !== this.signInStatusValue) {
+            this.signInStatusValue = newSignInStatus;
+            this.onSignInStatusChangeEmitter.fire(this.signInStatusValue);
+        }
+        await this.signInAndUpdateTenants({ createIfNone: true, clearSessionPreference: true });
+    }
+
+    private async signInAndUpdateTenants(options: AuthenticationGetSessionOptions): Promise<void> {
```
Tatsinnit, line 76: 💡 This whole experience needs quick test form folks with non msft account as well, I have tried to give a good shake, things look good to me,

I feel this is perfect opportunity for us to improve how first sign-up will happen. If there is any further enhancement is needed we can improve as well go forward.

src/auth/azureAuth.ts
```diff
@@ -0,0 +1,246 @@
...
+    };
+
+    return { succeeded: true, result: tokenInfo };
+}
+
+export function getDefaultScope(endpointUrl: string): string {
+    return endpointUrl.endsWith("/") ? `${endpointUrl}.default` : `${endpointUrl}/.default`;
+}
+
+interface Jwt {
```
Tatsinnit, line 185: 💡Question only: This interface is `jwt` because of `Json Web Token` or is this taken from some implementation et. al.?

src/auth/azureAuth.ts
```diff
@@ -0,0 +1,246 @@
...
+    const tokenInfo: TokenInfo = {
+        token: session.accessToken,
+        expiry: new Date(jwt.result.exp * 1000),
+    };
+
+    return { succeeded: true, result: tokenInfo };
+}
+
+export function getDefaultScope(endpointUrl: string): string {
+    return endpointUrl.endsWith("/") ? `${endpointUrl}.default` : `${endpointUrl}/.default`;
```
Tatsinnit, line 182: 💡 I wonder if we add sample above this for what is the `enpointURL` sample will look like et. al.

src/commands/aksAccount/aksAccount.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    if (allSubscriptions.result.length === 0) {
+        const noSubscriptionsFound =
+            "No subscriptions were found. Setup your account if you have yet to do so or check out our troubleshooting page for common solutions to this problem.";
+        const setupAccount = "Setup Account";
+        const openTroubleshooting = "Open Troubleshooting";
+        const response = await window.showInformationMessage(noSubscriptionsFound, setupAccount, openTroubleshooting);
+        if (response === setupAccount) {
+            env.openExternal(Uri.parse("https://aka.ms/AAeyf8k"));
```
Tatsinnit, line 27: 💡 I wonder what these alias are and worth to comment code the resolved name endpoint here.

Tatsinnit (approved):
Thanks heaps and looks good to me, lets get more eyes for testing and @hsubramanianaks eyes as well as he was keen to monitor these changes as well. + @sprab and @qpetraroia for testing help please ❤️❤️☕️ (Might be testing from non-msft account as well) 

This definitely seems progressive PR.

Screenshot 2024-03-19 at 7 22 36 PM

Screenshot 2024-03-19 at 7 22 44 PM

peterbom:
@hsubramanianaks found a bug where the following happens:
- You sign in using the "AKS: Sign in to Azure..." command
- You select the "Sign in to another account" option (or it's your first time signing in)
- A browser authentication window opens up
- You close that window without completing the sign-in process

In that case VS Code thinks you're still signing in (there's no way it can be notified you've closed the browser tab). And if you try to sign in _again_, it waits for the _original_ sign-in request to complete. This behaviour is part of the VS Code API and beyond our control. It is documented as a [feature request](https://github.com/microsoft/vscode/issues/113270) in the VS Code repo, with the request being to support cancellation. The discussion suggests the request will eventually time out, but that did not happen for me after several minutes and I got tired of waiting. The issue has been in its current state for 3 years, so it doesn't look as if it'll see any movement anytime soon.

If it were to be fixed (i.e. the API changed to support cancellation) we could present a popup similar to the Azure Account one:
image

As it is, I'm reluctant to make further changes at this stage, for what I hope is a relatively infrequent scenario.

One thing we _could_ do is display a cancellation dialog (like the one above), and if the user cancels we invoke the 'reload window' command. It's a bit of a sledgehammer approach, but it's currently the only way I can see for the user to get out of the trap. I think it'd make sense to try something like this in a separate PR, so it's easier to roll it back and replace with a better solution if we think of something.

Any thoughts, @hsubramanianaks and @Tatsinnit?

src/auth/azureAuth.ts
peterbom: Good call. For YAGNI reasons I'm removing all the fields except `exp`, because they aren't needed.

Background info: the interface exists because the access token is a JWT (Json Web Token), and these are (mostly) standard `jwt` properties. The whole `getTokenInfo` function which uses this interface is needed for storing cached tokens for AAD clusters, and these require an expiry time (otherwise running `kubectl` will launch an interactive login window immediately).

However, the VS Code authentication API does not expose token expiry (the access token is opaque), so we use this interface to parse and extract that. The properties exist only for typing purposes (they don't affect the executing JavaScript), and so the presence of extra properties which might not exist would not break it unless we tried to use them. But still, better to remove them.

src/commands/aksAccount/aksAccount.ts
peterbom: Great catch! That resolves to https://github.com/microsoft/vscode-azure-account/wiki/Troubleshooting#setup-your-azure-account which is specifically for the Azure Account extension that we're removing.

There is no general-purpose troubleshooting page for authentication in our extension, so I'm removing that option. For setting up an account I'm changing the URL to "https://azure.microsoft.com/" (which is what the Azure Account documentation links to).

src/auth/azureAuth.ts
peterbom: Agree - added.

Tatsinnit (dismissed):
Looks good to me, more unbiased test will be if @hsubramanianaks can please share his testing or anyone else who kindly help for testing, thanks heaps for hits.

peterbom:
Testing checklist:

- Uninstall Azure Account extension
- Edit user settings (JSON)
  - Make a note of any subscriptions listed under "azure.resourceFilter"
  - If settings contains the key "aks.selectedSubscriptions", delete it and reload window
- Run command: "AKS: Sign in to Azure..." and sign in to corp account
- Expand cloud explorer treeview
  - Check listed subscriptions match those under "azure.resourceFilter"
- Run command: "AKS: Select subscriptions..."
  - Select a different set of subscriptions
  - Check the "aks.selectedSubscriptions" setting is updated
  - Check the treeview refreshes
- Run command: "AKS: Sign in to Azure..." and sign in to personal account
  - Check treeview refreshes
  - Check all subscriptions appear (it shouldn't be necessary to set filters)
- Right-click on a cluster in a non-MSIT tenant and run "Install Azure Service Operator"
  - Enter a service principal clientId and secret and click "check"
  - Verify this runs without error
  - NOTE: this fails on the MSIT tenant for unrelated reasons (we have an issue for this)
- Sign back into MSIT tenant ("AKS: Sign in to Azure...")
  - This time there should be only one confirmation dialog
  - Cloud explorer treeview should refresh
- Repeat switching between tenants to make sure it keeps working smoothly
- Check the Account icon in bottom-left of screen
  - Verify the treeview matches one of the signed-in accounts
- Create a cluster using VS Code (in personal and/or MSIT tenant)
  - Check it completes successfully
  - After it's created, search for Action Group "RecommendedAlertRules-AG-11" in the Azure Portal
  - Check the email address in the Overview section of the portal (it should be the correct email address for the signed-in account)
- Show Properties
- Compare AKS clusters
- Delete a cluster
- Show in Azure Portal
- Reconcile a cluster
- Rotate cluster certificate
- Run AKS Periscope

qpetraroia (approved):

Tatsinnit (approved):
🎊 Thank you for this, given the size of this, I would highly recommend we are happy with testing path, I did tested previously (2/3 weeks back on my Mac Darwin, and it was looking good)

I will just fyi : @qpetraroia + @hsubramanianaks  if they have something to add. Thank you so much everyone. ❤️

Tatsinnit:
> @hsubramanianaks found a bug where the following happens:
> 
> * You sign in using the "AKS: Sign in to Azure..." command
> * You select the "Sign in to another account" option (or it's your first time signing in)
> * A browser authentication window opens up
> * You close that window without completing the sign-in process
> 
> In that case VS Code thinks you're still signing in (there's no way it can be notified you've closed the browser tab). And if you try to sign in _again_, it waits for the _original_ sign-in request to complete. This behaviour is part of the VS Code API and beyond our control. It is documented as a [feature request](https://github.com/microsoft/vscode/issues/113270) in the VS Code repo, with the request being to support cancellation. The discussion suggests the request will eventually time out, but that did not happen for me after several minutes and I got tired of waiting. The issue has been in its current state for 3 years, so it doesn't look as if it'll see any movement anytime soon.
> 
> If it were to be fixed (i.e. the API changed to support cancellation) we could present a popup similar to the Azure Account one: image
> 
> As it is, I'm reluctant to make further changes at this stage, for what I hope is a relatively infrequent scenario.
> 
> One thing we _could_ do is display a cancellation dialog (like the one above), and if the user cancels we invoke the 'reload window' command. It's a bit of a sledgehammer approach, but it's currently the only way I can see for the user to get out of the trap. I think it'd make sense to try something like this in a separate PR, so it's easier to roll it back and replace with a better solution if we think of something.
> 
> Any thoughts, @hsubramanianaks and @Tatsinnit?

Thanks for this discussion, this is great discussion. **Here is the moving forward solution**, we just open this tracking work-item and this could be linked to any other issue which links to the genesis of this login behaviour which Hari kindly found.

**Reason being:** For any new user finding this behaviour we will have clarity as to why the behaviour exist and if there is any work around for it? What do you all think?

* 💡 It is crucial that we document this, and document against this repo and since I got few kind pings from @qpetraroia who is also testing this, we get @qpetraroia fully understood about this behaviour in case of further fyi which is in need for any further internal or external comms.

* **Suggestions:** It will be worth to open preferably a tracking feature-request against this repo which could be one place to summarise the causal effect of any broken behaviour, but also kind of gist the solution and moving forward plan and link the other external issues which are causing this.

* **Observation:** This [issue mentioned in the link](https://github.com/microsoft/vscode/issues/113270) is created `Dec 22, 2020` it will be great to follow this up and find out if this is at all going to get fixed or this behaviour will stay, because then latter stage we need to draw some plans and engage with `vscode` folks we might have access to, (**Not** saying all need to be done right-now) basically this is archiving this thought as moving forward thoughts for when we need to refer back.  

* One key point I think what Hari is eluding to, is that if this is not getting fixed by `vscode` then we could scope something for hacking solution kind against this repo or if there is any temporary scaffold which can avoid the behaviour? (Open question here) Please do correct me if I am incorrectly understanding @hsubramanianaks 

Thank you so much as always. ❤️🙏 fyi this will **unblock this PR** but adds clarity to the future discussion or solution we might need around this and cc: @hsubramanianaks, @peterbom , @qpetraroia , @sprab + cc: @gambtho

Tatsinnit (approved):

Tatsinnit (dismissed):

alexweininger:
Hi Peter and team 👋, I stumbled on this PR from an email thread I was recently included on. This is a lot of work, thank you for being so proactive on moving off of the Azure Account extension. It's not technically deprecated yet, but we are working on deprecating it as we speak.

I'm the owner of the new [@microsoft/vscode-azext-azureauth](https://www.npmjs.com/package/@microsoft/vscode-azext-azureauth) npm package. I understand you didn't use it because it was missing some features. I know I'm very late, but I'm totally willing to add things to this package if it means we can share auth implementations.

From what I see, we'd need to add:
* Easy way to get credentials for the SDK
* Tenant selection

Those are things we plan to add anyway.

Anyway, please let me know if I can help in any way regardless of if you plan to use our package or not.

hsubramanianaks (commented):
Given @alexweininger has replied stating ```@microsoft/vscode-azext-azureauth``` this extension will support the features like get credentials and selecting tenants etc. do you think we can hold on to this PR and not having multiple changes related to this feature (one change now and another one latter when extension supports these features) what are your thoughts? @peterbom please let me know.

alexweininger:
If you want to pursue using the package, my suggestion is that we meet to go over your requirements in more detail to see if it's the right fit.

peterbom:
@alexweininger, @Tatsinnit, @hsubramanianaks - thank you all for the comments.

I'd really like to work on a shared implementation for the authentication library, which I'm keen to devote some careful thought to and work with @alexweininger on building. I think this will take some time and should not be rushed.

The core changes here are centralised in a couple of files, with the implementations abstracted away behind interfaces. I think that gives us a path forward:

1. We continue with this PR as is (subject to any further comments on the implementation as it stands).
2. We work on building any missing functionality/features into the [wrapper library](https://github.com/microsoft/vscode-azuretools/tree/main/auth)
3. We swap out our implementation for the shared one as soon as we're confident it supports all our requirements.

There is no need for these later changes to have any impact on UX. The wrapper code uses the same underlying VS Code API, so user sessions should remain intact if we swap out implementations. The wrapper also supports custom subscription filters so we can continue to use the same user settings introduced in this PR.

qpetraroia:
+1 to @peterbom's vision here and thank you @alexweininger for being open to meeting us and discussing our requirements!

Tatsinnit (dismissed):

qpetraroia:
Signing off. I have opened a bug on our side to alert customers of the current issues with Azure login and the appropriate steps on how to fix it #639.

qpetraroia (approved):

Tatsinnit (approved):

qpetraroia (approved):

PR #567, opened 2024-03-19
Replace download with more in-house implementation.

Tatsinnit:
Nice @peterbom! We need this because we have a package in use called `download` which is not updated for last 4 years and this in-house implementation saves us any cross package dependency and future security issues.

Thanks heaps, @peterbom ❤️❤️

Screenshot 2024-03-19 at 7 28 40 PM


Tatsinnit:
☕️ +1 for this implementation, @peterbom please approve it and we can merge it

peterbom (approved):
There is still an outstanding work item for this, #568, but we can address this later.

Issue #569, opened 2024-03-19
Feature Request - Integrate `Retina.sh` ~> `Capture` Feature in AKS VS Code Extension.


Tatsinnit:
### Proposal for a feature to integration Retina.sh - Capture command for Network Packets. - https://retina.sh 

Enable Network Observability, AKS Dev Experience, as part of this here is proposal enabling Retina.sh Capture feature vis VsCode Extension.

Retina capture command allows the user to capture network traffic and metadata for the capture target, and then send the capture file to the location by Output Configuration.

To conceptualise this aspects specifically for this proposal.

<img width="898" alt="Screenshot 2024-03-20 at 10 57 51 AM" src="https://github.com/Azure/vscode-aks-tools/assets/6233295/0ba094e8-edd1-4cff-84f3-74d7e303ab36">

### How/ What / Where:

Given we have now few network diagnostic / user experience / dev-ops / runoffs experience, we should bundle this into dedicated men-item with submenu options of:

- Collect TCP-Dump (Linux support) (Already present in vscode)
- Deploy/Undeploy and Use Inspector Gadget Commands like Trace, profile et. al. (ebpf goodness) (Already present in vscode)
- Retina.sh -> Distributed Network Capture (_Intent for this proposal_)

## Capture Details:

Retina Capture allows users to capture network traffic/metadata for the specified Nodes/Pods.

Captures are on-demand and can be output to the host filesystem, a storage blob, etc.

More detail here: https://retina.sh/docs/captures/

### Steps to install:

Vscode first will download latest release - we can leverage similar mechiains,m we do for Inspector Gadget tool from this: https://retina.sh/docs/installation/cli 

Option 1 aka download binaries from there: https://github.com/microsoft/retina/releases/tag/v0.0.1

Open Question: Currently I only see the Linus and [kubectl-retina-linux-amd64](https://github.com/microsoft/retina/releases/download/v0.0.1/kubectl-retina-linux-amd64)  - so is this one for all operating system? or other OS will come as time will go?

* https://retina.sh/docs/captures/cli - Once we have the gadget we can then run it like `kubectl retina .. command` 

### Scenarios It will benefit:

// To Do...

Thanks.

peterbom:
I'm keen to aim towards (not immediately, but as a future ideal) an experience where the user flow goes something like:
- **scenario** (e.g. networking problem)
  - **action** (e.g. packet capture)
    - **tool** for performing that action (e.g. tcpdump or retina capture).

[...]

Issue #260, opened 2023-10-10
Network Connectivity - DevOps Experience - what / how.


Tatsinnit:
https://learn.microsoft.com/en-us/azure/architecture/operator-guides/aks/troubleshoot-network-aks

PR #624, opened 2024-04-16
Day 0 - hook for Retina and Azure N/W Distributed Capture Scenarios with AKS VsCode Extension.

Tatsinnit:
### Description and Goal of this PR:

Thank you so much for all the cross collaboration for this and the quick loop back to fix various issues along the way across Retina et. al. Below are the Goal and Non-Goals and future representation of what we plan to achieve. Special mention to @gamntho and 

**Goal:** This PR is solely focused in on enabling Retina Tool Distributed Capture Enablement which intern detail below enables various moving pieces under `one-click` umbrella, covering specific user-scenarios of https://github.com/Azure/vscode-aks-tools/issues/569#issuecomment-2043735238 , proving ease to user and collecting log via Retina which helps for User Experiences. This PR is solely focus on short-term goal discussed under this feature request #569 which is closing the loop on #260 

**Non-Goal:** This PR is **not** the long term U/X based user-scenario work which we will do as part of internal PRD under which we need to identify the re-use or removing certain exisiting features incrementally. If we need to rethink U/X from various stand point of view along with possible reachout to vscode team to get their guidance if need be.

This PR enables a key aspect of Network loop which we wanted to close before we solely focus on upgrading and maintaining those 

### What we achieve:

This PR bundles few things into one: 

Under `one-click`:

User will navigate to their cluster and right-click select `network troubleshooting menu` submenu `Run Retina Capture`

* VsCode will get Binary from latest version release into the vscode extensions binary mechanism to use.
* Once Binary is there, user will get multi-select quick pick for selecting the node we can run Retina Distributed capture scenario on.
* Upon selection node and clicking `OK` the vscode heavy-lifting at the background will start, it will wait for https://retina.sh/docs/captures/cli#capture-targetrequired `capture network packets on the node selected by node names` which will kick in.
* Then, user get presented with successful and unsuccessful run.

Get the Retina logs to Host Machine:

* For this we have provided a button to Download Capture Retina Logs - once user will select that the pod gets run to download the log into those machine.
* User gets presented with option of where to store these logs. 
* Once Downloaded - we provide open folder link to directly open the downloaded folder.

Hence doing all these things with ease, this is not perfect and we will incrementally improve but providing this baseline really will enable users to cover the Azure N/W Log captures via vscode.

Screenshot below is a quick recap:

### Collaboration:

* I guess this could not be done without constant positive support from Neha Aggarwal, @rbtr ,  @vakalapa, @gambtho , @hsubramanianaks , @sprab , Praveen, Qi and Chandan
* Amongst various findings in smaller focused mindset we also collaborated in ganging our understanding for retina tool and helped raising few improvements like https://github.com/microsoft/retina/issues/100 or https://github.com/microsoft/retina/issues/258 Kudos to Evan, Prabha, Vamsi for this collaboration. Thank you so much to Hari for helping the U/X side of things.

### Future:

* We could continue improve iteratively for this, but at least with mindset of getting baseline feature, we intend to close gaps for dedicated AKS level tool which could ease and help user to get these logs for the user.
* We will do a long term improvement PRD which will also could spend enough time understanding maturing these features and iterative growth.

### VSIX to test:

* Latest vsix: [vscode-aks-tools-1.4.1-test-alpha.vsix.zip](https://github.com/Azure/vscode-aks-tools/files/15046379/vscode-aks-tools-1.4.1-test-alpha.vsix.zip)

### Screenshots:

Screenshot 2024-04-16 at 6 18 30 PM

Screenshot 2024-04-16 at 6 20 35 PM

Screenshot 2024-04-16 at 6 23 05 PM

Screenshot 2024-04-16 at 6 23 20 PM

Screenshot 2024-04-16 at 6 23 32 PM

Screenshot 2024-04-16 at 6 23 43 PM

Screenshot 2024-04-16 at 6 23 53 PM

Screenshot 2024-04-16 at 6 24 07 PM

https://github.com/Azure/vscode-aks-tools/assets/6233295/0a6f0e39-6ee9-4339-a3e5-785048cc4e6e


.vscode/settings.json
```diff
@@ -10,5 +10,8 @@
     "eslint.workingDirectories": [".", "./webview-ui"],
     "editor.defaultFormatter": "esbenp.prettier-vscode",
     "editor.formatOnSave": true,
-    "cloudcode.azureExplorerVisible": true
+    "cloudcode.azureExplorerVisible": true,
+    "[typescriptreact]": {
```
hsubramanianaks, line 14: do we really need this ? it was accidental commit?

src/panels/RetinaCapturePanel.ts
```diff
@@ -0,0 +1,169 @@
...
+        });
+
+        if (!localCaptureUri) {
+            return;
+        }
+
+        const localCpPath = getLocalKubectlCpPath(localCaptureUri);
+
+        const createPodYaml = `
+apiVersion: v1
```
hsubramanianaks, line 81: I think may be for future we can extract this to template file and use it for readability. leaving here as comment just to remember this and if you agree @Tatsinnit

src/panels/RetinaCapturePanel.ts
```diff
@@ -0,0 +1,169 @@
...
+
+        if (failed(waitResult)) {
+            vscode.window.showErrorMessage(`Failed to wait for Pod to be ready: ${waitResult.error}`);
+            return;
+        }
+
+
+        // // kubectl cp 
+        const nodeExplorerResult = await longRunning(`Copy captured data to local host location ${localCpPath}.`, async () => {
+            const cpcommand = `cp node-explorer-${node}:mnt/capture ${localCpPath} --retries 99`;
```
hsubramanianaks, line 131: I am wondering what is this 99 retries? does it retry 99 times?

src/panels/RetinaCapturePanel.ts
```diff
@@ -0,0 +1,169 @@
...
+            vscode.window.showErrorMessage(`Failed to apply copy command: ${nodeExplorerResult.error}`);
+            return;
+        }
+
+        const goToFolder = "Go to Folder";
+        vscode.window.showInformationMessage(`Successfully copied the Retina Capture data to ${localCpPath}`, goToFolder)
+            .then(selection => {
+                if (selection === goToFolder) {
+                    open(localCpPath);
+                    // vscode.commands.executeCommand('vscode.openFolder', vscode.Uri.file(`${localCpPath}`));
```
hsubramanianaks, line 145: we can remove this commented line once open works for all Operation System's like windows, mac and Linux. Leaving this comment as placeholder.

src/panels/RetinaCapturePanel.ts
Tatsinnit: ❤️🙏😊 you will get this answer in the kubectl copy command bug ☕️

https://github.com/kubernetes/kubernetes/issues/60140

rbtr:
@Tatsinnit FYI fixing the `--kubeconfig` flag accidentally broke `retina capture download` because of a conflict with our `-n,--name` flag and the inherited `-n,--namespace` flag. 
Fix in progress https://github.com/microsoft/retina/pull/275, will be `v0.0.7`
Note one behavior change: it will be _required_ to set a `--name` during `retina capture create` which was not previously an option (it was a randstring). This is probably better UX as the user specifies the name at create and then immediately knows what to download and later delete.

src/panels/RetinaCapturePanel.ts
Tatsinnit: Thanks for noticing this, Yea, that's the plan, there is good about of clean is now in order for this repo, so bridging this gap for n/w user exp and align that, doing iteratively more close related work PR I guess.

src/commands/aksRetinaCapture/aksRetinaCapture.ts
```diff
@@ -0,0 +1,133 @@
...
+    const nodesList = nodelistResult.result.stdout.trim().split(" ").map((nodeName) => ({ label: nodeName }));
+
+    // Pick a cluster from group to compare with
+    const nodeNamesSelected = vscode.window.showQuickPick(nodesList, {
+        canPickMany: true,
+        placeHolder: "Please select all the Nodes you want Retina to capture traffic from.",
+        title: "Select Nodes to Capture Traffic From",
+    });
+
+    const selectedNodes = await nodeNamesSelected.then((selectedItems) => {
```
peterbom, line 77: To avoid dealing with `undefined` and an extra conversions between strings and arrays, you could do:
```typescript
const nodeNamesSelected = await ...;
if (!nodeNamesSelected) {
    return;
}

const selectedNodes = nodeNamesSelected.map(item => item.label);
```

src/commands/aksRetinaCapture/aksRetinaCapture.ts
```diff
@@ -0,0 +1,133 @@
...
+            `retina capture create --host-path /mnt/capture --node-selectors "kubernetes.io/os=linux" --node-names "${selectedNodes}" --no-wait=false`,
+        )
+    });
+
+    if (failed(retinaCaptureResult)) {
+        vscode.window.showErrorMessage(`Failed to capture the cluster: ${retinaCaptureResult.error}`);
+        return;
+    }
+
+    if (retinaCaptureResult.result.stderr || retinaCaptureResult.result.code !== 0) {
```
peterbom, line 100: I don't think there's any need to check `code`, `strerr` or `stdout` here, so lines 100-107 could be removed. It's already covered by the `if (failed(retinaCaptureResult))` check.

src/commands/aksRetinaCapture/aksRetinaCapture.ts
```diff
@@ -0,0 +1,133 @@
...
+    const kubectlVersion = await getVersion(kubectl, kubeConfigFile.filePath);
+    if (failed(kubectlVersion)) {
+        vscode.window.showErrorMessage(kubectlVersion.error);
+        return;
+    }
+
+    // The pattern of the folder should reflect the pod names and time of the capture
+    const pat = /retina-capture-\w+/g;
+    const match = retinaCaptureResult.result.stdout.match(pat);
+    const startwith = match ? match[0] : null;
```
peterbom, line 118: Does it make sense to continue here if there's no match? If that happens, `folderName` will contain the literal string `"null"`.

src/commands/aksRetinaCapture/aksRetinaCapture.ts
```diff
@@ -0,0 +1,133 @@
...
+        kubeConfigFile.filePath,
+        kubectlVersion.result,
+        clusterInfo.result.name,
+        retinaCaptureResult.result.stdout,
+        selectedNodes.split(","),
+        `${foldername}`
+    );
+
+    const panel = new RetinaCapturePanel(extension.result.extensionUri);
+    panel.show(dataProvider);
```
peterbom, line 132: Passing in `kubeConfigFile` here will ensure it is cleaned up afterwards:
`panel.show(dataProvider, kubeConfigFile);`

src/commands/utils/config.ts
```diff
@@ -92,6 +92,24 @@ export function getDraftConfig(): Errorable<DraftConfig> {
     return { succeeded: true, result: config };
 }
 
+export function getRetinaConfig(): Errorable<DraftConfig> {
```
peterbom, line 95: The return type of `DraftConfig` is a bit confusing. Maybe it could just return a string, since the object only contains a single string property.

src/commands/utils/kubectl.ts
```diff
@@ -182,7 +182,7 @@ export async function streamKubectlOutput(
                     succeeded: false,
                     error: `Failed to run 'kubectl ${args.join(" ")}': ${getErrorMessage(e)}`,
                 }),
-            complete: () => resolve({ succeeded: true, result: new OutputStream(() => {}, new Observable()) }),
+            complete: () => resolve({ succeeded: true, result: new OutputStream(() => { }, new Observable()) }),
```
peterbom, line 185: automatic code formatter?

src/panels/RetinaCapturePanel.ts
```diff
@@ -0,0 +1,169 @@
...
+            getAllNodesResponse: [],
+        });
+    }
+}
+
+export class RetinaCaptureProvider implements PanelDataProvider<"retinaCapture"> {
+    constructor(
+        readonly kubectl: k8s.APIAvailable<k8s.KubectlV1>,
+        readonly kubeConfigFilePath: string,
+        readonly kubectlVersion: KubectlVersion,
```
peterbom, line 30: `kubectlVersion` doesn't seem to be used

src/panels/RetinaCapturePanel.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    getTitle(): string {
+        return `Retina Distributed Capture on ${this.clusterName}`;
+    }
+
+    getTelemetryDefinition(): TelemetryDefinition<"retinaCapture"> {
+        return {
+            getAllNodes: false,
+            runRetinaCapture: true,
+            retinaCaptureResult: true,
```
peterbom, line 45: Out of these, only one seems to be implemented, so we probably shouldn't be collecting telemetry for either of the other two (which should be removed?)

src/panels/RetinaCapturePanel.ts
```diff
@@ -0,0 +1,169 @@
...
+        const applyResult = await longRunning(`Deploying pod to capture ${node} retina data.`, async () => {
+            return await withOptionalTempFile(createPodYaml, "YAML", async (podSpecFile) => {
+                const command = `apply -f ${podSpecFile}`;
+                return await invokeKubectlCommand(this.kubectl, this.kubeConfigFilePath, command);
+            });
+        });
+
+        if (failed(applyResult)) {
+            vscode.window.showErrorMessage(`Failed to apply Pod: ${applyResult.error}`);
+            throw new Error(`Failed to apply Pod: ${applyResult.error}`);
```
peterbom, line 116: Here we're throwing an error in the event of something failing, but a few lines down we're calling `window.showErrorMessage(...)` and returning. I think the latter is more appropriate.

Another option would be to return results (including errors) to the webview. This would also allow the webview to present a link to the capture file, but that gets into a larger discussion which I guess we can postpone for now.

src/panels/RetinaCapturePanel.ts
peterbom: I think the `--retries` option is only available for `kubectl` versions >= 1.23.0. Are we assuming users will be on that version or later?

src/panels/RetinaCapturePanel.ts
```diff
@@ -0,0 +1,169 @@
...
+import { TelemetryDefinition } from "../webview-contract/webviewTypes";
+import { withOptionalTempFile } from "../commands/utils/tempfile";
+import { KubectlVersion, invokeKubectlCommand } from "../commands/utils/kubectl";
+import * as k8s from "vscode-kubernetes-tools-api";
+import { failed } from "../commands/utils/errorable";
+import * as vscode from "vscode";
+import { longRunning } from "../commands/utils/host";
+import { relative } from "path";
+import { platform } from "os";
+import open from 'open';
```
peterbom, line 14: Do we need to add this as a dependency in `package.json`?

src/panels/RetinaCapturePanel.ts
```diff
@@ -0,0 +1,169 @@
...
+            .then(selection => {
+                if (selection === goToFolder) {
+                    open(localCpPath);
+                    // vscode.commands.executeCommand('vscode.openFolder', vscode.Uri.file(`${localCpPath}`));
+                }
+            });
+    }
+}
+
+function getLocalKubectlCpPath(fileUri: Uri): string {
```
peterbom, line 151: This function is duplicated in TCPDump, so maybe should be extracted somewhere else?

src/panels/TcpDumpPanel.ts
```diff
@@ -76,7 +76,7 @@ export class TcpDumpDataProvider implements PanelDataProvider<"tcpDump"> {
         readonly kubectlVersion: KubectlVersion,
         readonly clusterName: string,
         readonly linuxNodesList: string[],
-    ) {}
+    ) { }
```
peterbom, line 79: Auto-formatting? We should probably standardize on code formatters. It should be using `prettier` for everything.

src/webview-contract/webviewDefinitions/retinaCapture.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface InitialState {
+    selectedNode: string;
+    clusterName: string;
+    retinaOutput: string[];
+    allNodes: string[];
+    captureFolderName: string;
+}
+
+export type ToVsCodeMsgDef = {
+    retinaCaptureResult: string;
```
peterbom, line 12: See comments in `RetinaCapturePanel.ts` - some of these should probably be renamed/removed.

src/webview-contract/webviewDefinitions/retinaCapture.ts
```diff
@@ -0,0 +1,19 @@
...
+    getAllNodes: void;
+    runRetinaCapture: string;
+};
+
+export type ToWebViewMsgDef = { (): void };
```
peterbom, line 17:
```typescript
export type ToWebViewMsgDef = Record<string, never>;
```
Because it's currently not sending any messages to the webview.

src/webview-contract/webviewDefinitions/retinaCapture.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type RetinaCaptureDefinition = WebviewDefinition<InitialState, ToVsCodeMsgDef, ToWebViewMsgDef & { [key: string]: unknown }>;
```
peterbom, line 19: Shouldn't this just be:
```typescript
export type RetinaCaptureDefinition = WebviewDefinition<InitialState, ToVsCodeMsgDef, ToWebViewMsgDef>;
```

webview-ui/src/RetinaCapture/RetinaCapture.module.css
```diff
@@ -0,0 +1,65 @@
+.content {
+    display: grid;
+    grid-template-columns: 6rem 50rem;
+    grid-gap: 1rem;
+    align-items: center;
+}
+
+.filterContent {
```
peterbom, line 8: The vast majority of these classes are unused. As far as I can tell, we only need `.content` and `.label`.

webview-ui/src/RetinaCapture/RetinaCapture.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { faInfoCircle } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import styles from "./RetinaCapture.module.css";
+
+export function RetinaCapture(initialState: InitialState) {
+    const { state } = useStateManagement(stateUpdater, initialState, vscode); //eventHandlers
+
+    function handleNodeExplorerPod(nodeName: string) {
+        vscode.postRunRetinaCapture(nodeName);
+        document.getElementById(nodeName)?.showPopover();
```
peterbom, line 14: The 'React' way of doing this is to use [refs](https://react.dev/learn/manipulating-the-dom-with-refs). This would avoid the need for `id` attributes on the elements.

But actually, I can't see where the popover is being created here. Does this do anything?

webview-ui/src/RetinaCapture/RetinaCapture.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+            <VSCodeDivider style={{ marginBottom: "1rem" }} />
+            <div>
+                <FontAwesomeIcon icon={faInfoCircle} /> Retina capture command allows the user to capture network traffic and metadata for the capture target, and then send the capture file to the location by Output Configuration. More info: <a href="https://retina.sh/docs/captures/cli/#output-configurationrequired">Retina Capture Command</a>
+            </div>
+            <VSCodeDivider style={{ marginBottom: "1rem" }} />
+            <pre>{state.retinaOutput}</pre>
+
+            <VSCodeDivider style={{ marginTop: "1rem" }} />
+            <h3>Retina Disctributed Capture is Successfully Completed for this Cluster</h3>
```
peterbom, line 31: Spelling: 'Disctributed'. Also capitalization?

webview-ui/src/RetinaCapture/RetinaCapture.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+            <div className={styles.content}>
+                <label htmlFor="node-dropdown" className={styles.label}>
```
peterbom, line 34: This looks as if it's a label for an input control called `node-dropdown` but I can't see the corresponding control.

If the label is intended as a kind of heading for the table beneath, maybe an 'h2' or 'h3' element would be more appropriate?

webview-ui/src/RetinaCapture/RetinaCapture.tsx
```diff
@@ -0,0 +1,53 @@
...
+                </label>
+                <table>
```
peterbom, line 36: This looks like a list rather than a table. Maybe use `ul` instead?

webview-ui/src/RetinaCapture/RetinaCapture.tsx
```diff
@@ -0,0 +1,53 @@
...
+                    <tbody>
+                        {state.allNodes.toString().split(",").map((node, index) => (
```
peterbom, line 38: The `.toString().split(",")` is redundant here.

webview-ui/src/TCPDump/TcpDump.tsx
```diff
@@ -140,25 +140,25 @@ export function TcpDump(initialState: InitialState) {
...
-                        )}
-                        {!hasStatus(NodeStatus.DeletingDebugPod) && (
-                            <span slot="start">
-                                <FontAwesomeIcon icon={faTrash} />
-                            </span>
-                        )}
-                        Delete
-                    </VSCodeButton>
-                )}
+                        <VSCodeButton
```
peterbom, line 143: The extra whitespace here doesn't look right to me.

webview-ui/src/manualTest/main.tsx
```diff
@@ -38,6 +38,7 @@ const contentTestScenarios: Record<ContentId, Scenario[]> = {
     kubectl: getKubectlScenarios(),
     aso: getASOScenarios(),
     tcpDump: getTCPDumpScenarios(),
+    retinaCapture: [],
```
peterbom, line 41: I think it's important to create the scenarios here to allow manual testing of the views. Doing this means you can launch the webview in a browser with hot reloading as you make code/styling changes. Without the ability to iterate quickly in this way it becomes very onerous to build more complex webviews.

src/commands/utils/config.ts
peterbom: Actually, I see the types for these config objects are declared all over the place. As a first step towards standardizing them, maybe create a type in this file, like:
```typescript
export type BinaryDownloadConfig = {
    releaseTag: string;
}
```

The return type of this function could then be `Errorable<BinaryDownloadConfig>`.

Later on, we can refactor existing config items, like the Kustomize config as:
```typescript
export type KustomizeConfig = BinaryDownloadConfig & {
    repoOrg: string;
    containerRegistry: string;
    imageVersion: string;
}
```

src/commands/aksRetinaCapture/aksRetinaCapture.ts
```diff
@@ -0,0 +1,126 @@
...
+        return;
+    }
+
+    // Retina Run Capture
+    const capturename = `retina-capture-${clusterInfo.result.name}`;
+    const retinaCaptureResult = await longRunning(`Retina Distributed Capture running for cluster ${clusterInfo.result.name}.`, async () => {
+        return await invokeKubectlCommand(
+            kubectl,
+            kubeConfigFile.filePath,
+            `retina capture create --namespace default --name ${capturename} --host-path /mnt/capture --node-selectors "kubernetes.io/os=linux" --node-names "${selectedNodes}" --no-wait=false`,
```
peterbom, line 94: Is there a plan to support user-selection of the namespace for this? And also duration, or any of the other capture options?

Might be worth specifying duration (this will capture for 1 minute as it stands, won't it?) in the long-running task text, so the user knows how long they have to wait.

src/panels/RetinaCapturePanel.ts
Tatsinnit: Thanks for this question. This comes bundles within one of the packages, so never required to require this.

src/commands/aksRetinaCapture/aksRetinaCapture.ts
Tatsinnit: Thanks for this question, its an interesting there is not hard recommendation to use the `capture` namespaces in some sample I see it does and in few it doesn't so keep day-0 simple command on `default namepsace` currently retina vanilla command run under default and capture, **RE:** duration, Retina has the `--no-wait` flag which is best way to make sure there is retina capture complete 100% before we move forward  which is best way for assurance for rest of stuff to work. ☕️❤️🙏

For anything else we could add future PR as and we enhance this please. If for anything uber missed please feel free to add the PR, thank you for this.

Kind fyi to @hsubramanianaks just in case. (also @sprab if anyone else has any thoughts.)

sprab (dismissed):
This looks good.

sprab (commented):
I have tested this.  Works as expected and looks good.

sprab (approved):
I have tested this, and also looked into the captures.  The download capture works fine and able to save this locally.  The downloaded capture has some crucial information that would help us in reducing the TTM.

src/commands/periscope/models/RetinaDownloadConfig.ts
```diff
@@ -0,0 +1,3 @@
+export interface RetinaDownloadConfig {
```
hsubramanianaks, line 1: we can change this to GithubReleaseConfig, I thought it was specific to retina, but it is both for draft and retina. Just a minor name change please.

hsubramanianaks (approved):
LGTM, tested as well works fine in windows and WSL2 (linux). A minor name change suggestion. Thank you @Tatsinnit.

src/commands/periscope/models/RetinaDownloadConfig.ts
Tatsinnit: Thanks for this, given its gh retina config we can discuss renaming for latter because draft work is the whole other PR which has full drraft hooked.

PR #636, opened 2024-04-23
Fix/build publish workflow

Tatsinnit:
Hiya all, The workflow was broken because the permission fix, we only checkout for build and not for release which lead to the failure like this: https://github.com/Azure/vscode-aks-tools/actions/runs/8793642136/job/24131868803#step:5:16  the issue is `Error: Error: ENOENT: no such file or directory, open 'package.json'`

* This fix is to add the checkout and build code for the publish job as well. here is a **sample run from my fork**, please note because I don't have token it fails for those reasons while again main repo it will not. Fix run is here: https://github.com/Tatsinnit/vscode-aks-tools/actions/runs/8793815797/job/24132285488 

Thank heaps, fyi: @peterbom, @hsubramanianaks , @sprab whoever gets to this first. thanks ❤️🙏☕️

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -70,7 +70,7 @@ jobs:
       - name: Install dependencies
         run: npm run install:all
       # Ensure project builds successfully before creating release
-      - name: Build 
+      - name: Build
         run: npm run webpack
       - name: Create a Release
         id: create_release
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,7 +8,7 @@ jobs:
   JobBuild:
       name: release
       runs-on: ubuntu-latest
-       # Expose step outputs as job outputs
+      # Expose step outputs as job outputs
       outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -59,7 +59,19 @@ jobs:
...
+        uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
+      - name: Use Node.js
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        with:
+          node-version: 20
+      # Run install dependencies
+      - name: Install dependencies
+        run: npm run install:all
+      # Ensure project builds successfully before creating release
+      - name: Build
```
peterbom, line 73: I don't think we should need the 'Build' step here. The failing step which creates the vsix must be running `vsce package`, which runs the `vscode:prepublish` script, which itself runs webpack.

hsubramanianaks (approved):

Issue #680, opened 2024-05-20
Treeview can get stuck displaying 'select tenant...' when there's only one tenant to choose from


peterbom:
**Describe the bug**

The Cloud Explorer treeview can end up displaying the 'select tenant' node (shown below), and when you select it you get a message saying you only have one tenant to choose from.
![image](https://github.com/Azure/vscode-aks-tools/assets/1817696/ace5b772-24f9-4c36-8e00-0af2dcb9847c)
![image](https://github.com/Azure/vscode-aks-tools/assets/1817696/6aa61fec-f04b-48ba-8a88-ec1ae71b1540)

The workaround is for the user to sign out and sign in again, but it is not at all obvious they need to do that.

**To Reproduce**

The circumstances under which this happens are:
- The user was previously signed in to the AKS extension, but hasn't used it for some time.
- The user's account is associated with exactly one Microsoft tenant.
- The user's session is valid for listing tenants, but cannot be silently exchanged for a token _for_ the tenant.

At this stage I'm unsure how to reproduce the last condition or the length of inactivity required to observe this, since token exchange is handled at a lower layer.

**Expected behavior**

If there is only one tenant to choose from, the treeview should automatically invoke any extra interaction required for the user to complete the authentication process.

Additionally, the 'select tenant' command should be made more robust. Rather than displaying an information message saying there's only one tenant, it could also invoke `getAuthSession` to ensure the user has a valid token for the tenant.

**Additional context**

To protect against this kind of bug in the future it would be worth defining more tightly what it means for a session to be "ready" and enforcing that in the [API contract](https://github.com/Azure/vscode-aks-tools/blob/b940df90b4fc6b70002b55c906ef89394d73a236/src/auth/types.ts#L30-L33).

The current design is too ambiguous. By convention we avoid setting the `selectedTenant` property of the session if it's not possible to silently obtain a token for that tenant, but that is not enforced by the contract (consumers can set the `selectedTenant` to a value which violates that convention). I'll create a separate issue for some clean-up work.

peterbom:
Fixed by #681

PR #681, opened 2024-05-20
Fix treeview stuck with only 'select tenant' node

peterbom:
Addresses #680 

This is just to fix the observed behaviour that:
1. The treeview can display a node saying 'select tenant' when in fact there's only one tenant to choose from; and
2. The 'select tenant' command currently doesn't do anything if there is only a single tenant to choose from.

There are some additional changes we can make to the session provider API that should make this kind of defect less likely. I'll address those in a separate issue and PR.

### DIFF
--- a/src/commands/aksAccount/aksAccount.ts
+++ b/src/commands/aksAccount/aksAccount.ts
@@ -17,7 +17,16 @@ export async function selectTenant(): Promise<void> {
     }
 
     if (sessionProvider.availableTenants.length === 1) {
-        window.showInformationMessage(`Only one tenant available (${sessionProvider.availableTenants[0].name}).`);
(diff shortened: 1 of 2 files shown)


Tatsinnit (approved):
👍 looks good to me, very specific to the scenario. I could test it latter tonight. (Due to specific nature I believe this might have got missed from the tests done around this)

peterbom:
> 👍 looks good to me, very specific to the scenario. I could test it latter tonight. (Due to specific nature I believe this might have got missed from the tests done around this)

Thank you. I agree we wouldn't have tested this scenario - I'm actually unable to come up with concrete repro steps, as it involves simply not using the extension for an extended number of days (I don't know how many).

PR #686, opened 2024-05-20
Replace "Create GitHub workflow" commands with equivalent Draft command

peterbom:
Now we have added the commands for running Draft, we have two ways of creating a GitHub workflow:
- The 'Create GitHub workflow' commands in the Cloud Explorer tree view
- The 'Draft: Create a GitHub workflow' command in the Command Palette

To avoid confusion, it would be best to provide a single approach. This PR is to make that change, by adding the relevant Draft commands to the tree view, in place of the existing commands:

**Current**
image

**New**
Screenshot 2024-05-21 100116

Once launched from the tree view, any cluster-related fields will be pre-populated:
image

I'm not _sure_ whether we'd want Draft Deployment to appear there. I added it because it _can_ take the cluster as an input, but it's not really necessary (it's only used for providing a list of namespaces to choose from).


peterbom:
Closing this as we are making these changes in two stages:
1. adding the Draft menu items and marking the starter workflow ones as deprecated
2. removing the starter workflow commands

Issue #682, opened 2024-05-20
Improve interface for retrieving authentication sessions


peterbom:
**Is your feature request related to a problem? Please describe.**

The recent issue #680 highlighted an ambiguity around the [AzureSessionProvider](https://github.com/Azure/vscode-aks-tools/blob/b940df90b4fc6b70002b55c906ef89394d73a236/src/auth/types.ts#L20-L33) interfaces.

Specifically, it is unclear whether the `selectedTenant` property is necessarily accessible without further user interaction. Internally (within the `AzureSessionProviderImpl` class) we try to ensure that it _is_ immediately accessible. That means we can return a `ReadyAzureSessionProvider` with that property populated, and be reasonably sure it can be used immediately, without the user being directed away to a browser authentication tab.

The problem here arises from the fact that the `selectedTenant` property serves two purposes: to represent the user's selected tenant AND to indicate they are fully authenticated.

Additionally, the parameters for `getAuthSession` do not allow for specifying additional/different scopes. These are needed in two scenarios:
1. When using a graph client (or any client for endpoints other than ARM).
2. When specifying an Entra ID application other than the default VS Code one (which doesn't have delegated permissions for some graph-related requests relating to managing applications and service principals).

**Describe the solution you'd like**

I think there are two possible ways to resolve the ambiguity of the `selectedTenant` property:
1. Give up on the idea that 'ready' means 'no further user interaction', and instead interpret it as 'user is signed in and has selected a tenant, but further authentication _may_ be required to access that tenant'.
2. Add an extra property (perhaps on the `Tenant` type) to indicate whether the user is fully signed in to the selected tenant.

The first option is technically simpler and so it makes sense to _try_ that first. If it degrades the user experience somehow, we can use that as justification for the more complex option.

For the `getAuthSession` parameters, I'd like to make a change similar to that discussed [here](https://github.com/microsoft/vscode-azuretools/pull/1722#discussion_r1585793261):
```typescript
export type GetAuthSessionParams {
    applicationClientId?: string;
    scopes?: string[];
}
```

peterbom:
It's [uncertain](https://github.com/microsoft/vscode-azuretools/pull/1722#issuecomment-2098765323) whether a similar `AzureSessionProvider` class will make it into the `azuretools` library.

In case that does go ahead, it might make sense to adhere to the structure proposed there, so that we can swap the implementation at a later date with minimal refactoring.

[...]

PR #695, opened 2024-05-23
AzureSessionProvider changes

peterbom:
Minimal changes for #682:
- add options for getting session
- don't require active token for selected tenant

This helps prevent issues like #680 from arising in the future. It also makes the `AzureSessionProvider` interface more resistant to breakages by adding an `Options` parameter for session retrieval options that might grow in the future (e.g. to enable specifying extra graph API scopes or application client IDs).

Design decisions:
1. We will make no immediate attempt to adhere to the structure proposed [here](https://github.com/microsoft/vscode-azuretools/pull/1722).
2. There is no implicit extra meaning to the `selectedTenant` property that assumes the user would be able to silently obtain a token for that tenant.

~~Opening as a draft since those decisions are debatable.~~

EDIT: I discussed this further with @alexweininger (who authored the [Azure authentication extension](https://github.com/microsoft/vscode-azuretools/tree/main/auth)). Some notes:
- There are likely to be [further changes to the VS Code API](https://github.com/microsoft/vscode/issues/152399#issuecomment-2145828468) to better support the handling of multiple accounts and scopes. We agreed it's not worth making significant changes here until that API is finalized.
- Our `AzureSessionProvider` interface currently manages two separate concerns which should probably be distinct:
  1. Authentication state (by account/tenant)
  2. Tenant (and implicitly account) selection (i.e. what account/tenant the user is currently accessing).
  Longer term it would probably make sense to explicitly separate these out. But again, we should wait for further developments in the VS Code API first.

### DIFF
--- a/src/auth/azureSessionProvider.ts
+++ b/src/auth/azureSessionProvider.ts
@@ -7,7 +7,7 @@ import {
     AuthenticationGetSessionOptions,
     AuthenticationSession,
 } from "vscode";
-import { AzureAuthenticationSession, AzureSessionProvider, SignInStatus, Tenant } from "./types";
+import { AzureAuthenticationSession, AzureSessionProvider, GetAuthSessionOptions, SignInStatus, Tenant } from "./types";
 import { Errorable, bindAsync, getErrorMessage, map as errmap, succeeded } from "../commands/utils/errorable";
 import { getDefaultScope, quickPickTenant } from "./azureAuth";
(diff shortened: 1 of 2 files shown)


alexweininger (approved):

Tatsinnit (approved):
🙏 Thanks, looks good to me.

@qpetraroia to be aware and I guess list you are forming to might be better track the co-relation work. Worth to document for chaining the work.

Thanks heaps all.

PR #696, opened 2024-05-23
Add draft context menu commands and deprecate starter workflows

peterbom:
This adds context menu items for the Draft commands, both in the file system explorer and the cloud explorer tree views:

image
image

It also marks the starter workflow commands as deprecated:
image

### DIFF
--- a/src/panels/draft/DraftDockerfilePanel.ts
+++ b/src/panels/draft/DraftDockerfilePanel.ts
@@ -31,6 +31,7 @@ export class DraftDockerfileDataProvider implements PanelDataProvider<"draftDock
     constructor(
         readonly workspaceFolder: WorkspaceFolder,
         readonly draftBinaryPath: string,
+        readonly initialLocation: string,
     ) {
         this.draftDirectory = path.dirname(draftBinaryPath);
     }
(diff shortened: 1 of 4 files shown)


src/panels/draft/DraftDockerfilePanel.ts
```diff
@@ -46,9 +47,9 @@ export class DraftDockerfileDataProvider implements PanelDataProvider<"draftDock
                 fullPath: this.workspaceFolder.uri.fsPath,
                 pathSeparator: path.sep,
             },
-            location: "",
+            location: this.initialLocation,
```
Tatsinnit, line 50: 💡 minor check: I guess this is safe and this will not lead to any exceptions in case containing folder location is deleted and the `intialLocation` is still set or of type `null`.

Tatsinnit (approved):
☕️💡👍 Thank you so much for adding for this review, looks good to me, just added minor fyi only.

This looks much safer and next we can deprecate. 

Re: Documentation:

* Can we please add change of behaviour documentations around this. 

Thanks.

src/panels/draft/DraftDockerfilePanel.ts
peterbom: `initialLocation` won't be `null`. In theory it could get set to a folder that has been deleted, but protecting against that would be a much larger piece of work.

Some kind of file system monitoring could potentially be done using the [createFileSystemWatcher workspace function](https://code.visualstudio.com/api/references/vscode-api#workspace), but I don't think there's a strong enough argument for doing that, given the effort and complexity involved.

PR #698, opened 2024-05-26
Ensure initial selection values are populated

### LINKED ISSUE #697: Draft: repository and namespace are not always retained between deployment and workflow webviews
**Describe the bug**

When launching 'draft workflow' from the 'draft deployment' webview, several properties are supposed to be automatically populated in the 'draft workflow' webview.

Mostly this works, but there is a specific scenario where it doesn't. If the user enters a 'new' repository or namespace (i.e. enter their own text rather than selecting from an item from the pre-populated lists), those values are not retained.

**To Reproduce**

- Run 'Draft: create a deployment'
- Fill in all values, entering a repository and/or namespace that is not in the dropdown list
- Click 'create'
- In the 'next steps' box, click the 'create a workflow' command link

**Expected behavior**

The repository and namespace fields should be populated with the values from 'create a deployment'. They are not.

**Screenshots**

image
image
image

peterbom:
Fixes #697 by making sure that:
1. initial values are set as selected when the webview is initialized
2. the selected values are included in the drop-down lists


Tatsinnit (approved):
👍☕️

PR #720, opened 2024-06-04
Add webview for connecting ACR to cluster

peterbom:
As #709 violates the PR size checker, this is the first part of two (I hope) smaller PRs to introduce the changes more gradually.

This part is just the webview, and can be run locally using the webview launch task. It looks like this:
image

It does not involve any functional changes to the compiled VS Code extension.


Tatsinnit (dismissed):
💡 Thank you for this PR. I assume this is necessary for the Draft Hook integration in the latest VSCode extension update.

* I'll let @qpetraroia review them. If the decisions made between you and Quentin align U/X then worth to just get eyes, I'm happy to proceed.

* It would be beneficial to document the next steps and any other upcoming changes.

Thank you, as always, to everyone for reviewing and contributing or even looking helping this work. Humble and kind fyi to @hsubramanianaks + @sprab + @gambtho for eyes.

webview-ui/src/ConnectAcrToCluster/ConnectAcrToCluster.tsx
```diff
@@ -0,0 +1,280 @@
...
+        lazyAcrs,
+        lazyClusters,
+        lazyAcrRoleState,
+    } = prepareData(state, updates);
+    useEffect(() => {
+        updates.map((fn) => fn(eventHandlers));
+    });
+
+    function getAcrAuthorizationActionItemProps(): ActionItemProps {
+        const createAction = makeFixAction(faSave, "Authorize", null);
```
hsubramanianaks, line 43: passing action as null here was confusing for me , but then realized we need sub, rg etc.. other properties selected so that you can enabled authorize and deauthorize buttons. I would suggest getting UX validations for these terms from Quentin or Ivelisse. But can you please add a comment why action was selected later for other understanding. May be is it possible to simple this method ?

```
createAction.canPerformAction = !isDone;
        createAction.action = () => handleCreateAcrRoleAssignment(acr, cluster);
        deleteAction.canPerformAction = isDone;
        deleteAction.action = () => handleDeleteAcrRoleAssignment(acr, cluster);

```

webview-ui/src/ConnectAcrToCluster/ConnectAcrToCluster.tsx
```diff
@@ -0,0 +1,280 @@
...
+        lazyAcrRoleState,
+    };
+}
+
+function makeFixAction(icon: IconDefinition, name: string, action: (() => void) | null): FixAction {
+    return {
+        icon,
+        name,
+        action: action ? action : () => {},
+        canPerformAction: action !== null,
```
hsubramanianaks, line 220: instead of going by action !== null, can we pass in hasAcrPull boolean variable so that buttons can be enabled? so that we can avoid passing null values.

webview-ui/src/ConnectAcrToCluster/ConnectAcrToCluster.tsx
```diff
@@ -0,0 +1,280 @@
...
+import { ResourceSelector } from "../components/ResourceSelector";
+import { VSCodeButton } from "@vscode/webview-ui-toolkit/react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { IconDefinition, faCheckCircle, faClock, faSave, faTrash } from "@fortawesome/free-solid-svg-icons";
+import { AcrRoleState } from "./state/stateTypes";
+
+export function ConnectAcrToCluster(initialState: InitialState) {
+    const { state, eventHandlers } = useStateManagement(stateUpdater, initialState, vscode);
+
+    const updates: EventHandlerFunc[] = [];
```
hsubramanianaks, line 29: While names like updates, lazeSubscriptions are not bad, they could be more descriptive. Consider using names that clearly describe what the variable is or does.

hsubramanianaks (commented):
@peterbom Thanks for splitting up the PR, changes looks good overall. I would suggest getting feedback on the UX from Quentin/Ivelisse. Also some minor comments for others to understand. Thank you again.

src/webview-contract/webviewDefinitions/connectAcrToCluster.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export type ClusterKey = SubscriptionKey & {
+    resourceGroup: string;
+    clusterName: string;
+};
+
+export type Cluster = ClusterKey; // Fully-defined by its key
+
+// https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles
+export const acrPullRoleDefinitionName = "7f951dda-4ed3-4680-a7ca-43fe172d538d";
```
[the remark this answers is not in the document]
peterbom: I'm adding `// Role definition names are always guids. Here we want the role definition name for AcrPull.`

Adding this because even though I feel like the variable name is already pretty explanatory, I don't feel _very_ strongly either way.

webview-ui/src/ConnectAcrToCluster/ConnectAcrToCluster.tsx
peterbom: Commented.

@qpetraroia: please share if you have some better names here. I went with 'Authorize'/'Deauthorize' because that's what 'connect' really means: you add role assignments to _authorize_ the cluster identity to pull images from the ACR.

My main intent is to get something working that can be iterated on. This has been the spirit of all work so far, rather than aiming for 'perfect'. With all work, we have to make a lot of best guesses, and those guesses won't always be right.

webview-ui/src/ConnectAcrToCluster/ConnectAcrToCluster.tsx
peterbom: Unfortunately, no. There will be times when we don't know the value of `hasAcrPull`. The `action` parameter is intentionally nullable to indicate _either_ the action to perform _or_ `null` meaning action not applicable.

webview-ui/src/ConnectAcrToCluster/ConnectAcrToCluster.tsx
peterbom: Happy to discuss suggestions, otherwise I'll go with "not bad".

`lazySubscriptions` is a collection of subscriptions that are lazily-loaded. I.e. at any given point in time they might be loaded, or they might not. A more descriptive term might be `lazilyLoadedSubscriptions`. But that's a lot more typing and also inconsistent (since the type is called `Lazy` and I'd prefer not to rename the type and all usages of it).

Personally, I think `lazy` is adequate, and unlikely to cause confusion if you have some familiarity with the codebase.

webview-ui/src/ConnectAcrToCluster/ConnectAcrToCluster.tsx
hsubramanianaks: hasAcrPull is boolean variable true | false , passing null is not good practice. canPerformAction: hasAcrpull === true? something like that. this is kind suggestion. if this doesn't work out you have already added comments. it should be good to go. Thank you.

hsubramanianaks (approved):

PR #723, opened 2024-06-06
Add direct node-fetch dependency

### LINKED ISSUE #568: Add direct node-fetch dependency
**Is your feature request related to a problem? Please describe.**
We don't have a direct dependency on `node-fetch`.

This will become a problem if/when we remove the `cross-fetch` dependency (which is likely to happen during the fix for #552), which indirectly ensures we have the `node-fetch` library needed for downloads.

Ideally we would have added a direct reference to `node-fetch` as part of #567, but doing this caused a patch version bump to several `@microsoft/applicationinsights` packages that are transitive dependencies of `@vscode/extension-telemetry`, which unearthed a [bug](https://github.com/microsoft/ApplicationInsights-JS/issues/2307) that causes the extension to fail to load.

When the above bug is fixed, we can remove the `@types/node-fetch` dependency and add `node-fetch` directly.

peterbom:
Fixes #568 

Previously I was unable to add a direct dependency to `node-fetch` (which we use for downloading binaries) because it caused some transitive dependency updates that broke the extension.

These are now fixed, and the dependency can be added without any problems. Because it ships with TypeScript types, the `@types/node-fetch` dependency could also be removed.


hsubramanianaks (approved):
lgtm

Issue #552, opened 2024-03-12
Install ASO command fails on some tenants


peterbom:
**Describe the bug**
ASO requires a Service Principal (actually it no longer does, but it still supports deployment using a Service Principal). The Service Principal belongs to an Entra ID tenant. This tenant may have [Conditional Access](https://learn.microsoft.com/en-us/entra/identity/conditional-access/overview) policies configured which prevent "token issuance". That is, they prohibit authentication flows in which a client ID and secret are exchanged for an access token (at least from certain machines - it still seems to work from within AKS clusters, which is why ASO itself is not broken by this).

In these cases (the MSIT tenant is an example), the ASO deployment command fails at the stage of checking the service principal. That is because we attempt to check subscription access using the Service Principal's own credentials (which are a client ID and secret), rather than those of the signed-in user.

This means we probably have to remove the functionality to check which subscriptions the service principal can access. We can use the signed-in user's token to examine role assignments, but it is very possible the user will not have the required permissions to do this, even if the service principal does. We don't want this command to fail in those cases, and so it will be simpler and more robust to remove the check altogether.

**Mention the platform you are using**
Occurs on all platforms

**To Reproduce**
Run the ASO command in any cluster on the MSIT tenant, and enter a service principal's client ID and secret.

**Expected behavior**
The extension checks which subscriptions the service principal can access without error.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

peterbom:
Fixed by #724

Issue #253, opened 2023-10-05
Link Azure account and GitHub Account


qpetraroia:

peterbom:
I can see a couple of approaches for authenticating the Azure user account with GitHub: via the DevHub API, or directly through VS Code.

### DevHub API

To consume the DevHub API, we add a dependency to [@azure/arm-devhub](https://www.npmjs.com/package/@azure/arm-devhub/v/1.0.0-beta.3), which gives us a [JS Client](https://learn.microsoft.com/en-us/javascript/api/@azure/arm-devhub/developerhubserviceclient?view=azure-node-preview) that we can use to call `gitHubOAuth`. If the user has previously authorized the 'AKS Developer Hub' oauth application to access their GitHub account (see screenshot), this will return a token allowing the user to interact with the GitHub API as this application.

![image](https://github.com/Azure/vscode-aks-tools/assets/1817696/0bdb998b-7794-457c-960f-4349c8f7dfaf)

However, if the user is attempting to authenticate for the first time, `gitHubOAuth` does not return a token. It just returns a URL that can be used to initiate an oauth flow in which the user can grant the AKS Developer Hub application permission to access their GitHub account. It's not clear to me if/how we would handle the callbacks required to complete this flow. It _might_ be achievable by building a [custom authentication provider](https://www.eliostruyf.com/create-authentication-provider-visual-studio-code/).

### Through VS Code

The VS Code API has a built-in [`getSession` function](https://code.visualstudio.com/api/references/vscode-api) which can be used to achieve almost the same result. It retrieves a token for an oauth application (called 'GitHub for VS Code', see screenshot) that has been registered with GitHub, and can also initiate an interactive flow to allow the user to authorize the application in the first place.

![image](https://github.com/Azure/vscode-aks-tools/assets/1817696/28d25061-25bf-4d1a-a2ef-403ec8b3c285)

### Which approach to use?

Consuming the DevHub API would maximise code reuse since we'd be making use of an existing RP, and might even mean we wouldn't depend on the Draft binary. It would also mean that authorization performed in the Portal would also take effect in VS Code, and vice versa.

However, that would make us entirely dependent on the RP implementation and would make it harder to innovate independently. If we were simply building "Automated Deployments for VS Code", this is probably what we'd want, but we are really building our own experience, which we don't want to be tied to a specific Portal feature.

So, in terms of both development effort and agility I think it'd be preferable to use the built-in VS Code API for GitHub authorization.

PR #724, opened 2024-06-07
Avoid querying graph APIs for ASO

peterbom:
Addresses #552.

Ideally, we would want to query graph APIs to get the service principal associated with the specified application client ID.

We cannot do that at the moment, because the client application used to obtain an access token does not have the appropriate delegated permissions. We can add this in the future, but it would require defining our own first party application (which would also be required for #253).

Until then we need to remove the service principal retrieval and role assignment checking. The onus will now be on the user to specify the correct application client ID and secret, and ensure the service principal has access to the selected subscription. This is how the feature was originally designed, so it's not an excessively severe limitation.


Tatsinnit (approved):
Thank you so much for this, it will be nice to document this for the knowledge and loop back of @qpetraroia as well. 

I have nothing much to add, given this is specific to the fix, let's get it in. I have trust that we have tested this. Thanks heaps.

Issue #799, opened 2024-07-16
Support more information for Kubernetes version inside the cluster properties page.


Tatsinnit:
**Is your feature request related to a problem? Please describe.**

Lets explore if we can clearly add what version of Kubernetes will be supported LTS (Long term)

* https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli
* https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/list-kubernetes-versions?view=rest-aks-2024-02-01&tabs=HTTP
* https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/list-kubernetes-versions?view=rest-aks-2024-02-01&tabs=HTTP#kubernetessupportplan 

**Mention what platform you want to support the new feature**

<img width="1008" alt="Screenshot 2024-07-17 at 10 44 12 AM" src="https://github.com/user-attachments/assets/c8ef4629-ce1b-466c-a10c-3eb355a8a164">

**Describe the solution you'd like**

Solution space will reside in this nested property of `listKubernetesVersion` 

*  https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/list-kubernetes-versions?view=rest-aks-2024-02-01&tabs=HTTP#kubernetessupportplan 

Please do feel free to ping me back in case anyone needs more support. I can squeeze this in soon if no one wants. Thank you so much. ❤️☕️🙏

PR #800, opened 2024-07-17
Add Support Plan and Preview information in cluster Properties Page

hsubramanianaks:
This PR for https://github.com/Azure/vscode-aks-tools/issues/799

image

### DIFF
--- a/src/panels/ClusterPropertiesPanel.ts
+++ b/src/panels/ClusterPropertiesPanel.ts
@@ -260,5 +260,6 @@ function asKubernetesVersionInfo(version: KubernetesVersion): KubernetesVersionI
     return {
         version: version.version || "",
         patchVersions: version.patchVersions ? Object.keys(version.patchVersions) : [],
+        capabilities: version.capabilities ? Object.keys(version.capabilities) : [],
     };
 }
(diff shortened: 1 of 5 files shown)


src/panels/ClusterPropertiesPanel.ts
```diff
@@ -260,5 +260,7 @@ function asKubernetesVersionInfo(version: KubernetesVersion): KubernetesVersionI
     return {
         version: version.version || "",
         patchVersions: version.patchVersions ? Object.keys(version.patchVersions) : [],
+        capabilities: version.capabilities ? Object.values(version.capabilities?.supportPlan || {}) : [],
```
ReinierCC, line 263: Would the "?" in version.capabilities?.supportPlan be redundant?

ReinierCC (approved):
New object fields and version display look good.

src/panels/ClusterPropertiesPanel.ts
Tatsinnit: 🛡️ 💡  Thank you @ReinierCC for this comment, I see what you mean, yes indeed it does, because `capabilities` is **already** `nullable` checked,  I think inside if we really want to check it will be `supportPlan`. Likely a slipped the `?` in wrong place.

```*.ts
capabilities: version.capabilities ? Object.values(version.capabilities.supportPlan? || {}) : [],
```

* 🫶  Just minor thing this brings is the if we are returning the `supportPlan` then it will good to change the object name from `capabilities` to `supportPlan`.

qpetraroia (approved):

Issue #786, opened 2024-07-10
Draft a deployment screen is frozen


hsubramanianaks:
**Describe the bug**
Draft a deployment screen is frozen

**Mention the platform you are using**
vscode - ubuntu 22.04

**To Reproduce**
1. Draft a deployment
2. choose the Location* Under Deployment Details as current location ex: ./
3. Create
4. Screen is frozen forever

**Expected behavior**
Deployment files should be created.

**Screenshots**
![image](https://github.com/Azure/vscode-aks-tools/assets/105889062/6b879584-6530-4d15-bb30-f94e39f02865)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #829, opened 2024-07-31
Adding flag to prevent draft from checking for manifests

ReinierCC:
Added flag "--skip-file-detection" 

The draft screen was freezing because it detected manifest files in the directory specified or child directories. By adding the "--skip-file-detection" to the draft command, we can remove the check on the draft side and rely on the extension to inform the user that manifest files are present.

Draft deployment screen was freezing as seen  [https://github.com/Azure/vscode-aks-tools/issues/786](here)


hsubramanianaks (approved):
lgtm, thanks for fixing this @ReinierCC

PR #832, opened 2024-08-02
Updated github auth scope to repo

ReinierCC:
Previous github authentication token only provided access to publicly available repos., which prevented the extension from populating any private repos the user might have had.

### DIFF
--- a/src/commands/draft/draftCommands.ts
+++ b/src/commands/draft/draftCommands.ts
@@ -215,6 +215,7 @@ type DraftDependencies = {
 async function getGitHubAuthenticationSession(): Promise<Errorable<AuthenticationSession>> {
     try {
         // Repo scope required to see public/private repos.
+        // Reference for Github scopes: https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps
         const scopes: string[] = ["repo"];
         const session = await authentication.getSession("github", scopes, { createIfNone: true });
         return { succeeded: true, result: session };
--- a/src/commands/draft/draftCommands.ts
+++ b/src/commands/draft/draftCommands.ts
@@ -214,8 +214,8 @@ type DraftDependencies = {
 
 async function getGitHubAuthenticationSession(): Promise<Errorable<AuthenticationSession>> {
     try {
-        // No special scopes are required for GitHub - we are just listing repositories/branches.
-        const scopes: string[] = [];
+        // Repo scope required to see public/private repos.
+        const scopes: string[] = ["repo"];
         const session = await authentication.getSession("github", scopes, { createIfNone: true });
         return { succeeded: true, result: session };
     } catch (e) {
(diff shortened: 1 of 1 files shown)


ReinierCC:
image

ReinierCC:
VSIX for testing: [vscode-aks-tools-1.4.8-deploy-workflow-repo-population-fix.zip](https://github.com/user-attachments/files/16474756/vscode-aks-tools-1.4.8-deploy-workflow-repo-population-fix.zip)

src/commands/draft/draftCommands.ts
```diff
@@ -214,8 +214,8 @@ type DraftDependencies = {
 
 async function getGitHubAuthenticationSession(): Promise<Errorable<AuthenticationSession>> {
     try {
-        // No special scopes are required for GitHub - we are just listing repositories/branches.
-        const scopes: string[] = [];
+        // Repo scope required to see public/private repos.
```
hsubramanianaks, line 217: @ReinierCC looks good to me, can you provide the details of official documentation for these scopes would be nice to have for historical purpose. Also, previous developer had the scopes defined and removed it later here is the PR [link](https://github.com/Azure/vscode-aks-tools/pull/537#discussion_r1603326957), adding here for more context. I agree we need scopes repo here to make it work for forked repo's but we need documentation link. Thank you.

hsubramanianaks (commented):
looks good to me, Minor suggestion about the documentation. Thanks @ReinierCC

src/commands/draft/draftCommands.ts
ReinierCC: @hsubramanianaks Thanks for that suggestion.

Here's the documentation on scopes: https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps

Would this be the best place to document or should something like this be a comment next to the code?

hsubramanianaks: Comment will also be good, Thanks Reinier.

ReinierCC: Sounds good!

src/commands/draft/draftCommands.ts
```diff
@@ -229,13 +230,12 @@ async function getRepo(octokit: Octokit, remote: Remote): Promise<GitHubRepo | n
...
-        .replace(/\.git$/, "")
-        .split("/")
-        .slice(-2);
+    const parts = url.replace(/\.git$/, "").split(/[:/]/); //Split on both : and /, Removes .git
+    const [owner, repo] = parts.slice(-2);
+
     let response: RestEndpointMethodTypes["repos"]["get"]["response"];
     try {
-        response = await octokit.repos.get({ owner, repo });
+        response = await octokit.rest.repos.get({ owner, repo });
```
Tatsinnit, line 238: 💡 This is interesting: minor quesiton please: Can you please add doc as to why `octokit.repos.get(`  worked in same way as `octokit.rest.repos.get(` worth to archive in case this comes in any future discussion please. Thank you!!

ReinierCC: Good catch.

When debugging the bug, I could not get the original "octokit.repos.get(" to function. This was an older syntax used in [octokit v17](https://octokit.github.io/rest.js/v17l) and older.

The [current version](https://octokit.github.io/rest.js/v20) standard uses the rest namespace. It may be that when implemented previously an older version of the octokit was being utilized.

PR #840, opened 2024-08-05
[POC] Deploy local app to AKS cluster

kejatura-dev:


Tatsinnit:
Thank you so much @kejatura-dev for this and for the PR, Branch name and Fork would be ok but **_please feel free to covert this to Draft PR_** as this is for POC reasons only. 

Thank you so much for sharing context. ❤️🙏

PR #844, opened 2024-08-08
Escaping special regex chars

ReinierCC:
To properly sanitize input strings we have to escape any special characters.

Wanted to get some feedback on this implementation, where I manually specified the special characters and are escaping.

Some other potential routes would be to use a library with this feature but, this would create a new dependency.

Currently have a simple test function in the same file. Can move this to a more appropriate location after feedback.

### DIFF
--- a/src/panels/TcpDumpPanel.ts
+++ b/src/panels/TcpDumpPanel.ts
@@ -28,7 +28,7 @@ const captureFileBasePathEscaped = escapeRegExp(captureFileBasePath);
 const captureFilePathRegex = `${captureFileBasePathEscaped}(.*)\\.cap`; // Matches the part of the filename after the prefix
 
 // Escape all regex meta characters to ensure sanitation and '/' for later use in regex pattern
-function escapeRegExp(input: string): string {
+export function escapeRegExp(input: string): string {
     return input.replace(/[.*+?^${}()|[\]\\/]/g, "\\$&"); //Escapes by adding '\' to every matched string $&
 }
 //Reference for regex syntax chars: https://262.ecma-international.org/13.0/index.html#prod-SyntaxCharacter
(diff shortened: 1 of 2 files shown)


src/panels/TcpDumpPanel.ts
```diff
@@ -24,7 +24,14 @@ const tcpDumpCommandBase = "tcpdump --snapshot-length=0 -vvv";
 const captureDir = "/tmp";
 const captureFilePrefix = "vscodenodecap_";
 const captureFileBasePath = `${captureDir}/${captureFilePrefix}`;
-const captureFilePathRegex = `${captureFileBasePath.replace(/\//g, "\\$&")}(.*)\\.cap`; // Matches the part of the filename after the prefix
+const captureFileBasePathEscaped = escapeRegExp(captureFileBasePath);
+const captureFilePathRegex = `${captureFileBasePathEscaped}(.*)\\.cap`; // Matches the part of the filename after the prefix
+
+// Escape all regex meta characters to ensure sanitation and '/' for later use in regex pattern
+function escapeRegExp(input: string): string {
```
Tatsinnit, line 31: Thanks for this. 👍 worth to just check what is the standard `tcpDump` file name structure - I believe anything standardised file naming convention + `*.pcap` but worth to just verify please.

src/panels/TcpDumpPanel.ts
```diff
@@ -53,6 +60,20 @@ function getCaptureFromFilePath(filePath: string): string | null {
     return fileMatch && fileMatch[1];
 }
 
+// Test function to verify escaping
+function testEscapeRegExp() {
```
Tatsinnit, line 64: 💡 Idea: Can we please move this to the util file for latter wider consumption? if you think its a good idea? we can then test it via test suite run.

src/panels/TcpDumpPanel.ts
ReinierCC: Yes! you're right. ".pcap" is the standard convention. Currently ".cap" is referenced in many places so if we want to change the convention would probably be best for a separate PR. Lmn what you think.

src/panels/TcpDumpPanel.ts
ReinierCC: Added a mocha test suite for the escaping function

hsubramanianaks (approved):
lgtm, just make sure to run a manual test for tcp dump to make sure it works. Thanks for finding this and fixing it @ReinierCC

PR #859, opened 2024-08-14
possible fix.

Tatsinnit:
This PR is aimed to fix #852 

Thanks heaps and can we please get a review and probable test for the vsix attached below. 

cc: @qpetraroia , @hsubramanianaks , @ReinierCC and @tejhan thanks heaps 🙏☕️

VSIX below: 

[vscode-aks-tools-1.4.8-auth-util-update-test.vsix.zip](https://github.com/user-attachments/files/16617640/vscode-aks-tools-1.4.8-auth-util-update-test.vsix.zip)

### DIFF
--- a/src/tree/azureAccountTreeItem.ts
+++ b/src/tree/azureAccountTreeItem.ts
@@ -16,6 +16,7 @@ import { getCredential, getEnvironment } from "../auth/azureAuth";
 import { SelectionType, getSubscriptions } from "../commands/utils/subscriptions";
 import { Subscription } from "@azure/arm-resources-subscriptions";
 import { AzureSessionProvider, ReadyAzureSessionProvider, isReady } from "../auth/types";
+import { TokenCredential } from "@azure/identity";
 
 export function createAzureAccountTreeItem(
     sessionProvider: AzureSessionProvider,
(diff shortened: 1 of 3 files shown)


src/tree/azureAccountTreeItem.ts
```diff
@@ -200,5 +201,17 @@ function getSubscriptionContext(
...
         isCustomCloud: environment.name === "AzureCustomCloud",
+        createCredentialsForScopes: async (scopes: string[]): Promise<TokenCredential> => {
+            const authSession = await sessionProvider.getAuthSession({ scopes });
+            if (failed(authSession)) {
+                throw new Error(`No Microsoft authentication session found: ${authSession.error}`);
+            }
+
+            return {
+                getToken: async () => {
+                    return { token: authSession.result.accessToken, expiresOnTimestamp: 0 };
```
hsubramanianaks, line 212: so expiresOnTimestamp: 0  means this token never expires?

Tatsinnit: I believe that's the case, not documented outright but see its usage in this repo here: https://github.com/Azure/vscode-aks-tools/blob/main/src/auth/azureSessionProvider.ts#L288

hsubramanianaks: I tried setting the expiration time as 5 minutes but it didn't affect anything. But the official docs [here](https://learn.microsoft.com/en-us/javascript/api/@azure/core-auth/accesstoken?view=azure-node-latest#@azure-core-auth-accesstoken-expiresontimestamp) says it's the token expiration time. I think we need explore more on this. But for this PR it should be fine.

hsubramanianaks (approved):

PR #871, opened 2024-08-20
fix version to node-version

hsubramanianaks:
This PR fixes the github action for build with right syntax of node-version instead of version.


Tatsinnit (approved):
Nice!! also for archival reasons for folks who ask why, here read this please: https://github.com/actions/setup-node?tab=readme-ov-file#usage

PR #887, opened 2024-08-29
Added option to easily provide feedback

ReinierCC:
Calls the standard VS Code command to create a github issue. 

Screenshot 2024-08-29 at 11 10 42 AM
Screenshot 2024-08-29 at 11 11 11 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,6 +201,10 @@
                 "title": "Install Azure Service Operator"
             },
             {
+                "command": "aks.provideFeedback",
+                "title": "Report a Bug/Request a Feature"
+            },
+            {
                 "command": "aks.aksCRUDDiagnostics",
                 "title": "Create, Upgrade, Delete and Scale"
             },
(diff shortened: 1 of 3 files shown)


Tatsinnit:
Thanks for this, this needs discussion as to how this is getting bubbling up especially this submenu option, lets jump on a quick call, proving report bug from the submenu is confusing for various reasons and that should only be a thing when there is a failure or from some webviews.

The commands are the Actionable diagnostics or feature enable thing, we should conceptualise the difference between this specific thing vs enabling / diagnostic commands.

<3 rest that's for opening for thought process et. al.

ReinierCC:
Thanks Tats for the feedback.

We discussed that this concept of allowing extension users to report a bug would be useful but, would be more logical that only after an error occurs, the user would be prompted with the option to report the error. 

Going to be exploring how this can be accomplished.

PR #913, opened 2024-09-12
Proper URL Sanitization

ReinierCC:
Created whitelist of approved azure URLs based on currently utilized URLs.

### DIFF
--- a/src/tree/azureResourceNodeContributor.ts
+++ b/src/tree/azureResourceNodeContributor.ts
@@ -1,7 +1,6 @@
 import * as k8s from "vscode-kubernetes-tools-api";
 import * as vscode from "vscode";
 import { Errorable } from "../commands/utils/errorable";
-const fs = require("fs");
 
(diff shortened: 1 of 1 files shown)


hsubramanianaks (approved):
@ReinierCC  Thanks for tackling this interesting problem, only concern I have is what if we miss a service from the list and end up in errors or unexpected behavior since we saw this morning this ASO command is used frequently. Please test this well to make sure it works.

ReinierCC:
@hsubramanianaks 

`const azureResources = customResources.filter((r) => {
        const isDirectGroupMatch = approvedGroups.has(r.group);
        if (r.name.includes("azure.com") && !isDirectGroupMatch) {
            throw new Error(`Invalid Azure resource group: ${r.group}`);
        }
        return isDirectGroupMatch;
    });`
    
   Good point, I tested with the above code. This is also something that could be included as a safeguard.

Issue #930, opened 2024-09-19
AKS Welcome page ALWAYS opens


Jaykul:
**Describe the bug**

The AKS Extension (unlike **any** other extension I've used) keeps showing the welcome page every time I open VS Code, or open a new folder, regardless of the fact that there's no new news, and everything in there has been marked "Done". 

After the first dozen times a day it's actually quite annoying. See also: 
- https://github.com/microsoft/vscode/issues/212926
- https://github.com/microsoft/vscode/issues/112494
- https://github.com/alefragnani/vscode-bookmarks/issues/175

**Mention the platform you are using**

VSCode, Windows

**To Reproduce**

1. Install AKS Extension in VSCode
2. Open VS Code ("Welcome" shows)
   ![image](https://github.com/user-attachments/assets/53c210c8-a69f-4732-833b-fd2125968140)
3. Mark everything done ... 
2. Open VS Code ("Welcome" still shows)
3. Click to mark everything done ... again
4. Disable "Welcome" globally 
   ![image](https://github.com/user-attachments/assets/83c55a93-c46a-42ee-9756-41a3a0e27fed)
2. Open VS Code ("Welcome" still shows)

**Expected behavior**

1. Welcome pages should **only** show if I have "Show welcome page on startup"
2. Extension-specific welcome pages should **only** show if there's new information that's not marked done
3. Extension should register it's "Welcome" page in the command palette so I can open it later myself

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox

Tatsinnit:
Thank you so much for reporting this, really appreciate it! We've added a fix for this, and once tested and merged it will be part of next release. Thanks. Thank you so much for repro @ReinierCC for testing as well.

Tatsinnit:
💡A quick loop back, post release: the fix for this is released under version `1.4.10` thank you so much for this. ❤️ 🙏

PR #931, opened 2024-09-19
Fix for walkthrough and show welcome dependency.

Tatsinnit:
This fix is purely for the aspect of things which leads to the fixing #930 Now once user select the aka **uncheck** `show welcome page` the walk thought will not show up.

because upon uncheck the property in workspace gets set like this, and we are making sure walkthrough is not called until that is set. Thank you so much @ReinierCC for testing and fyi/collab. <3 

* VSIX for testing: 
[vscode-aks-tools-1.4.9-fixwalkthrough.vsix.zip](https://github.com/user-attachments/files/17067610/vscode-aks-tools-1.4.9-fixwalkthrough.vsix.zip)

* This one is proper zip file so unzip and then use the vsix under that to install.
[Uploading vscode-aks-tools-1.4.9-fixwalkthrough.vsix.zip…]()

Screenshot 2024-09-20 at 10 18 07 AM

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -69,9 +69,11 @@ export async function activate(context: vscode.ExtensionContext) {
         context.subscriptions.push(uiExtensionVariables.outputChannel);
 
         registerUIExtensionVariables(uiExtensionVariables);
-        vscode.commands.executeCommand("workbench.action.openWalkthrough", {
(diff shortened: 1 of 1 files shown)


ReinierCC (approved):
Looks good, tested VSIX as well.

tejhan (approved):
Tested & working good! 👍

Issue #944, opened 2024-09-24
"Automated Deployments: Draft a Dockerfile" Page freezes when creating python program dockerfile


ReinierCC:
**Describe the bug**

**To Reproduce**
Create any dockerfile for a python program.

Reason for bug:
![Screenshot 2024-09-23 at 3 22 08 PM](https://github.com/user-attachments/assets/606c98f6-df69-4de3-8cf8-71a9a7310041)

Page is freezing because the draft binary is waiting for user input. Draft has functionality to autodetect the entrypoint for python application but its cli requires user input to confirm the default entrypoint.

I've communicated with @davidgamero who said it might be useful to have a flag on the draft side to prevent additional dialog and use the default options to facilitate non cli interactions.

PR #945, opened 2024-09-24
Fixing freezing dockerfile create page 

ReinierCC:
This PR is to fix the following issue:

Screenshot 2024-09-24 at 11 15 44 AM

where the "Automated Deployments: Draft a Dockerfile" page would freeze for any user input after clicking the create button.

This occured because the draft command being executed to create the dockerfile was waiting for cli style user input about the BUILDVERSION.

Screenshot 2024-09-24 at 11 17 50 AM

Previously the builderImageTag was or'ed with an empty string. But builderImageTag was already of type string | null .
Screenshot 2024-09-24 at 11 19 46 AM

Screenshot 2024-09-24 at 11 21 28 AM

By removing the empty string, the draft binary uses BUILDERTAG as null and does not request user input.

There still exists issue #944 for python dockerfiles. But the fix for that would best come from a modification to the draft binary itself.

### DIFF
--- a/src/panels/draft/DraftDockerfilePanel.ts
+++ b/src/panels/draft/DraftDockerfilePanel.ts
@@ -135,7 +135,7 @@ export class DraftDockerfileDataProvider implements PanelDataProvider<"draftDock
         const variables = {
             PORT: port,
             VERSION: runtimeImageTag,
-            BUILDERVERSION: builderImageTag || "null",
+            BUILDERVERSION: builderImageTag,
         };
 
         const variableArgs = Object.entries(variables)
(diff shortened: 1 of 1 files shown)


ReinierCC:
VSIX For Testing (Remove .zip from file name):

[vscode-aks-tools-1.4.9-docker-create-freeze-fix.vsix.zip](https://github.com/user-attachments/files/17117153/vscode-aks-tools-1.4.9-docker-create-freeze-fix.vsix.zip)

Tatsinnit (approved):
Thank you so much for this, looks good to me. I will FYI this in our internal channel just as information. 🙏 ❤️

hsubramanianaks (approved):

PR #948, opened 2024-09-27
Add cluster options util for GH Copilot scenarios

kejatura-dev:
PR adds cluster option selection util for GitHub Copilot scenarios:
 
image

User has option to select three options: Create a new aks cluster, select existing cluster, or select recently used cluster

- **Create cluster:**
    - If selected, we will first ask user to select subscription then guide user to the create AKS panel.

- **Existing cluster:**
   - Guide user using command palette to select from list of subscriptions -> then select cluster. This selected cluster is saved as default in temporary storage for subsequent commands.

- **Recently used cluster:**
   - Utilizes temporary file storage to store recently selected cluster (from existing cluster flow) information into a temporary JSON file, which can be retrieved and saved.


src/plugins/shared/clusterOptions/selectClusterOptions.ts
```diff
@@ -0,0 +1,91 @@
...
+        {
+            label: "Create new cluster from AKS VS Extension",
+            type: Options.NewCluster
+        }
+    ];
+
+    return options;
+}
+
+export async function selectClusterOptions(sessionProvider: ReadyAzureSessionProvider): Promise<Errorable<ClusterPreference | boolean>> {
```
hsubramanianaks, line 41: @kejatura-dev I don't see this used anywhere or called from extensions.ts. This needs to be exposed as command instead of plugins/shared? what do you think ? I am thinking might be useful if in future there are other usecases for extension to re-use this code piece, since we cannot hack much of tree view part. Thoughts? If you agree, this can be exposed as src/commands/aksSelectCluster so that when GH copilot invokes it would be vscode.commands.executeCommand("aksSelectCluster", args); Like you do for the other GH copilot command for create automatic cluster ?

hsubramanianaks (changes requested):

src/plugins/shared/clusterOptions/selectClusterOptions.ts
kejatura-dev: Thanks Hari for reviewing, I agree we should create a VS Code command for this. I will add the command a follow up PR

src/commands/utils/tempfile.ts
```diff
@@ -16,9 +16,9 @@ export async function withOptionalTempFile<T>(
     }
 }
 
-export async function createTempFile(content: string, extension: string): Promise<TempFile> {
+export async function createTempFile(content: string, extension: string, prefix?: string): Promise<TempFile> {
```
Tatsinnit, line 19: 💡 Note to self: let's extract this to separate function which can be used for more broader reasons.

src/plugins/shared/clusterOptions/options/selectExistingClusterOption.ts
```diff
@@ -0,0 +1,102 @@
...
+import { longRunning } from "../../../../commands/utils/host";
+import { RecentCluster } from "../state/recentCluster";
+
+
+export async function selectExistingClusterOption(sessionProvider: ReadyAzureSessionProvider): Promise<Errorable<ClusterPreference>> {
+    // allow user to select subscriptions
+    await selectSubscriptions();
+
+    // get subscriptions
+    const subscriptions = getFilteredSubscriptions();
```
Tatsinnit, line 17: 💡 Error handling:

Key aspect here might be we need to add if missing is error_handlers. I can checkout the code and might see what's going on through this.

src/assets.ts
```diff
@@ -9,6 +9,10 @@ export function setAssetContext(context: ExtensionContext) {
     EXTENSION_CONTEXT = context;
 }
 
+export function getAssetContext(): ExtensionContext | null {
+    return EXTENSION_CONTEXT;
```
Tatsinnit, line 13: 💡 observation: what instance of this returns `null` given the return type is `ExtensionContext | null`

src/plugins/shared/clusterOptions/state/recentCluster.ts
```diff
@@ -0,0 +1,52 @@
+import * as fs from "fs/promises";
+import { getAssetContext } from "../../../../assets";
+import { ClusterPreference } from "../../types";
+import { Errorable } from "../../../../commands/utils/errorable";
+import { createTempFile, TempFile } from "../../../../commands/utils/tempfile";
+
+export class RecentCluster {
+    private static RECENT_CLUSTER_FILEPATH_KEY = "recent-cluster-file-path";
+    private static RECENT_CLUSTER_TEMPFILE: TempFile | undefined = undefined;
```
Tatsinnit, line 9: 💡 observation: Consistent naming: Replace `RECENT_CLUSTER_TEMPFILE` with a more consistent camelCase `recentClusterTempFile`.

Please note I have made these changes.

kejatura-dev: Thanks Tats for making these changes

src/plugins/shared/clusterOptions/selectClusterOptions.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export async function selectClusterOptions(sessionProvider: ReadyAzureSessionProvider, exclude?: SelectClusterOptions[]): Promise<Errorable<ClusterPreference | boolean>> {
+    const options = getClusterOptions();
+
+    let doesRecentClusterExist = false;
+    try {
+        doesRecentClusterExist = await RecentCluster.doesRecentlyUsedClusterExist();
+    } catch (error) {
+        // do nothing if recent cluster does not exist, we do not show this option
```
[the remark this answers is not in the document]
kejatura-dev: This is now updated. We now check if recentcluster exists before getting cluster. If it doesn't exist we just remove these options to the user

src/assets.ts
kejatura-dev: not entirely sure when this would be null, just referenced return type from line 6 of the file. But, if this is null, we handle it in recentCluster.ts

src/plugins/shared/clusterOptions/options/selectExistingClusterOption.ts
kejatura-dev: Yes good point, updated to handle if users don't have any subscriptions

src/commands/utils/tempfile.ts
Tatsinnit: Let me add this in this PR, will be much better for the entirety of this PR.

src/commands/utils/tempfile.ts
```diff
@@ -23,6 +23,13 @@ export async function createTempFile(content: string, extension: string): Promis
     return new TempFile(tempFile);
 }
 
+export async function createTempFileWithPrefix(content: string, extension: string, prefix: string): Promise<TempFile> {
+    // TODO: try/catch and return errorable?
```
hsubramanianaks, line 27: simple try/catch here?

src/commands/utils/clusters.ts
```diff
@@ -575,3 +576,31 @@ function isDefinedManagedCluster(cluster: azcs.ManagedCluster): cluster is Defin
...
 }
+
+export type Cluster = {
+    name: string;
+    clusterId: string;
+    resourceGroup: string;
+    subscriptionId: string;
+}
+
+export async function getClusters(sessionProvider: ReadyAzureSessionProvider, subscriptionId: string): Promise<Cluster[]> {
```
hsubramanianaks, line 587: suggestion here please to use javascript sdk:

```
export async function listManagedClusters(
    subscriptionId: string,
): Promise<Errorable<Cluster[]>> {
    const client = getAksClient(sessionProvider, subscriptionId);
    let clusters: Cluster[] = [];
    try {
        const managedClustersList = client.managedClusters.list();
        for await (const cluster of managedClustersList) {
            clusters.push({
                name: cluster.name!,
                clusterId: cluster.id!,
                resourceGroup: cluster.nodeResourceGroup!, // if you need resourceGroup you can pass it in or parse it from resourceID
                subscriptionId: subscriptionId,
            });
        }

        return { succeeded: true, result: clusters };
    } catch (e) {
        return { succeeded: false, error: `Failed to retrieve list of managed cluster for ${subscriptionId}: ${e}` };
    }
}
```

src/plugins/shared/clusterOptions/options/selectExistingClusterOption.ts
```diff
@@ -0,0 +1,77 @@
...
+): Promise<Errorable<ClusterPreference>> {
+    await selectSubscriptions();
+
+    const subscriptions = getFilteredSubscriptions();
+    if(subscriptions.length === 0) { 
+        handleNoSubscriptionsFound();
+        return { succeeded: false, error: "No subscriptions found." };
+    }
+
+    const clusters = (
```
hsubramanianaks, line 23: suggestion here to add: I hit this last time when I had a subscription without any aks clusters

```
cluster.length === 0 and show a error message ?

```

hsubramanianaks (approved):
lgtm but left some improvements which can be dealt with next set of PR's I will post it as improvement. For now it is good. Thanks @kejatura-dev

src/plugins/shared/clusterOptions/options/selectExistingClusterOption.ts
kejatura-dev: Ah yes, good one, will include error handling for no clusters in follow up PR

Tatsinnit (approved):
<3 Lets get this in, gentle fyi to @qpetraroia and @hsubramanianaks  + @kejatura-dev.

* Since both Hari and I have contributed to this, it will be worth to get one more approval please. (testing approval)

src/commands/utils/clusters.ts
kejatura-dev: Thanks Hari, Can make the improvements in follow up PR

qpetraroia (approved):

PR #961, opened 2024-10-01
role assignments for automatic sku

hsubramanianaks:
This PR address two issues 

1. enhances the automatic sku preset cluster creation with Azure Kubernetes Service RBAC Cluster Admin role assigned to the user whoever creates the cluster. 
2. fixes the deployment url same to portal deploment blades

### DIFF
--- a/src/panels/CreateClusterPanel.ts
+++ b/src/panels/CreateClusterPanel.ts
@@ -6,7 +6,7 @@ import { Uri, window } from "vscode";
 import { getEnvironment } from "../auth/azureAuth";
 import { AzureAuthenticationSession, ReadyAzureSessionProvider } from "../auth/types";
 import { getAksClient, getFeatureClient, getResourceManagementClient } from "../commands/utils/arm";
-import { getPortalResourceUrl } from "../commands/utils/env";
+import { getDeploymentPortalUrl, getPortalResourceUrl } from "../commands/utils/env";
 import { failed, getErrorMessage } from "../commands/utils/errorable";
 import {
     createMultipleFeatureRegistrations,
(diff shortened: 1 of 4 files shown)


src/panels/CreateClusterPanel.ts
```diff
@@ -421,3 +439,10 @@ function isRestError(ex: unknown): ex is RestError {
 function isDefaultK8sVersion(version: KubernetesVersion): boolean {
     return "isDefault" in version && version.isDefault === true;
 }
+function getServicePrincipalId(result: AzureAuthenticationSession): string {
```
Tatsinnit, line 442: 💡 Key note please: this only get and I am guessing we are not creating any service principal on the fly? also the sample below if is nothing active and well randomised. :)

hsubramanianaks: we are not just getting the id from already authenticated azure session.

hsubramanianaks: removed the sample value.

src/panels/templates/AutomaticCreateCluster.json
```diff
@@ -53,9 +176,21 @@
...
+                "supportPlan": "[parameters('supportPlan')]"
             },
             "identity": "[parameters('clusterIdentity')]"
+        },
+        {
+            "type": "Microsoft.Authorization/roleAssignments",
+            "apiVersion": "2020-04-01-preview",
+            "name": "[guid(resourceGroup().id)]",
+            "properties": {
+                "roleDefinitionId": "[concat('/subscriptions/', parameters('subscriptionId'), '/providers/Microsoft.Authorization/roleDefinitions/', 'b1ff04bb-8a4e-4dc4-8eb5-8693973ce19b')]",
```
hsubramanianaks, line 189: this is role definition ID of b1ff04bb-8a4e-4dc4-8eb5-8693973ce19b of Azure K8s Cluster Admin - https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles/containers#azure-kubernetes-service-rbac-cluster-admin

Tatsinnit (approved):
👍☕️

ReinierCC (approved):
LGTM

PR #1002, opened 2024-10-15
New Publish Action on Secure Runner

ReinierCC:


.github/workflows/publish-v2.yml
```diff
@@ -0,0 +1,116 @@
...
+        with:
+          auth-type: IDENTITY
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+
+      - name: Create a Github Release
+        id: create_release
+        uses: softprops/action-gh-release@v2
+        if: startsWith(github.ref, 'refs/tags/')
```
hsubramanianaks, line 103: I don't think we create tags and then start release; it is always anything on main branch get's released. Do we need this ?

.github/workflows/publish-v2.yml
```diff
@@ -0,0 +1,116 @@
...
+          files: '*.vsix'
+          fail_on_unmatched_files: true
+          tag_name : ${{ needs.JobBuild.outputs.currentversion }}
+          name: ${{ needs.JobBuild.outputs.currentversion}}
+          body: Publish ${{ needs.JobBuild.outputs.changelog_reader_changes }}
+        env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
+        
+      - name: Publish packaged extension
+        if: startsWith(github.ref, 'refs/tags/v')
```
hsubramanianaks, line 114: Same here, we release from main branch.

hsubramanianaks (changes requested):
Couple of minor changes to remove if condition. Thank you @ReinierCC

.github/workflows/publish-v2.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Create a Github Release
+        id: create_release
+        uses: softprops/action-gh-release@v2
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          files: '*.vsix'
+          fail_on_unmatched_files: true
+          tag_name : ${{ needs.JobBuild.outputs.currentversion }}
+          name: ${{ needs.JobBuild.outputs.currentversion}}
+          body: Publish ${{ needs.JobBuild.outputs.changelog_reader_changes }}
```
hsubramanianaks, line 109: Kind suggestion: there is flag called ```make_latest``` we can use it to mark the release as latest. I know we are testing this but just a suggestion :)

.github/workflows/publish-v2.yml
ReinierCC: great catch hari, thank you!

Tatsinnit:
❤️  Thank you for this, I wonder we give it a try , but with some really old version and I thin the over-ride will fail at least we should get 401 or something I guess?

.github/workflows/publish-v2.yml
```diff
@@ -0,0 +1,114 @@
...
+
+      # Ensure project builds successfully before creating release
+      - name: Build 
+        run: npm run webpack
+
+      - name: Get current package version
+        id: package_version
+        uses: martinbeentjes/npm-get-version-action@3cf273023a0dda27efcd3164bdfb51908dd46a5b # v1.3.1
+
+      - name: Check version is mentioned in Changelog
```
Tatsinnit, line 48: 💡Please note:  For the nature of testing sake, we might have to disable this **unless we find a way to test from fork** for a test run of this. :) so we don't have to make new changeling changes.

.github/workflows/publish-v2.yml
ReinierCC: Good point, after discussing with @Tatsinnit, we'll look to incorporate this after the initial test, to make sure we isolate any changes.

.github/workflows/publish-v2.yml
```diff
@@ -0,0 +1,106 @@
+name: Build & Publish
```
Tatsinnit, line 1: 💡 Lets rename this to more relevant name - so the existing one don't get mixed.

Tatsinnit (approved):
❤️ Thank you so much for this @ReinierCC , let's merge this progressive we can add back the `changelog` and `create the github` release back into the workflow post.

PR #1022, opened 2024-10-25
kaito needs to be KAITO named.

Tatsinnit:
This PR fixes the typo of misrepresenting Kaito as word to KAITO as I guess (Kubernetes AI ToolChain Operation) 

fyi to all.

Screenshot 2024-10-25 at 1 42 26 PM

Screenshot 2024-10-25 at 1 42 10 PM

PR #1023, opened 2024-10-25
Fix multiclick by disabling the button.

Tatsinnit:
This PR fixes few leftover KAITO stuff, which we noticed, it made sense to merge the long running PR and whatever was left we cover in small one part of which was covered in previous #1022 and next part is the disable of the button so user cannot press install multiple times which lead to failure of too many operation.

Idea for the implementation is: 

- The `isDisabled` state is **initialized** as `false`.
- After clicking the button, `onClickKaitoInstall` is triggered, and the button is disabled by setting `**isDisabled**` to `true`.
- The `disabled` attribute on the button prevents further clicks once it's **disabled**.

Thanks heaps, and result is here:

Screenshot 2024-10-25 at 3 26 35 PM

### DIFF
--- a/src/panels/KaitoPanel.ts
+++ b/src/panels/KaitoPanel.ts
@@ -224,7 +224,7 @@ export class KaitoPanelDataProvider implements PanelDataProvider<"kaito"> {
         };
 
         try {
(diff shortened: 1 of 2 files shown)


tejhan (approved):
LGTM 👍

PR #1025, opened 2024-10-25
Remove unused and unnecessary parameters for automatic cluster creation

hsubramanianaks:
This PR is to remove unused, unnecessary parameters from arm template for automatic sku cluster creation. Thanks to @sabbour   for pointing out the params and initial contribution via https://github.com/Azure/vscode-aks-tools/pull/1004


src/panels/utilities/ClusterSpecCreationBuilder.ts
```diff
@@ -182,7 +142,3 @@ function generateNodeResourceGroup(resourceGroupName: string, clusterName: strin
 function removeWhitespace(str: string): string {
     return str.replace(/\s+/g, "");
 }
-
-function generateRbacName() {
```
Tatsinnit, line 186: 💡Thank you for this, will take a look into this soon. A key note here please, is this going to affect the way `Dev/test` cluster works? or this part of code only touches the `automatic` cluster? I did notice how `automatic` cluster don't work with Kaito at the moment.

Tatsinnit (approved):
❤️ 💡  Thank you for this. I have left a key comment here for checking before we merge this, hopefully it's an isolated change for the `animatic cluster` related changes and will not effect the `dev/test` cluster functionality. 

* Comment left here: https://github.com/Azure/vscode-aks-tools/pull/1025#discussion_r1819882694 

Thanks!

src/panels/utilities/ClusterSpecCreationBuilder.ts
hsubramanianaks: this method was only used for automatic cluster, it should not affect dev preset cluster.

Tatsinnit: Thanks, sounds good!! I assume this is not the fix for the KAITO - install failing for `animatic cluster` asking here just to archive this as info. Thanks

PR #1049, opened 2024-11-07
Disable/Enable AKS Welcome page

kejatura-dev:


Tatsinnit (commented):
Thanks heaps for this, for wider folks its kind of inspired by the work here:

* https://github.com/Azure/vscode-aks-tools/compare/main...Tatsinnit:feature/gh-coilot-work-feature-flag?expand=1 

I wonder for now if we can se this to Draft PR please @kejatura-dev , what do you think? = )

PR #1094, opened 2024-11-26
Fix the dependabot related issue.

Tatsinnit:
Hiya, This PR fixes #1088 

### Sample VISX:

[vscode-aks-tools-1.5.2-dependabot-fix.vsix.zip](https://github.com/user-attachments/files/17910482/vscode-aks-tools-1.5.2-dependabot-fix.vsix.zip)

### Details

I think the error `TS2322: Type '{ children: unknown[]; }' is not assignable to type '{ children?: ReactNode; }'` occurs because the `children` prop in JSX expects a type assignable to `ReactNode`, and in the code, the `children` prop contains `unknown[]`, which is incompatible.

This issue typically arises from returning `<>{value}</>` where `value` has the type `unknown`. To resolve the error, you need to ensure that the `value` is safely cast or transformed into a type compatible with `ReactNode` before being rendered.

### Additional Details:

   - In cases where `value` is `unknown`, explicitly cast or transform it into a string using `String(value)`. This makes the type compatible with `ReactNode`.

   - For `ValueType.Bytes` and the default case, `value` is wrapped in `String(value)` to ensure it's rendered as a string.

   - For `ValueType.AddressArray`, ensure that the `value` is correctly parsed into a string array and joined with a separator. Ensure `JSON.parse` operates on valid JSON to avoid runtime errors.

This approach ensures type safety and fixes the `TS2322` error while maintaining the intended functionality.

fyi: please: ❤️ @hsubramanianaks , @ReinierCC , @tejhan , @qpetraroia

### DIFF
--- a/webview-ui/package-lock.json
+++ b/webview-ui/package-lock.json
@@ -29,7 +29,7 @@
         "eslint": "^8.57.0",
         "eslint-plugin-react": "^7.37.2",
         "eslint-plugin-react-hooks": "^5.0.0",
-        "typescript": "^5.6.3",
+        "typescript": "^5.7.2",
         "vite": "^5.4.11",
         "vite-plugin-checker": "^0.8.0",
         "vite-plugin-eslint": "^1.8.1"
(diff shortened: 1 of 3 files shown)


hsubramanianaks (approved):
Looks good to me, Hope testing worked for you. Thank you @Tatsinnit

ReinierCC:
Screenshot 2024-11-26 at 3 13 04 PM

When running traces on Windows had some of the namespaces undefined. Not sure if this issue is related.

ReinierCC:
Screenshot 2024-12-02 at 5 51 43 PM
Tested new changes, and functionality looks good!

ReinierCC (approved):

PR #1120, opened 2024-12-11
add node os upgrade channel

hsubramanianaks:
This PR fixes an issue where nodeOSUpgradeChannel was missing in the properties for AKS cluster creation ARM template causing CVE alerts. This has been set to NodeImage which updates NodeImage weekly - more details below 
https://learn.microsoft.com/en-us/azure/aks/auto-upgrade-node-os-image?tabs=azure-cli#update-ownership-and-schedule

image

customers control these upgrades via maintenance configurations. More details here - https://learn.microsoft.com/en-us/azure/aks/auto-upgrade-node-os-image?tabs=azure-cli#node-os-planned-maintenance-windows


Tatsinnit (approved):
❤️ Thank you so much for this, Looks great to me, we can get this tested and ship it <3

ReinierCC (approved):
LGTM

PR #1159, opened 2025-01-07
Fix the upgrade for eslint 9.17.0.

Tatsinnit:
Possible fix for https://github.com/Azure/vscode-aks-tools/pull/1132 

Opening this PR to check if all checks will pass. 👮‍♂️ 

very early checkin, but we should also check after removing `cjs` files as those must be redundant now.

* auto generated files / migration: https://eslint.org/docs/latest/use/configure/migration-guide#migrate-your-config-file 

thanks. and ❤️  cc: @tejhan , @ReinierCC , @hsubramanianaks


tejhan (approved):
Built & tested locally, LGTM 👍

PR #1161, opened 2025-01-13
Create Fleet: panel implementation

JunyuQian:
### Description

This pull request introduces the basic panel implementation and required webview definitions for creating fleets in the VSCode extension.

### Key Changes

1. **Added Fleet Management Capabilities**:
    - Introduced `ContainerServiceFleetClient` to manage AKS fleets.
    - Implemented `getAksFleetClient` function in `arm.ts` to create a fleet client.
    - Added `CreateFleetPanel` and `CreateFleetDataProvider` classes in `CreateFleetPanel.ts` for handling fleet creation UI and logic.

2. **New Webview Definitions**:
    - Created `createFleet.ts` under `webviewDefinitions` to define interfaces and message types related to fleet creation.
    - Updated `webviewTypes.ts` to include `CreateFleetDefinition`.

3. **UI Enhancements**:
    - Added placeholder for `createFleet` in `main.tsx` and `manualTest/main.tsx` to ensure compilation and future implementation. The code in `main.tsx` is hardcoded to satisfy dependencies for the project to compile.

### Changes

- **Files Modified**:
    - `src/commands/utils/arm.ts`: Added `ContainerServiceFleetClient` and `getAksFleetClient`.
    - `src/panels/CreateFleetPanel.ts`: Created new panel and data provider for fleet creation.
    - `src/webview-contract/webviewDefinitions/createFleet.ts`: Defined types for fleet creation webview.
    - `src/webview-contract/webviewTypes.ts`: Added `CreateFleetDefinition`.
    - `webview-ui/src/main.tsx`: Added placeholder for fleet creation with hardcoded code to satisfy dependencies.
    - `webview-ui/src/manualTest/main.tsx`: Added placeholder for fleet creation test scenarios with hardcoded code to satisfy dependencies.

### Testing

- Placeholder components have been added to ensure dependencies are resolved during compilation.
- Full implementation and test cases will be added in subsequent PRs.

### Notes

- The right-click entry point for `createFleet` is not yet visible to users to avoid incomplete experience.
- The creation of resource groups is yet to be included.
- Test cases are yet to be added.

### DIFF
--- a/src/commands/utils/arm.ts
+++ b/src/commands/utils/arm.ts
@@ -13,6 +13,7 @@ import { getCredential, getEnvironment } from "../../auth/azureAuth";
 import { ReadyAzureSessionProvider } from "../../auth/types";
 import { Errorable, getErrorMessage } from "./errorable";
 import { ComputeManagementClient } from "@azure/arm-compute";
+import { ContainerServiceFleetClient } from "@azure/arm-containerservicefleet";
 
 export function getSubscriptionClient(sessionProvider: ReadyAzureSessionProvider): SubscriptionClient {
     return new SubscriptionClient(getCredential(sessionProvider), { endpoint: getArmEndpoint() });
@@ -34,6 +35,15 @@ export function getAksClient(
     subscriptionId: string,
 ): ContainerServiceClient {
     return new ContainerServiceClient(getCredential(sessionProvider), subscriptionId, { endpoint: getArmEndpoint() });
+}
+
+export function getAksFleetClient(
+    sessionProvider: ReadyAzureSessionProvider,
+    subscriptionId: string,
+): ContainerServiceFleetClient {
+    return new ContainerServiceFleetClient(getCredential(sessionProvider), subscriptionId, {
+        endpoint: getArmEndpoint(),
+    });
 }
 
 export function getMonitorClient(sessionProvider: ReadyAzureSessionProvider, subscriptionId: string): MonitorClient {
--- a/src/panels/CreateFleetPanel.ts
+++ b/src/panels/CreateFleetPanel.ts
@@ -1,4 +1,132 @@
 import { ContainerServiceFleetClient, Fleet } from "@azure/arm-containerservicefleet";
+import { BasePanel, PanelDataProvider } from "./BasePanel";
+import { Uri, window } from "vscode";
+import { MessageHandler, MessageSink } from "../webview-contract/messaging";
+import {
+    InitialState,
+    ProgressEventType,
+    ToVsCodeMsgDef,
+    ToWebViewMsgDef,
+    ResourceGroup as WebViewResourceGroup,
+} from "../webview-contract/webviewDefinitions/createFleet";
+import { TelemetryDefinition } from "../webview-contract/webviewTypes";
+import { ResourceGroup as ARMResourceGroup, ResourceManagementClient } from "@azure/arm-resources";
+import { ReadyAzureSessionProvider } from "../auth/types";
+import { getAksFleetClient, getResourceManagementClient } from "../commands/utils/arm";
+import { getResourceGroups } from "../commands/utils/resourceGroups";
+import { failed } from "../commands/utils/errorable";
+
+export class CreateFleetPanel extends BasePanel<"createFleet"> {
+    constructor(extensionUri: Uri) {
+        super(extensionUri, "createFleet", {
+            getLocationsResponse: null,
+            getResourceGroupsResponse: null,
+            progressUpdate: null,
+        });
+    }
+}
+
+export class CreateFleetDataProvider implements PanelDataProvider<"createFleet"> {
+    private readonly resourceManagementClient: ResourceManagementClient;
+    private readonly fleetClient: ContainerServiceFleetClient;
+
+    constructor(
+        private readonly sessionProvider: ReadyAzureSessionProvider,
+        private readonly subscriptionId: string,
+        private readonly subscriptionName: string,
+    ) {
+        this.resourceManagementClient = getResourceManagementClient(sessionProvider, this.subscriptionId);
+        this.fleetClient = getAksFleetClient(sessionProvider, this.subscriptionId);
+    }
+
+    getTitle(): string {
+        return `Create Fleet in ${this.subscriptionName}`;
+    }
+
+    getInitialState(): InitialState {
+        return {
+            subscriptionId: this.subscriptionId,
+            subscriptionName: this.subscriptionName,
+        };
+    }
+
+    getTelemetryDefinition(): TelemetryDefinition<"createFleet"> {
+        return {
+            getResourceGroupsRequest: false,
+            getLocationsRequest: false,
+            createFleetRequest: true,
+        };
+    }
+
+    getMessageHandler(webview: MessageSink<ToWebViewMsgDef>): MessageHandler<ToVsCodeMsgDef> {
+        return {
+            getLocationsRequest: () => this.handleGetLocationsRequest(webview),
+            getResourceGroupsRequest: () => this.handleGetResourceGroupsRequest(webview),
+            createFleetRequest: (args) =>
+                this.handleCreateFleetRequest(args.resourceGroupName, args.location, args.name),
+        };
+    }
+
+    private async handleGetLocationsRequest(webview: MessageSink<ToWebViewMsgDef>) {
+        const provider = await this.resourceManagementClient.providers.get("Microsoft.ContainerService");
+        const resourceTypes = provider.resourceTypes?.filter((type) => type.resourceType === "fleets");
+        if (!resourceTypes || resourceTypes.length > 1) {
+            window.showErrorMessage(
+                `Unexpected number of fleets resource types for provider (${resourceTypes?.length || 0}).`,
+            );
+            return;
+        }
+
+        const resourceType = resourceTypes[0];
+        if (!resourceType.locations || resourceType.locations.length === 0) {
+            window.showErrorMessage("No locations found for fleets resource type.");
+            return;
+        }
+
+        webview.postGetLocationsResponse({ locations: resourceType.locations });
+    }
+
+    private async handleGetResourceGroupsRequest(webview: MessageSink<ToWebViewMsgDef>) {
+        const groups = await getResourceGroups(this.sessionProvider, this.subscriptionId);
+        if (failed(groups)) {
+            webview.postProgressUpdate({
+                event: ProgressEventType.Failed,
+                operationDescription: "Retriving resource groups for fleet creation",
+                errorMessage: groups.error,
+                deploymentPortalUrl: null,
+                createdFleet: null,
+            });
+            return;
+        }
+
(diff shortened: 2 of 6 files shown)


src/commands/utils/arm.ts
```diff
@@ -36,6 +37,15 @@ export function getAksClient(
     return new ContainerServiceClient(getCredential(sessionProvider), subscriptionId, { endpoint: getArmEndpoint() });
 }
 
+export function getAksFleetClient(
+    sessionProvider: ReadyAzureSessionProvider,
+    subscriptionId: string,
+): ContainerServiceFleetClient {
+    return new ContainerServiceFleetClient(getCredential(sessionProvider), subscriptionId, {
+        endpoint: getArmEndpoint(),
+    });
```
serbrech, line 46: nit: the indenting in this file seems to be a bit all over the place.
- arguments on separate lines vs single line
- named argument vs implicit named arguments

@Tatsinnit Do you use a code formatter/linter in the repo? if not, should be considered to remove this type of concerns/review imo.

src/panels/CreateFleetPanel.ts
```diff
@@ -1,4 +1,132 @@
...
+    ResourceGroup as WebViewResourceGroup,
+} from "../webview-contract/webviewDefinitions/createFleet";
+import { TelemetryDefinition } from "../webview-contract/webviewTypes";
+import { ResourceGroup as ARMResourceGroup, ResourceManagementClient } from "@azure/arm-resources";
+import { ReadyAzureSessionProvider } from "../auth/types";
+import { getAksFleetClient, getResourceManagementClient } from "../commands/utils/arm";
+import { getResourceGroups } from "../commands/utils/resourceGroups";
+import { failed } from "../commands/utils/errorable";
+
+export class CreateFleetPanel extends BasePanel<"createFleet"> {
```
serbrech, line 19: can/should this be a const string variable? it appears a number of times throughout the file

src/panels/CreateFleetPanel.ts
```diff
@@ -1,4 +1,132 @@
...
+            await createFleet(this.fleetClient, resourceGroupName, name, resource);
+        } catch (error) {
+            console.log(error);
+        }
+    }
+}
+
+function isValidResourceGroup(group: ARMResourceGroup): group is WebViewResourceGroup {
+    if (!group.name || !group.id) return false;
+    if (group.name?.startsWith("MC_")) return false;
```
serbrech, line 126: that's interesting, where does this validation rule come from?

src/webview-contract/webviewDefinitions/createFleet.ts
```diff
@@ -0,0 +1,52 @@
...
+    subscriptionId: string;
+    subscriptionName: string;
+}
+
+export interface ResourceGroup {
+    name: string;
+    location: string;
+}
+
+export enum ProgressEventType {
```
serbrech, line 13: are there shared "progressEventTypes" in the extension?
are there standard ones?

src/commands/utils/arm.ts
Tatsinnit: =) we **don't** use an specific code formatter for this repo, `eslint` are enable for code **type** smells, but this is local user level code formatter tool.

I like your idea: Most dev does use the `prettier` as preferred so the file passes the prettier check but this specific code uses 

We could use `npm install --save-dev prettier` but that I can bring in our scrum meetup. (Does sound like good idea) - Let me add it Just for fun. (Check the prettier run below could fix some code formatting evangelism.)

 Adding below stuff for Junyu to just have fun with I guess

For example: if my machine has `prettier` formatter and below is the indentation I get so if I use it it will indent like this, and I believe in this case might be what junyu has is `"vscode.typescript-language-features"`

* This could guide how to change this formatter locally: https://stackoverflow.com/a/58246538 

* https://code.visualstudio.com/docs/languages/typescript#_formatting 

```*.ts
export function getAksFleetClient(
    subscriptionId: string
): ContainerServiceFleetClient {
    return new ContainerServiceFleetClient(
        getCredential(sessionProvider),
        subscriptionId,
        {
            endpoint: getArmEndpoint(),
        }
    );
}
```

`Cmd + k + f` gives that option: 

Screenshot 2025-01-13 at 9 50 06 PM

You can also reset it by doing `ctrl + shift + p` and then `settings.json` file part to `null` and then reset to whatever formatter you like.

Screenshot 2025-01-13 at 9 54 48 PM

**Extra**

Screenshot 2025-01-13 at 10 13 03 PM

tejhan (commented):
Setup of everything LGTM !👍, would say some descriptive comments for some functions could help (Like for the validation criteria being used). I'm assuming the ProgressEventTypes are placeholders for now & could be modified in further PR's.

src/panels/CreateFleetPanel.ts
dvadas: It looks like it's copy/pasted from https://github.com/Azure/vscode-aks-tools/blob/a1051c316eee33621b5e528f8d3c8c9405ac2024/src/panels/CreateClusterPanel.ts#L174, as most of the code here is based on that file.

It might be better to factor this function out to another file. Maybe src/panels/utilities/webview.ts?

src/panels/CreateFleetPanel.ts
```diff
@@ -1,4 +1,132 @@
...
+                createdFleet: null,
+            });
+            return;
+        }
+
+        const usableGroups = groups.result.filter(isValidResourceGroup).map((group) => ({
+            label: `${group.name} (${group.location})`,
+            name: group.name,
+            location: group.location,
+        }));
```
dvadas, line 106: The createClusterPanel code also does an alphabetic sort here, which I imagine is a lot better for the user. Could you add that in?

It's a good example of why factoring out common code is a good idea.

src/panels/CreateFleetPanel.ts
JunyuQian: Hi Stephane, yeah as David mentioned this function is from createCluster.

I wasn't entirely sure about the validation rule, but I felt it might relate to a security aspect and hence, I decided to include it in createFleet.

src/webview-contract/webviewDefinitions/createFleet.ts
JunyuQian: Yep I also feel like it'd be better to have a shared "progressEventTypes", as the one in createFleet.ts is the same as the one in createCluster.ts (src/webview-contract/webviewDefinitions/createCluster.ts).

The current plan is to complete the full lifecycle of createFleet first and then consider code refactoring and optimization. But I can also do it now if that's a better practice.

src/panels/CreateFleetPanel.ts
JunyuQian: Yeah that sounds great!

I'll reachout to Tatsat for further discussions regarding the feasibility and potential issues of this refactoring 😄

src/panels/CreateFleetPanel.ts
serbrech: We don't have this restriction on Fleet api.
I think you should remove this validation from this PR.
the validation should match our service rules if they are added client side or just rely on the server side validation and not care about it here.

src/webview-contract/webviewDefinitions/createFleet.ts
serbrech: I think that's a good plan.
leave it as is, and send a separate PR for the small refactoring changes later.
resolving this comment

src/panels/CreateFleetPanel.ts
```diff
@@ -1,4 +1,124 @@
...
+    }
+
+    getInitialState(): InitialState {
+        return {
+            subscriptionId: this.subscriptionId,
+            subscriptionName: this.subscriptionName,
+        };
+    }
+
+    getTelemetryDefinition(): TelemetryDefinition<typeof contentId> {
```
serbrech, line 54: is typeof contentId going to work?
I don't know typescript well enough to answer that, but syntax wise, typeof contentId looks like it would return a string type, and not the value "createFleet". is that a problem?

dvadas: I was wondering that too, but I think it does make sense with `typeof`. When you look at how the generic type is used, it's matched up to the keys of [this `AllWebviewDefinitions` object](https://github.com/Azure/vscode-aks-tools/blob/main/src/webview-contract/webviewTypes.ts#L39), which are themselves strings. You can in Typescript define a type which is just a union of string values (e.g. for an enum):
```
type colorEnum = "Red" | "Blue" | "Green"
```
So this is doing effectively the same thing as that, just with only a single legal value.

dvadas (approved):

src/panels/CreateFleetPanel.ts
```diff
@@ -1,4 +1,124 @@
...
+import { TelemetryDefinition } from "../webview-contract/webviewTypes";
+import { ResourceManagementClient } from "@azure/arm-resources";
+import { ReadyAzureSessionProvider } from "../auth/types";
+import { getAksFleetClient, getResourceManagementClient } from "../commands/utils/arm";
+import { getResourceGroups } from "../commands/utils/resourceGroups";
+import { failed } from "../commands/utils/errorable";
+
+const contentId = "createFleet";
+
+export class CreateFleetPanel extends BasePanel<typeof contentId> {
```
hsubramanianaks, line 20: it should be extends BasePanel<"createFleet"> would work, base panel is abstract class that has the basic methods for a panel and data providers.

hsubramanianaks: this type refers to the one you have defined it in webviewTypes.ts file @JunyuQian 

image

src/panels/CreateFleetPanel.ts
hsubramanianaks: TelemetryDefinition<"createFleet"> would work here as well.

Tatsinnit (dismissed):
🐲  Thank you for this PR—this second part of the web view looks good to me! While it’s not yet connected, it’s clear that it’s heading in the right direction, and the U/X folks seem happy with it as well.

We’re now entering a crucial stage beyond this point, where connecting everything together will become key. I’d strongly encourage you to engage with Tejhan and Reinier pro-actively with the next PR, as it will likely be even more critical in tying everything together and ensuring success.

Once again, thank you for your efforts on this PR—it’s an important step forward. Feel free to reach out if there’s anything I can help with, or if more collaboration is needed as we move toward the next phase!

Humble thanks to everyone above for valuable comments. 

* Once you cater left over comments, we can plan to `squash, merge` this mid-day today. 🤞 thanks.

ReinierCC:
Looks good! Solid foundation so far.

As you start connecting things the next tricky part will be setting up the "state.ts" file which can be thought as the message/event handling on the webview side. 

I'd recommend you try to load a simple "Hello World" before tackling the full fleet webview.

Feel free to reach out with any questions!

ReinierCC (approved):

JunyuQian:
Thank you all for your valuable comments!! I have carefully reviewed them and addressed all the issues in the latest commit :)

Tatsinnit (approved):

Issue #1162, opened 2025-01-13
Add Code Formatter Check!! Thoughts?


Tatsinnit:
This idea came through suggestions and discussion under this thread. https://github.com/Azure/vscode-aks-tools/pull/1161/files#r1912891247

I think given we have many contributors this could be a nice idea, hence sharing jsut for wider community thoughts. ❤ 

To use a TypeScript code formatter in a GitHub Actions workflow, a popular choice is **Prettier**, a widely adopted code formatter for TypeScript, JavaScript, and other languages. Here's how we can set it up:

Does anyone has any specific preference?

### 1. **Add Prettier to Your Project**

First, ensure Prettier is installed in our project:

```bash
npm install --save-dev prettier
```

Optionally, we can create a `.prettierrc` file to configure formatting preferences:

```json
{
  "semi": true,
  "singleQuote": true,
  "tabWidth": 2
}
```

---

### 2. **Set Up Prettier Check in GitHub Actions**

Create a GitHub Actions workflow to check formatting on pull requests or pushes. Add the following YAML file to `.github/workflows/format-check.yml`:

```yaml
name: Prettier Check

on:
  pull_request:
    paths:
      - '**/*.ts'
      - '**/*.tsx'
  push:
    branches:
      - main

jobs:
  prettier-check:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout repository
        uses: actions/checkout@v3

      - name: Setup Node.js
        uses: actions/setup-node@v3
        with:
          node-version: '18'

      - name: Install dependencies
        run: npm install

      - name: Run Prettier check
        run: npx prettier --check "**/*.{ts,tsx,json,css,md}"
```

---

### 3. **Run the Workflow**

- On each pull request, the workflow will check if the TypeScript files adhere to Prettier's formatting rules.
- If formatting issues are detected, the action will fail and list the files that need reformatting.

---

Or we could add pre commit check like `Husky` 

Please feel free to add thoughts and Idea. Thanks.

serbrech:
> Optionally, we can create a .prettierrc file to configure formatting preferences

That's the approach I would take. Encode it as part of the repo.
Here is the prettierrc file from the api spec repo:

https://github.com/Azure/azure-rest-api-specs/blob/main/.prettierrc.json

PR #1163, opened 2025-01-13
Add prettier format check.

Tatsinnit:
This PR take care of this work item: https://github.com/Azure/vscode-aks-tools/issues/1162 

For now its for thought share for folks.

thanks heaps.

### DIFF
--- a/.github/workflows/format-check.yml
+++ b/.github/workflows/format-check.yml
@@ -1,0 +1,29 @@
+name: Prettier Check
+
+on:
+  pull_request:
+    paths:
+      - '**/*.ts'
+      - '**/*.tsx'
+  push:
+    branches:
+      - main
+
+jobs:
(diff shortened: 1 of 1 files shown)


ReinierCC:
Looks very pretty (lol). 

💡 Is it possible to separate the needed changes to get the prettier format check to work and the format changes it would make to the repo?

tejhan (dismissed):
The GH action LGTM. 👍

I'm assuming moving forward we will all use the prettier package cleanup before pushing a PR? 

Also would a prettier.rc file need to also be added to the repo to ensure PR code is checked for the same formatting rules regardless of who opens one up? (As mentioned in #1162) (Wondering if any slight issue could arise with people using their own local config files)

ReinierCC (approved):
LGTM

PR #1193, opened 2025-01-22
Initial Setup for Automated Deployments (Panel & Webview)

ReinierCC:
Putting this PR out with current state of feature development, before it becomes much more complex with full functionality.

Next PR will build out backend functionality on Panel to support UI design which is still being created.

Current functionality is **NOT** available to user.


Tatsinnit (approved):
❤️ Looks good to me!! We have removed a key issue which was previously causing the dependency alert <3 Thanks heaps for this!!

src/commands/devhub/aksAutomatedDeployments.ts
```diff
@@ -0,0 +1,169 @@
...
+        vscode.window.showErrorMessage(sessionProvider.error);
+        return;
+    }
+
+    if (!cloudExplorer.available) {
+        vscode.window.showWarningMessage(`Cloud explorer is unavailable.`);
+        return;
+    }
+
+    if (!clusterExplorer.available) {
```
hsubramanianaks, line 28: I think this is duplicate. can you remove it please

src/commands/devhub/aksAutomatedDeployments.ts
```diff
@@ -0,0 +1,169 @@
...
+            return gitHubOAuthResp.token;
+        }
+    } catch (error) {
+        console.error("Error occurred during GitHub OAuth:", error);
+        return undefined;
+    }
+}
+
+// Utility function to poll for a token
+async function pollForToken(
```
hsubramanianaks, line 144: May be move this to utils?

src/extension.ts
```diff
@@ -118,6 +119,7 @@ export async function activate(context: vscode.ExtensionContext) {
         registerCommandWithTelemetry("aks.aksOpenKubectlPanel", aksOpenKubectlPanel);
         registerCommandWithTelemetry("aks.getAzureKubernetesServicePlugins", getPlugins);
         registerCommandWithTelemetry("aks.aksDraftValidate", draftValidate);
+        //registerCommandWithTelemetry("aks.aksAutomatedDeployments", aksAutomatedDeployments);
```
hsubramanianaks, line 122: these are placeholders?

src/commands/devhub/aksAutomatedDeployments.ts
ReinierCC: Good catch. Thank you!

src/extension.ts
ReinierCC: Yes! When command is ready to be shown to user, will get uncommented

hsubramanianaks (dismissed):
looks fine for boilerplate code, but left some minor comments. Hope this get's cleaned up in the upcoming PR's. Thank you Reinier.

Tatsinnit (approved):

src/commands/devhub/aksAutomatedDeployments.ts
ReinierCC: Seems like it was similar naming but different objects. 😆 Thought it was a duplicate as well.

src/commands/devhub/aksAutomatedDeployments.ts
ReinierCC: Future PRs will have a different implementation for auth flow.

PR #1179, opened 2025-01-17
Create Fleet: Full Cycle Completed

JunyuQian:
# Create Fleet: Full Cycle Completed

This pull request completes the full cycle of creating a fleet. Starting from right-clicking on the subscription, then loading the front-end UI after clicking "Create Fleet," prompting the user to enter inputs, validating the inputs, and finally creating the fleet which is visible in the Azure portal.

### New Files:
- **CreateFleet.tsx**: The front-end UI for creating a fleet, which changes based on the current state of creating a fleet. For example, displaying a "Loading" message while fetching the existing resource groups and available locations, and calling the `CreateFleetInput` to show the input form for the user to enter the fleet name, resource group, and location.
- **CreateFleetInput.tsx**: The input form that will be displayed when the state is in `CollectingInput`. It allows the user to input fleet information and provides an input validation method to ensure the inputs are legal before calling the API and creating the resource.
- **state.ts**: Defines the necessary structures and event handlers, handling various stages of the fleet creation process.

### Modified Files:
- **askFleetManager.ts**: Removed the hardcoded parts, added the code to display the panel for user inputs, and the data provider to pass the front-end inputted data to the backend API.
- **CreateFleetPanel.ts**: Added webview to the function parameters, handling the `InProgress`, `Success`, and `Failed` states, and the transitions among these states. The URL for the user to click if the resources are created successfully is also added in this file.
- **webview-ui/src/main.tsx**: Replaced the hardcoded parts with the actual function call.
- **src/extension.ts**: Added the function hook for fleet creation.

## Note
The right-click option (the entry point) is still not enabled in this PR. To use the function, modifying `package.json` is required.

## Next Steps
- May include the testing code for this PR later.
- If test cases are not included in this PR, they will be in the next one, together with UI enhancement and extra detailed parameters for the fleet creation.

PR #1194, opened 2025-01-24
Create Fleet: UX Enhancement and Webview Testing

JunyuQian:
# Create Fleet: UX Enhancement and Webview Testing

This PR is the release version of the Create Fleet feature, with the user entry point now enabled. It adds two extra parameters for user input, completes the webview testing file for creating a fleet, and enhances the UI for a better user experience.
image

### New Files
- **createFleetTests.tsx:** The webview testing file for creating a fleet. It uses the input file directly and makes mock responses to locally simulate a full cycle of fleet creation. To enter the "Failed" state and the "Cancelled" state, choose "thiswillfail" and "thiswillbecancelled" respectively for the location field. All other options bring the user to the "Success" state.
- **CreateFleetModeInput.tsx:** A subfile of `CreateFleetInput`, which displays the extra two input parameters: the HubClusterMode selection button and the DnsPrefix input box.

### Modified Files
- **package.json:** Added necessary code for enabling the entry point of the fleet manager (which includes the create fleet feature).
- **extension.ts:** Added the command hook so that the `askCreateFleet` function is called when the user clicks the "Create Fleet" button.
- **CreateFleetPanel.ts:** Edited the function parameters for `createFleet` to include the extra two parameters. The error message is now more detailed for a better troubleshooting experience.
- **createFleet.ts:** Added the two extra parameters to the `createFleet` parameter list and created a new enum structure `HubMode` to record hub cluster mode.
- **CreateFleet.tsx:** Improved program efficiency by adding an empty list to the function parameter of React `useEffect`.
- **CreateFleetInput.tsx:** Added a function call to the new child class `CreateFleetModeInput.tsx` to enable the extra input fields. Also added a reference to the new CSS file to apply formatting rules and enhance UX.
- **manualTest/main.tsx:** Removed the hard-coded webview test case scenarios and replaced them with the new webview testing file `createFleetTests.tsx`.

## Next Steps (Potential UX Enhancement)
- **DNS Prefix Auto Fill-in Feature:** Like the portal, the DNS prefix will be auto-filled based on the user's input in the name field, unless the user manually inputs a different DNS prefix.
- **Highlight Missing Required Fields:** If there are required fields that are not filled in when the user submits the form, it would be better to show which fields are missing and highlight them for the user to fill in.

## Refactor Suggestions
- **Unify File Names:** The file names can be unified (e.g., using either UpperCamelCase or lowerCamelCase).
- **Extract and Generalize VSCode Buttons:** VSCode buttons can be extracted and used as general buttons.

### DIFF
--- a/package.json
+++ b/package.json
@@ -331,6 +331,10 @@
             {
                 "command": "aks.aksKaitoTest",
                 "title": "Test KAITO models"
+            },
+            {
+                "command": "aks.aksCreateFleet",
+                "title": "Create Fleet"
             }
         ],
         "menus": {
@@ -442,6 +446,10 @@
                     "submenu": "aks.kaitoInstallSubMenu",
                     "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.cluster/i",
                     "group": "9@3"
+                },
+                {
+                    "submenu": "aks.fleetMangerSubMenu",
+                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /aks\\.subscription/i"
                 }
             ],
             "aks.createClusterSubMenu": [
@@ -541,6 +549,12 @@
                     "command": "aks.aksReconcileCluster",
                     "group": "navigation"
                 }
+            ],
+            "aks.fleetMangerSubMenu": [
+                {
+                    "command": "aks.aksCreateFleet",
+                    "group": "navigation"
+                }
             ]
         },
         "submenus": [
@@ -571,6 +585,10 @@
             {
                 "id": "aks.kaitoInstallSubMenu",
                 "label": "Deploy a LLM with KAITO"
+            },
+            {
+                "id": "aks.fleetMangerSubMenu",
+                "label": "Fleet Manager"
             }
         ]
     },
--- a/src/panels/CreateFleetPanel.ts
+++ b/src/panels/CreateFleetPanel.ts
@@ -128,7 +128,9 @@ export class CreateFleetDataProvider implements PanelDataProvider<"createFleet">
         webview: MessageSink<ToWebViewMsgDef>,
     ) {
         const hubProfile = {
-            // If HubClusterMode is "With", use the given dnsPrefix; Otherwise set it to undefine
+            // The variable that determines whether the Fleet is created with a hub cluster.
+            // If hubProfile is provided in the API call, the Fleet will be created with a hub cluster.
+            // More Info: https://learn.microsoft.com/en-us/azure/kubernetes-fleet/concepts-choosing-fleet
             dnsPrefix: dnsPrefix ?? undefined,
         };
         const resource = {
@@ -162,6 +164,12 @@ async function createFleet(
     try {
         const result = await client.fleets.beginCreateOrUpdateAndWait(resourceGroupName, name, resource);
         if (!result || !result.id) {
+            window.showWarningMessage(
+                `Fleet creation failed: No ID returned. 
+                Resource Group Name: ${resourceGroupName},
+                Fleet Name: ${name}, 
+                Location: ${resource.location}.`,
+            );
             throw new Error(
                 `Fleet creation failed: No ID returned. 
                 Resource Group Name: ${resourceGroupName},
--- a/src/panels/CreateFleetPanel.ts
+++ b/src/panels/CreateFleetPanel.ts
@@ -3,6 +3,7 @@ import { BasePanel, PanelDataProvider } from "./BasePanel";
 import { Uri, window } from "vscode";
 import { MessageHandler, MessageSink } from "../webview-contract/messaging";
 import {
+    HubMode,
     InitialState,
     ProgressEventType,
     ToVsCodeMsgDef,
@@ -64,7 +65,14 @@ export class CreateFleetDataProvider implements PanelDataProvider<"createFleet">
             getLocationsRequest: () => this.handleGetLocationsRequest(webview),
             getResourceGroupsRequest: () => this.handleGetResourceGroupsRequest(webview),
             createFleetRequest: (args) =>
-                this.handleCreateFleetRequest(args.resourceGroupName, args.location, args.name, webview),
+                this.handleCreateFleetRequest(
+                    args.resourceGroupName,
+                    args.location,
+                    args.name,
+                    args.hubMode,
+                    args.dnsPrefix,
+                    webview,
+                ),
         };
     }
 
@@ -115,10 +123,18 @@ export class CreateFleetDataProvider implements PanelDataProvider<"createFleet">
         resourceGroupName: string,
         location: string,
         name: string,
+        hubMode: HubMode,
+        dnsPrefix: string | undefined,
         webview: MessageSink<ToWebViewMsgDef>,
     ) {
+        const hubProfile = {
+            dnsPrefix: dnsPrefix,
+        };
         const resource = {
-            location: location,
+            // Fleet does not support the full name of the location at this moment
+            // Change "location" to lowercase and remove spaces to match the required format
+            location: location.toLowerCase().replaceAll(" ", ""),
+            hubProfile: hubMode === HubMode.With ? hubProfile : undefined,
         };
 
         await createFleet(this.fleetClient, resourceGroupName, name, resource, webview);
@@ -144,17 +160,14 @@ async function createFleet(
     const environment = getEnvironment();
     try {
         const result = await client.fleets.beginCreateOrUpdateAndWait(resourceGroupName, name, resource);
-        if (!result.id) {
-            throw new Error("Fleet creation did not return an ID");
-        }
-        const deploymentPortalUrl = getDeploymentPortalUrl(environment, result.id);
+        const deploymentPortalUrl = getDeploymentPortalUrl(environment, result.id!);
         webview.postProgressUpdate({
             event: ProgressEventType.Success,
             operationDescription,
             errorMessage: null,
             deploymentPortalUrl,
             createdFleet: {
-                portalUrl: getPortalResourceUrl(environment, result.id),
+                portalUrl: getPortalResourceUrl(environment, result.id!),
             },
         });
     } catch (error) {
--- a/src/panels/CreateFleetPanel.ts
+++ b/src/panels/CreateFleetPanel.ts
@@ -124,11 +124,12 @@ export class CreateFleetDataProvider implements PanelDataProvider<"createFleet">
         location: string,
         name: string,
         hubMode: HubMode,
-        dnsPrefix: string | undefined,
+        dnsPrefix: string | null,
         webview: MessageSink<ToWebViewMsgDef>,
     ) {
         const hubProfile = {
-            dnsPrefix: dnsPrefix,
+            // If HubClusterMode is "With", use the given dnsPrefix; Otherwise set it to undefine
+            dnsPrefix: dnsPrefix ?? undefined,
         };
         const resource = {
             // Fleet does not support the full name of the location at this moment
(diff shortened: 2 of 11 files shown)


src/panels/CreateFleetPanel.ts
```diff
@@ -144,17 +160,14 @@ async function createFleet(
     const environment = getEnvironment();
     try {
         const result = await client.fleets.beginCreateOrUpdateAndWait(resourceGroupName, name, resource);
-        if (!result.id) {
```
ReinierCC, line 147: How come this check was removed?

webview-ui/src/manualTest/createFleetTests.tsx
```diff
@@ -0,0 +1,120 @@
...
+        const deploymentPortalUrl = `https://portal.azure.com/#resource/subscriptions/${initialState.subscriptionId}/resourceGroups/${groupName}/providers/Microsoft.Resources/deployments/testdeployment?referrer_source=vscode&referrer_context=vscode-aks-tools-test`;
+        webview.postProgressUpdate({
+            operationDescription: `Creating Fleet ${name}`,
+            event: ProgressEventType.InProgress,
+            errorMessage: null,
+            deploymentPortalUrl,
+            createdFleet: null,
+        });
+
+        const waitMs = location === failLocationMarker ? 500 : location === cancelLocationMarker ? 3000 : 10000;
```
ReinierCC, line 85: kint: Doing this check twice, could you do the "location === cancelLocationMarker" check beforehand so it's easier to read.

JunyuQian: Yep I'll change it to a switch statement and check cancelLocationMarker first to improve readability :)

webview-ui/src/CreateFleet/CreateFleetInput.tsx
```diff
@@ -83,16 +88,28 @@ export function CreateFleetInput(props: CreateFleetInputProps) {
...
         if (!isValid(fleetName)) return nothing();
+        // dnsPrefix validation is only required when hubMode is set to "With"
+        if (hubModeSelected === HubMode.With && (!isValid(dnsPrefix) || dnsPrefix.value === "")) return nothing();
 
         const parameters: CreateFleetParams = {
             resourceGroupName,
             location: location.value,
             name: fleetName.value,
+            hubMode: hubModeSelected,
+            dnsPrefix: hubModeSelected === HubMode.With && isValid(dnsPrefix) ? dnsPrefix.value : undefined,
```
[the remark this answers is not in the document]
JunyuQian: Yeah actually that's something I'm not quite sure how to do.

As the screenshot below shows, I've ensured the validity of the `dnsPrefix` variable when `hubMode` is set to `With`. However, the compiler keeps indicating that `dnsPrefix` may be unset, which is why I had to add duplicated checks.

If there is any better approach, please do let me know!

src/panels/CreateFleetPanel.ts
```diff
@@ -144,17 +160,14 @@ async function createFleet(
...
-            throw new Error("Fleet creation did not return an ID");
-        }
-        const deploymentPortalUrl = getDeploymentPortalUrl(environment, result.id);
+        const deploymentPortalUrl = getDeploymentPortalUrl(environment, result.id!);
```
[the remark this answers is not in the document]
JunyuQian: Well for most of the time... Cuz the API is developed by Microsoft 😂

But since it is not good to fully trust the API, I'll add the check back with more detailed error message

webview-ui/src/CreateFleet/CreateFleetInput.tsx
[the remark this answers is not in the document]
dvadas: I think you can get rid of the `hubModeSelected === HubMode.With` entirely, and then just call one of the existing functions on Validatable, e.g. `toNullable()`. You'll have to change the type from `string | undefined` to `string | null` but that should be fine.

webview-ui/src/CreateFleet/CreateFleetModeInput.tsx
```diff
@@ -0,0 +1,100 @@
...
+    onModeSelected: (modeSelected: HubMode) => void;
+    onDnsPrefixChange: (dnsPrefix: Validatable<string>) => void;
+}
+
+export function CreateFleetModeInput(props: CreateFleetInputProps) {
+    function handleModeClick(modeSelected: HubMode) {
+        props.onModeSelected(modeSelected);
+    }
+
+    function handlednsPrefixChange(e: ChangeEvent) {
```
dvadas, line 25: nit: Should capitalise the `d` in `dns`.

dvadas (approved):

webview-ui/src/CreateFleet/CreateFleet.tsx
```diff
@@ -14,7 +14,8 @@ export function CreateFleet(initialState: InitialState) {
             vscode.postGetResourceGroupsRequest();
             eventHandlers.onSetInitializing(); // Set stage to Stage.Loading
         }
-    });
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
Tatsinnit, line 17: 💡 Thoughts/question: What are we disabling linting for this?

src/panels/CreateFleetPanel.ts
```diff
@@ -115,10 +123,19 @@ export class CreateFleetDataProvider implements PanelDataProvider<"createFleet">
         resourceGroupName: string,
         location: string,
         name: string,
+        hubMode: HubMode,
+        dnsPrefix: string | null,
         webview: MessageSink<ToWebViewMsgDef>,
     ) {
+        const hubProfile = {
+            // If HubClusterMode is "With", use the given dnsPrefix; Otherwise set it to undefine
```
Tatsinnit, line 131: 💡 Idea: Please add any doc link if this needs further reference.

src/panels/CreateFleetPanel.ts
```diff
@@ -144,8 +161,13 @@ async function createFleet(
     const environment = getEnvironment();
     try {
         const result = await client.fleets.beginCreateOrUpdateAndWait(resourceGroupName, name, resource);
-        if (!result.id) {
-            throw new Error("Fleet creation did not return an ID");
+        if (!result || !result.id) {
+            throw new Error(
```
Tatsinnit, line 165: 💡 Shall we add this jsut before throw, aka adding `vscode.window.showWarningMessage(Full error here as well)` line above.

src/webview-contract/webviewDefinitions/createFleet.ts
```diff
@@ -25,6 +25,13 @@ export interface CreateFleetParams {
     resourceGroupName: string;
     location: string;
     name: string;
+    hubMode: HubMode;
+    dnsPrefix: string | null; // required, if hubMode is "With"
+}
+
+export enum HubMode {
```
Tatsinnit, line 32: 💡Thoughts: can we please add some msdn link for relevance of this as doc comment.

Tatsinnit (commented):
❤️ Looks good to me, by adding the full `package.json` changes will mean its a full release which means please add the final `vsix` so that @tejhan , @ReinierCC can play around for a bit. Also fyi : @qpetraroia as fyi. 

I would also suggest and express that I am fine with a silent release of this feature if we have covered the use-case will and handled baseline cases. 🥷 

Thank you so much again!

JunyuQian:
> ❤️ Looks good to me, by adding the full `package.json` changes will mean its a full release which means please add the final `vsix` so that @tejhan , @ReinierCC can play around for a bit. Also fyi : @qpetraroia as fyi.
> 
> I would also suggest and express that I am fine with a silent release of this feature if we have covered the use-case will and handled baseline cases. 🥷
> 
> Thank you so much again!

Thank you very much Tats! Here is the VSIX file:
[vscode-aks-tools-1.5.5-vsix-edition.vsix.zip](https://github.com/user-attachments/files/18583555/vscode-aks-tools-1.5.5-vsix-edition.vsix.zip)

tejhan (approved):
Reviewed code & tested out .vsix - LGMT & everything worked perfectly on my end👍.

webview-ui/src/CreateFleet/CreateFleet.tsx
JunyuQian: Ah, this is related to [this discussion](https://github.com/Azure/vscode-aks-tools/pull/1179#discussion_r1921675231). Since we know the state transition from `Initializing` to `Loading` only happens once, we can improve the program efficiency by disabling the listening for this `useEffect` function only, so that it will not check the if-statement every time the page is rendered.

src/webview-contract/webviewDefinitions/createFleet.ts
JunyuQian: Yeah, but for this one, this rule is learned from the Azure portal. I'm not particularly sure about the doc that records the rule. Or actually, I can leave a comment like "according to the Azure portal's Fleet creation procedure,..."

webview-ui/src/CreateFleet/CreateFleet.tsx
Tatsinnit: Cool, thanks for this and we should revisit this as it exist here as well :) https://github.com/Azure/vscode-aks-tools/blob/main/webview-ui/src/CreateCluster/CreateClusterPresetInput.tsx#L16 fyi @tejhan and @hsubramanianaks  ❤️ 

Thanks all

package.json
```diff
@@ -331,6 +331,10 @@
             {
                 "command": "aks.aksKaitoTest",
                 "title": "Test KAITO models"
+            },
+            {
+                "command": "aks.aksCreateFleet",
```
Tatsinnit, line 336: 💡 ❤️  Thank you so much Junyu  for the ping and I will take a deeper look latter as I am little busy, but I have key observation to share here:

* Yesterday you share a new vision of some treeNode changes, and that latter kept me wondering where this command will be hooked then? Is this the right place or do we need to rethink this as well.

If the plan is to make new Fleet --> Sub-folder with details aka FleetManager --> Subs + fleetCluster relation, then will it make sense to hook this createFleet command on top of that "Fleet" subfolder (level)?

serbrech: We discussed this today.
there are other drawbacks to move the fleet out of the subscription tree.
the one you mention is one of them.

The trade-offs become clearer when you spike the solution and try it out.
It's hard to realize them ahead of time.

We'll keep the flee item under the subscription, like this spike PR does #1211 

- the clusters that are in a fleet move under the fleet node (to avoid the confusion of duplicates in the UI).
the clusters that are in in other subs can be shown under the fleet as well (not done in this PR)

We discussed how this can be done today, andit will work without adding more requests to load the data.

We can have a chat next week on this, but it seems like a good balance between the ability of the current UI, the integration with the cluster treenode functionality, and the complexity.

I do have some recommendation on some abstractions that currently leak to the commands implementation.
searching through the code, only the devhub crosses that line by looking at the cluster location through the clusterTreenode.
All other command strictly rely on subId, RG and name of the resource to do their work.

serbrech (approved):
🚀

package.json
Tatsinnit: ❤️ thanks for this— sounds great! 🎉 I believe the create-fleet is correctly hooked in this case then, and at the very least, this PR aligns well with the goal.

Regarding the separate Fleet folder for NodeTree (which I really like!), I'd love for us to dive deeper into the best way to approach it. And of course, count me in for helping implement the concept—no matter how challenging, I’m all in for reaching the best outcome! 😃

It might need some initial brainstorming, but together, we can refine it into something even more optimal.

Appreciate it ! 🙌✨FYI and CC:  @tejhan, @ReinierCC , @qpetraroia , @hsubramanianaks as we need to revise the node tree design revisit sometime. (so we keep this in mind please for that enhancement)

serbrech: Can we keep a tree redesign separate from the scope of core fleet support?

as I explained, and as the multiple draft PR show, we have experimented with how to load the fleet elements.
this is the least intrusive and works well. Moving the fleets to a separate section, IMO, would require major refactoring and design changes to avoid a large amount of additional API calls.

Tatsinnit: ❤️ Oh, absolutely 100 % —let’s not dive into major NodeTree design changes, and I’d definitely prefer to avoid any major breaking behavior changes as minor work. 😊

As for this PR, I’ll take a final look and get it over the line. I do recall some discussion last Friday, so just want to double-check—are we all set for this PR, or is there any talk about moving the command to another part of the tree, like a "Fleet" subdirectory? Let me know! 🚀✨ But I think that is not the case = )

Tatsinnit (approved):
❤️🙏 thank you so much!! Great discussions above between us so let’s keep them for archival and next in drop reasons, let’s get this in!!

thank you all!

PR #1208, opened 2025-01-29
Feature: Add Cluster Filtering in AKS VS Code Extension Tree Node

Tatsinnit:
#### **Description:**  

Enhance the **AKS (Azure Kubernetes Service) VS Code extension** by introducing a **cluster filtering** feature within the tree view. This will allow users to **easily locate and manage specific clusters** based on predefined filters, improving usability and efficiency.  

#### **Key Features:**  

✅ **Search & Filter:** Users can apply filters based on cluster name.
✅ **Persistent Filtering:** Selected filters remain active across sessions for a consistent user experience.  
✅ **Dynamic Updates:** The tree node updates in real-time as filters are applied or removed.  
✅ **Multi-Filter Support:** Users can combine multiple filter criteria to refine the cluster list further.  
✅ **Clear & Reset Option:** A quick way to reset filters and display all available clusters.  

#### **User Benefits:**  

*  **Improved Navigation** – Quickly find relevant clusters without scrolling through long lists.  
*  **Enhanced Productivity** – Focus on specific clusters based on user-defined criteria.  
*  **Seamless Integration** – Works natively within the VS Code extension tree node, ensuring a familiar experience.  🚀

Thank you for bringing this idea and humble fyi to @qpetraroia , @ReinierCC , @tejhan , @hsubramanianaks ❤️ 

## VSIX to test:

[vscode-aks-tools-1.5.5-clusterFilterTest1.vsix.zip](https://github.com/user-attachments/files/18600008/vscode-aks-tools-1.5.5-clusterFilterTest1.vsix.zip)


ReinierCC (dismissed):
Looks Good! Cool Feature!

src/commands/utils/clusterfilter.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (failed(sessionProvider)) {
+        vscode.window.showErrorMessage(sessionProvider.error);
+        return;
+    }
+
+    const containerServiceClient = getAksClient(sessionProvider.result, subscriptionNode.result.subscriptionId);
+    const clusterList: ManagedCluster[] = [];
+
+    await longRunning(`Getting AKS Cluster list for ${subscriptionNode.result.name}`, async () => {
+        const iterator = containerServiceClient.managedClusters.list();
```
hsubramanianaks, line 39: I think we have another method in clusters.ts --> getClusters which does similar thing but uses resource management client instead of containerservice client. we should have follow up item to merge them both and clean. May be as part of this story can you move this to src/command/utils/clusters.ts file and then I can do a refactor to have one method which would impact cluster Filters and github copilot. Please let me know your thoughts. Thank you.

Tatsinnit: Indeed, and this is the cause of tiny delay for once and once its cached, its much better,

src/commands/utils/clusterfilter.ts
```diff
@@ -0,0 +1,95 @@
...
+    // Set Cluster Instance
+    const newFilteredClusters = [
+        ...selectedItems.map((item) => item.Cluster), // Retain filters in any other tenants.
+    ];
+
+    await setFilteredClusters(newFilteredClusters);
+}
+
+async function getUniqueClusters() {
+    const filteredClusters = await getFilteredClusters();
```
hsubramanianaks, line 83: await has no effect here since getFilteredClusters is not async method. can we correct it or you meant getFilteredClusters to be async?

hsubramanianaks, line 83: await has no effect here since getFilteredClusters is not async method. can we correct it or you meant getFilteredClusters to be async?

src/commands/utils/clusterfilter.ts
```diff
@@ -0,0 +1,95 @@
...
+    // Set Cluster Instance
+    const newFilteredClusters = [
+        ...selectedItems.map((item) => item.Cluster), // Retain filters in any other tenants.
+    ];
+
+    await setFilteredClusters(newFilteredClusters);
+}
+
+async function getUniqueClusters() {
+    const filteredClusters = getFilteredClusters();
```
hsubramanianaks, line 83: Initially this function confused me that container service client was listing duplicates but then realized this is list of clusters that are already selected and stored in the user settings config. can we rename the function ex: getSelectedClustersFromConfig ? or give a description to the function method with description of what it does Please ?

Tatsinnit: ok, lets chat : `getFileterestCluster` means all the cluster which are filtered? `getSelectedClustersFromConfig` is more kind of developer level lingo, is what I think, I am open to thoughts. Lets chat. Thanks.

src/commands/utils/config.ts
```diff
@@ -82,6 +114,43 @@ export async function setFilteredSubscriptions(filters: SubscriptionFilter[]): P
     }
 }
 
+export async function setFilteredClusters(filters: ClusterFilter[]): Promise<void> {
```
hsubramanianaks, line 117: method name suggestion: setSelectedClustersToConfig ? :)

Tatsinnit: 💡 Actually this breaks the contract of current filter name: see how existing `getFilteredSubscription` is named. like I said name is not issue, but lets chat please. thanks

ReinierCC (approved):

src/commands/utils/clusterfilter.ts
hsubramanianaks: Agreed for consistency purposes. we can refactor the method names later. Thank you.

src/commands/utils/config.ts
hsubramanianaks: same as above, agreed on the consistency with existing methods.

src/commands/utils/clusterfilter.ts
hsubramanianaks: we can debug this to see what calls are made to backend and which api takes longer, but that doesn't have to do anything with this PR. this section can be revisited later for moving into utils method. Thanks Tats.

hsubramanianaks (approved):

PR #1211, opened 2025-01-30
DRAFT: nest members under fleet

serbrech:
spike to nest member clusters under the fleet object

issue:
- this only nests the clusters from the same sub as thefleet

screenshot:
image

outcome/suggestion:

- ~~Move the fleet items outside of the subscription, under a single node called "Fleets"~~
- ~~Keep the clusters as-is under the subscriptions~~
- List the members under each fleet, and make the node behave the same as the clusters under subscriptions.
This does not require the full cluster object, the FleetMember has the clusterResourceId and that's enough to enable all context menu item of a cluster tree node. (tested in #1212)


src/tree/subscriptionTreeItem.ts
```diff
@@ -76,16 +84,79 @@ class SubscriptionTreeItem extends AzExtParentTreeItem implements SubscriptionTr
...
+                const fleetTreeItem = createFleetTreeNode(this, this.subscriptionId, r);
+                fleetTreeNodes.set(r.id, fleetTreeItem);
+                return fleetTreeItem;
+            } else if (r.type === "Microsoft.ContainerService/managedClusters") {
+                const member = clusterToMemberMap.get(r.id);
+                if (member) {
+                    const fleet = fleetTreeNodes.get(member.parentResourceId);
+                    fleet?.addCluster([r]);
+                } else {
+                    const cluster = createClusterTreeNode(this, this.subscriptionId, r);
```
serbrech, line 147: only the clusters that are NOT members of a fleet should be added to the tree

src/tree/subscriptionTreeItem.ts
```diff
@@ -76,16 +84,79 @@ class SubscriptionTreeItem extends AzExtParentTreeItem implements SubscriptionTr
...
+        allResources.forEach((r) => {
+            if (r.type === "Microsoft.ContainerService/fleets") {
+                const fleetTreeItem = createFleetTreeNode(this, this.subscriptionId, r);
+                fleetTreeNodes.set(r.id, fleetTreeItem);
+                return fleetTreeItem;
+            } else if (r.type === "Microsoft.ContainerService/managedClusters") {
+                const member = clusterToMemberMap.get(r.id);
+                if (member) {
+                    const fleet = fleetTreeNodes.get(member.parentResourceId);
+                    fleet?.addCluster([r]);
```
serbrech, line 145: we are only adding clusters from current subscription as member to the fleet

we should add all members to the fleet, even if they aren't part of this subscription

PR #1212, opened 2025-01-31
Other way to load members under fleet

serbrech:
- Load fleet and clusters under subscriptions.
- Load the members on expansion of the fleet
    - use ListMember to construct the aksResourceTreeItem under fleet, but only via the member resource info.
    - this allows to show members from other subscriptions
    
known gap: 
does not remove the clusters from under the subscription when they are part of the fleet (TODO?)
Could make sense for clusters in the same sub as the fleet, and be a good user experience trade-off.

image


src/azure-api-utils.ts
```diff
@@ -1,7 +1,15 @@
-export function parseResource(armId: string): { resourceGroupName: string | undefined; name: string | undefined } {
+export function parseResource(armId: string): {
```
serbrech, line 1: needs unit tests

src/commands/utils/azureResources.ts
```diff
@@ -25,6 +35,11 @@ export type DefinedResourceWithGroup = DefinedResource & {
     resourceGroup: string;
 };
 
+export type DefinedFleetMemberWithGroup = DefinedResourceWithGroup & {
```
serbrech, line 38: not in use anymore. might not be needede at all.

src/commands/utils/azureResources.ts
```diff
@@ -46,3 +61,23 @@ function asResourceWithGroup(resource: DefinedResource): DefinedResourceWithGrou
 function getResourceGroup(resource: DefinedResource): string {
     return parseResource(resource.id).resourceGroupName!;
 }
+
+function asFleetMemberWithGroup(member: FleetMember): DefinedFleetMemberWithGroup {
```
serbrech, line 65: related to DefinedFleetMemberWithGroup, might go if the type goes away

src/commands/utils/azureResources.ts
```diff
@@ -46,3 +61,23 @@ function asResourceWithGroup(resource: DefinedResource): DefinedResourceWithGrou
...
+        name: member.name!,
+        clusterResourceId: member.clusterResourceId!,
+        parentResourceId: parseResource(member.id!).parentResourceId!,
+    };
+}
+
+export async function getFleetMembers(
+    sessionProvider: ReadyAzureSessionProvider,
+    fleet: DefinedResourceWithGroup,
+): Promise<Errorable<DefinedFleetMemberWithGroup[]>> {
```
serbrech, line 78: can we avoid this type?

src/tree/fleetTreeItem.ts
```diff
@@ -0,0 +1,98 @@
...
+    public readonly armId: string;
+    public readonly resourceGroupName: string;
+    public readonly name: string;
+    private readonly fleetResource: DefinedResourceWithGroup;
+
+    constructor(
+        parent: AzExtParentTreeItem & SubscriptionTreeNode,
+        sessionProvider: ReadyAzureSessionProvider,
+        readonly subscriptionId: string,
+        readonly fleet: DefinedResourceWithGroup,
```
serbrech, line 45: should not be marked readonly. only internal.
probably, read about how class constructor work in typespec :)

src/tree/subscriptionTreeItem.ts
```diff
@@ -76,16 +77,41 @@ class SubscriptionTreeItem extends AzExtParentTreeItem implements SubscriptionTr
...
+            );
+            return [];
+        }
+
+        const allResources = fleetResources.result.concat(clusterResources.result);
+
+        const allTreeItems = allResources.map((r) => {
+            if (r.type?.toLowerCase() === "microsoft.containerservice/fleets") {
+                const fleetTreeItem = createFleetTreeNode(this, this.sessionProvider, this.subscriptionId, r);
+                return fleetTreeItem as unknown as AzExtTreeItem;
```
serbrech, line 105: any way to avoid this??

Issue #1207, opened 2025-01-27
Deprecation Notice: Migrate from webview-ui-toolkit to Maintained UI Packages


JunyuQian:
### Issue

The project uses many UI elements (such as `VSCodeButton` and `VSCodeDropdown`) from the [webview-ui-toolkit](https://github.com/microsoft/vscode-webview-ui-toolkit) package. However, the original repo has been archived on **January 7, 2025**. According to the announcement from the repo administrator, this toolkit has been deprecated since the end of 2024.

### Alternative Packages

As the administrator also mentioned, consider using the [VS Code Elements](https://github.com/vscode-elements/elements) package or the [VS Code React UI](https://github.com/estruyf/vscrui) package, which are under maintenance. Original announcement: [vscode-webview-ui-toolkit Announcement](https://github.com/microsoft/vscode-webview-ui-toolkit/issues/561)

Thank you @serbrech and @Tatsinnit for the discussion last week :)

tejhan:
Closing this issue as webview-ui-toolkit has now been fully deprecated.

PR #1222, opened 2025-02-05
Beginning migration away from @vscode/webview-ui-toolkit/react package. 

tejhan:
This PR serves as a start point for the eventual migration away from the @vscode/webview-ui-toolkit/react package.

**Background** 

As mentioned in issue #1207 , the @vscode/webview-ui-toolkit has recently been deprecated & will no longer be maintained. 

There were 2 popular alternatives specializing in vscode styled components, but upon testing them, both the <a href= https://github.com/vscode-elements/elements >Vscode Elements Package</a> and the <a href= https://github.com/estruyf/vscruis >Vscode React UI Package</a> have peer dependencies of react version ^18 || ^17. Neither have been updated to work with v19 as of yet, and it doesn't seem that there are any plans to do so in the immediate future. 
We have a couple packages relying on v19 of react, most notably the fontawesome icons used largely throughout the app alongside a couple other packages like react-dom & react-markdown. 

When looking at our usage of the toolkit in the extension, it's all primarily wrapped up in about 6-7 components, such as the VSCodeButton, VScodetextfield, VsCodeDivider, etc.. All of which use vscodes' built in style variables. 

Because of the way they are implemented in the react packages, & the fact that the version we use isn't compatible, my approach I'm suggesting here is to just utilize our already existing main.css file to standardize elements across the extension with the built in vscode variables to provide the exact same behavior we already have + additional customization without relying on packages. In this case, the packages we use do the same vs-code variable assignment as this PR.

**Changes**
- Created custom styling rules to universally use in the application.
- Changed buttons, horizontal rulings, anchor tags, & inputs to mimic our elements from the toolkit.
- Also added classes to mimic other things such as secondary buttons.

**TBD**
- If this looks good, we can just go ahead and replace most of the elements in the extension seamlessly in a follow up PR.
- For now I've just swapped a couple elements in the KAITO ecosystem, & have left a couple elements such as VSCodeProgressRing open to recreate in a future PR as well once we have a decision.

.vsix for testing
[vscode-aks-tools-1.5.5-custom-ui-test.vsix.zip](https://github.com/user-attachments/files/18679312/vscode-aks-tools-1.5.5-custom-ui-test.vsix.zip)

(For anyone testing, I'd just advise taking a look at a few random pages to see if styles of any elements were messed up (which they shouldn't), and look at some of the Kaito pages to see the replacements (they should appear visually identical to their toolkit alternatives))

### DIFF
--- a/webview-ui/src/Kaito/Kaito.tsx
+++ b/webview-ui/src/Kaito/Kaito.tsx
@@ -1,4 +1,4 @@
-import { VSCodeButton, VSCodeProgressRing, VSCodeDivider } from "@vscode/webview-ui-toolkit/react";
+import { VSCodeProgressRing } from "@vscode/webview-ui-toolkit/react";
 import { InitialState, ProgressEventType } from "../../../src/webview-contract/webviewDefinitions/kaito";
 import { useStateManagement } from "../utilities/state";
 import styles from "./Kaito.module.css";
@@ -27,7 +27,7 @@ export function Kaito(initialState: InitialState) {
         <>
             <div className={styles.container}>
                 <h2>Kubernetes AI Toolchain Operator (KAITO) - {state.clusterName}</h2>
-                <VSCodeDivider />
+                <hr />
                 <div className={styles.subHeader}>
                     Using KAITO, the workflow of onboarding and deploying large AI inference models on your cluster is
                     largely simplified. KAITO manages large model files using container images and hosts them in the
@@ -105,9 +105,9 @@ export function Kaito(initialState: InitialState) {
                             <p className={styles.thin}>You can now create a workspace by clicking the button below.</p>
                             <div>
                                 {" "}
-                                <VSCodeButton className={styles.generateButton} onClick={onClickGenerateWorkspace}>
+                                <button className={styles.generateButton} onClick={onClickGenerateWorkspace}>
                                     Generate Workspace
-                                </VSCodeButton>
+                                </button>
                             </div>
                         </div>
                     )}
--- a/webview-ui/src/main.css
+++ b/webview-ui/src/main.css
@@ -16,10 +16,13 @@ body {
 a,
 a code {
     color: var(--vscode-textLink-foreground);
+    font-size: var(--vscode-font-size);
+    text-decoration: none;
 }
 
 a:hover {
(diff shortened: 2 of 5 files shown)


Tatsinnit:
❤️ II like these changes, thank you so much for details attached. I will gentle fyi @qpetraroia for the kind help for the folks in VSCE team for some cold here for futuristic approach we might want to approach with.

Regarding this PR,I would love it if you want to do something incremental and this is part-1 of that process and non-breaking?

Thanks

Tatsinnit (approved):
❤️❤️ Thank you so much for this, worth to share VSIX as well but looks well contained changes. Thank you!

tejhan:
> ❤️❤️ Thank you so much for this, worth to share VSIX as well but looks well contained changes. Thank you!

VSIX is attached in description, but putting here as well for visibility: 
[vscode-aks-tools-1.5.5-custom-ui-test.vsix.zip](https://github.com/user-attachments/files/18694904/vscode-aks-tools-1.5.5-custom-ui-test.vsix.zip)

webview-ui/src/Kaito/Kaito.tsx
```diff
@@ -27,7 +27,7 @@ export function Kaito(initialState: InitialState) {
         <>
             <div className={styles.container}>
                 <h2>Kubernetes AI Toolchain Operator (KAITO) - {state.clusterName}</h2>
-                <VSCodeDivider />
+                <hr />
```
serbrech, line 30: I think that there is something to learn here with regards to (not) using react components to replace a built-in html "hr" element... :)

webview-ui/src/main.css
```diff
@@ -28,9 +31,56 @@ select:focus,
 textarea:focus {
     outline: -webkit-focus-ring-color solid 1px;
     outline-offset: -1px;
+    outline-color: var(--vscode-focusBorder);
+}
+
+button {
```
hsubramanianaks, line 37: **For future sake:** we should move these css to utilities and reuse this button in all other pages to be more consistent. Not in this PR, something to think about it in upcoming PR's. Thank you.

hsubramanianaks (approved):
LGTM, tested as well. I would recommend testing this in different platforms and vscode themes to make sure it is working as expected.

PR #1228, opened 2025-02-10
Proper Devhub Callback Auth

ReinierCC:
To leverage DevHub's callback abilities I registered callback URI inside the extension.

This is a unique uri that only the vscode extension can utilize: "vscode://ms-kubernetes-tools.vscode-aks-tools/callback".

Once the user provides goes through OATH, DevHub will redirect back to the VScode extension alerting the user that the authentication was successful or it will timeout.

Currently, there's a small edge case in where if the user revokes authorization to DevHub and quickly tries to login in again they are met with an invalid state error. 
' {"error":{"code":"400", "message":"invalid state parameter", "details":[]}} '

I believe this issue might be from the DevHub side temporarily caching the state, but the state has changed.

Still the Automated Deployments command is not available to the user but, the following vsix allows you to test the login functionality: 
[vscode-aks-tools-1.5.5-callback-test.vsix.zip](https://github.com/user-attachments/files/18740137/vscode-aks-tools-1.5.5-callback-test.vsix.zip)
Under "Automated Deployments" -> "Deploy an App with DevHub"

Clicking on the command will initiate the auth flow.

### DIFF
--- a/src/commands/devhub/aksAutomatedDeployments.ts
+++ b/src/commands/devhub/aksAutomatedDeployments.ts
@@ -111,7 +111,7 @@ async function checkAndAuthenticateWithGitHub(
 ): Promise<string | undefined> {
     try {
         const parameters: GitHubOAuthCallRequest = {
-            redirectUrl: "vscode://ms-kubernetes-tools.vscode-aks-tools/callback",
+            redirectUrl: "vscode://ms-kubernetes-tools.vscode-aks-tools/callback", //After GitHub redirects to DevHub with token, DevHub will redirect the user back to the extension. Causing the flow to continue. 
         };
         const properOptionsWithCallback: GitHubOAuthOptionalParams = { parameters };
         client.listGitHubOAuth(location, properOptionsWithCallback);
@@ -130,7 +130,7 @@ async function checkAndAuthenticateWithGitHub(
             try {
                 await onCallbackHandled;
             } catch (error) {
-                vscode.window.showErrorMessage(`Failed to handle callback: ${getErrorMessage(error)}`);
+                vscode.window.showErrorMessage(`Failed to handle GitHub Auth callback: ${getErrorMessage(error)}`);
                 console.error("Failed to handle callback:", getErrorMessage(error));
             }
 
--- a/src/commands/devhub/aksAutomatedDeployments.ts
+++ b/src/commands/devhub/aksAutomatedDeployments.ts
@@ -1,6 +1,6 @@
 import { getReadySessionProvider, getCredential } from "../../auth/azureAuth";
 import { IActionContext } from "@microsoft/vscode-azext-utils";
-import { failed } from "../utils/errorable";
+import { failed, getErrorMessage } from "../utils/errorable";
 import * as k8s from "vscode-kubernetes-tools-api";
 import * as vscode from "vscode";
 import { getAksClusterTreeNode, getKubernetesClusterInfo } from "../utils/clusters";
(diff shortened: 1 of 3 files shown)


src/commands/devhub/aksAutomatedDeployments.ts
```diff
@@ -115,12 +108,15 @@ export default async function aksAutomatedDeployments(_context: IActionContext,
 async function checkAndAuthenticateWithGitHub(
     client: DeveloperHubServiceClient,
     location: string,
-    options: GitHubOAuthOptionalParams,
-    timeout: number = 300000,
-    interval: number = 5000,
 ): Promise<string | undefined> {
     try {
-        const gitHubOAuthResp = await client.gitHubOAuth(location, options); // Call to authenticate with GitHub
+        const parameters: GitHubOAuthCallRequest = {
```
Tatsinnit, line 113: 💡 Suggestion: Can we please add some comment for the behaviour for this `redirectUrl` and the edge case this will take care please?

src/commands/devhub/aksAutomatedDeployments.ts
```diff
@@ -131,46 +127,26 @@ async function checkAndAuthenticateWithGitHub(
                 "GitHub authentication initiated. Please complete the process in your browser.",
             );
 
-            // Poll for token availability
-            const token = await pollForToken(client, location, options, timeout, interval);
+            try {
+                await onCallbackHandled;
+            } catch (error) {
+                vscode.window.showErrorMessage(`Failed to handle callback: ${getErrorMessage(error)}`);
```
Tatsinnit, line 133: 💡 Idea/Suggesstion: I wonder if this will be GitHub Auth failure if we can add more explicit like `Failed to handle GitHub Auth callback` or something closer, what do you think?

src/commands/devhub/aksAutomatedDeployments.ts
```diff
@@ -131,46 +127,26 @@ async function checkAndAuthenticateWithGitHub(
...
     } catch (error) {
-        console.error("Error occurred during GitHub OAuth:", error);
+        console.error("Error occurred during DevHub GitHub OAuth:", error);
+        vscode.window.showErrorMessage(`Error occurred during DevHub GitHub OAuth: ${getErrorMessage(error)}`);
         return undefined;
     }
 }
-
-// Utility function to poll for a token
-async function pollForToken(
```
Tatsinnit, line 151: 💡 Why is this function no longer needed? Reason I ask is that if I see this function is seemingly with intent to typically use in OAuth authentication flows where the API takes time to generate and return a token? I wonder if we need this?

Just sharing as thought clarity and  idea?

Tatsinnit (approved):
❤️ Thank you so for this work, I've left few comments for archival reasons and thoughts. Please take a look, rest looks really good to me. 🎊

tejhan (dismissed):
Tested out auth flow & it works well, LGTM👍

src/commands/devhub/aksAutomatedDeployments.ts
ReinierCC: Good point to clarify. This previous function would keep polling and dummy waiting which was inefficient because it wouldn't leverage the callback functionality that the OAuth flow has. Originally was done this way because I thought we would need to run a small server on the vscode side to receive the callback but after finding out vscode can natively support the callback through the unique link opted for this version.
 
In the future if we have any new Oath flows, it'll be best to handle the callback natively if possible.

src/commands/devhub/aksAutomatedDeployments.ts
ReinierCC: Good suggestion, I'll make the change

tejhan (approved):

Tatsinnit (approved):

src/extension.ts
```diff
@@ -63,6 +64,8 @@ export async function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(new Reporter());
     setAssetContext(context);
 
+    registerUriHandler(context);
```
hsubramanianaks, line 67: Can you please leave some documentation on how this registerUriHandler works? if official docs are present ?  I found some in the vscode official docs is this the same ? 

https://code.visualstudio.com/api/references/vscode-api#UriHandler
image

hsubramanianaks (approved):
lgtm, just a minor comment. Thanks @ReinierCC

PR #1220, opened 2025-02-05
(Outdated) Showing Fleets in the Treeview

JunyuQian:
# Showing Fleets in the Treeview

This PR adds the Fleet resource to the treeview, based on @serbrech's initial work. The image below shows the final result of the treeview.
treeview

Due to the special characteristics that a Fleet can have member clusters from different subscriptions, the initial treeview design, which classifies clusters based on subscriptions, does not quite fit Fleets. The current design can be summarized into three cases below:
design

### New Files
- `fleet-tree-icon.png`: The fleet icon displayed in the treeview.
- `fleetTreeItem.ts`: The class responsible for the Fleet node in the treeview. It includes the `createFleetTreeNode` function and all necessary properties.

### Modified Files
- `azure-api-util.ts`: Modified the `parseResource` function to enable parsing a member cluster to a cluster resource.
- `azureResources.ts`: Defined the `DefinedFleetMemberWithGroup` type, added Fleet-related resource types, and added functions to retrieve Fleet members and convert a FleetMember to a DefinedFleetMemberWithGroup.
- `config.ts`: Refactored and added the cluster filtering function implemented in this PR [#1208](https://github.com/Azure/vscode-aks-tools/pull/1208).
- `aksClusterTreeItem.ts`: Deleted the `subscriptionId` field and used the `clusterResource`'s id to get the `subscriptionId` for more concise code. Modified the parent node type so that the parent node of a cluster can be either a subscription or a Fleet.
- `subscriptionTreeItem.ts`: Merged the cluster filter into the current PR by refactoring and moving the related code into `config.ts`. Added the process of fetching fleets and mapping members to their corresponding fleets before loading.

## Conflict Resolution (with cluster-filter PR [#1208](https://github.com/Azure/vscode-aks-tools/pull/1208))
This PR also resolves conflicts with the cluster-filter PR by refactoring the filter-related code (mentioned in `config.ts` and `subscriptionTreeItem.ts` in the Modified Files section above). However, further integration testing is needed to ensure both functionalities work well together.

Some known issues are:
- When the user clicks the filter button next to a subscription, the cluster pool only contains clusters from the current subscription. However, as mentioned above, a Fleet can have member clusters from different subscriptions. In this case, member clusters from other subscriptions will never be visible to the user. The user must go to the corresponding subscription to see the member clusters. It also makes sense to leave it as it is since it is a filter based on the subscription and there might not be a need to have members from another subscription.
no external cluster
[...]

PR #1219, opened 2025-02-04
showing fleet in the treeview

JunyuQian:

PR #1210, opened 2025-01-30
[feat] Add fleet in the tree

serbrech:
The changes include defining new fleet tree item types, creating tree items for fleet resources, and updating the subscription tree to handle both cluster and fleet resources.

image

### Support for fleets:

* [`src/commands/utils/azureResources.ts`](diffhunk://#diff-c8b219f70153e2e69c9fa95ac2f013ea486a4fb6fc530015e95124e3b43a488dR10-R18): Added `fleetResourceName` and `fleetResourceType` constants, and included `fleetResourceType` in the `resourceTypes` array.

### Tree item creation and management:

* [`src/tree/fleetTreeItem.ts`](diffhunk://#diff-fa53d9e90cb3432b20329fef2b51a9e05591b1c06e0b4cec7d643f9ef4435be2R1-R64): Introduced `FleetTreeNode` interface and `FleetTreeItem` class to represent fleet resources in the tree view. Added `createFleetTreeNode` function to instantiate fleet tree items.

### Subscription tree updates:

* [`src/tree/subscriptionTreeItem.ts`](diffhunk://#diff-38c51d7f05aaf0d3fd5b401cee6e7a14b6e2c8c0b319929719e50e4d0a3b1ecaR8-R12): Imported `createFleetTreeNode` function and `fleetResourceType` constant. Updated the `SubscriptionTreeItem` class to fetch fleet resources and concatenate them with cluster resources before creating tree nodes. [[1]](diffhunk://#diff-38c51d7f05aaf0d3fd5b401cee6e7a14b6e2c8c0b319929719e50e4d0a3b1ecaR8-R12) [[2]](diffhunk://#diff-38c51d7f05aaf0d3fd5b401cee6e7a14b6e2c8c0b319929719e50e4d0a3b1ecaL79-R109)

PR #1237, opened 2025-02-12
Showing Fleets in the Treeview

JunyuQian:
# Showing Fleets in the Treeview

This PR adds the Fleet resource to the treeview, based on @serbrech's initial work. The image below shows the final result of the treeview.
treeview

Due to the special characteristics that a Fleet can have member clusters from different subscriptions, the initial treeview design, which classifies clusters based on subscriptions, does not quite fit Fleets. The current design can be summarized into three cases below:
design

### New Files
- `fleet-tree-icon.png`: The fleet icon displayed in the treeview.
- `fleetTreeItem.ts`: The class responsible for the Fleet node in the treeview. It includes the `createFleetTreeNode` function and all necessary properties.

### Modified Files
-  `package.json` & `arm.ts`: Included the arm resource Graph API which is used to fetch resources in a more efficient way.
- `azure-api-util.ts`: Modified the `parseResource` function to enable parsing a member cluster to a cluster resource.
- `azureResources.ts`: Defined the `DefinedFleetMemberWithGroup` type and Fleet-related resource types. Added functions to retrieve all clusters and Fleets using Graph API and convert a `FleetMember` to a `DefinedFleetMemberWithGroup`. 
- `clusterfilter.ts`: Used Graph API to fetch all resources efficiently.
- `config.ts`: Added the interface that fits to both clusters and Fleets.
- `aksClusterTreeItem.ts`: Modified the parent node type so that the parent node of a cluster can be either a subscription or a Fleet.
- `subscriptionTreeItem.ts`: Merged the cluster filter [#1208](https://github.com/Azure/vscode-aks-tools/pull/1208) into the current PR. Added the process of fetching fleets and mapping members to their corresponding fleets before loading.

Some known issues are:
- When the user clicks the filter button next to a subscription, the cluster pool only contains clusters from the current subscription. However, as mentioned above, a Fleet can have member clusters from different subscriptions. In this case, member clusters from other subscriptions will never be visible to the user. The user must go to the corresponding subscription to see the member clusters. It also makes sense to leave it as it is since it is a filter based on the subscription and there might not be a need to have members from another subscription.
no external cluster
- Once the user sets some filter conditions, there is no way to clear the filter. This issue has been discussed, and later the user will be able to clear the conditions by unselecting all the filtering conditions.

## Next Steps
- Implement the `hasMoreChildren` function in `fleetTreeItems.ts`, so that the resources will be loaded gradually (e.g., load 10 lines, and display the "load more" button), rather than load them all at once. Currently, in the testing environment, the number of resources is not quite big. But in reality, it should have the ability to handle a large amount of resources.

---

Last but not least, a huge thank-you to @serbrech and @Tatsinnit, who provided immense assistance to this PR and spent hours peer-programming with me ❤️ Looking forward to moving on to the next feature!


src/tree/subscriptionTreeItem.ts
```diff
@@ -27,7 +36,7 @@ export function isSubscriptionTreeNode(node: unknown): node is SubscriptionTreeN
     return node instanceof SubscriptionTreeItem;
 }
 
-export function createSubscriptionTreeItem(
+export function createSubscriptionTreeItem( // create subscription node
```
JunyuQian, line 39: will remove the unnecessary comment

src/tree/subscriptionTreeItem.ts
```diff
@@ -73,43 +82,115 @@ class SubscriptionTreeItem extends AzExtParentTreeItem implements SubscriptionTr
...
+        const fleetTreeNodes = new Map<string, FleetTreeNode>();
+        const clusterTreeItems = new Map<string, AzExtTreeItem>();
+        fleetResources.concat(clusterResources).forEach((r) => {
+            if (r.type?.toLocaleLowerCase() === fleetResourceType.toLowerCase()) {
+                const fleetTreeItem = createFleetTreeNode(this, this.subscriptionId, r);
+                fleetTreeItem.addCluster(fleetToMembersMap.get(r.id) || []);
+                fleetTreeNodes.set(r.id, fleetTreeItem);
+                return fleetTreeItem;
+            } else if (r.type?.toLocaleLowerCase() === clusterResourceType.toLowerCase()) {
+                // Check if the subscription is in the filter for SelctedClustersFilter
```
JunyuQian, line 170: will fix the typo

Tatsinnit:
🎊 Than you for this, can we please close most of the PR like: #1220 , #1219 & #1210 which are doing same thing I guess, I have close #1224 

Thanks you!

src/commands/utils/azureResources.ts
```diff
@@ -35,6 +51,25 @@ export async function getResources(
     return errmap(list, (resources) => resources.filter(isDefinedResource).map(asResourceWithGroup));
 }
 
+export async function getClusterAndFleetResourcesFromGraphAPI(
+    sessionProvider: ReadyAzureSessionProvider,
+    subscriptionId: string,
+): Promise<Errorable<AksClusterAndFleet[]>> {
+    const client = getGraphResourceClient(sessionProvider);
+    const query = {
+        query: `Resources | where type =~ '${clusterResourceType}' or type =~ '${fleetResourceType}' | project id, name, location, resourceGroup, subscriptionId, type`,
```
Tatsinnit, line 60: 💡 Please note: This is the place we use introduce `resourceGraph` API call and for archival reasons we would make this for read operation because of various ease factor but some for my reasons beneficial are. Here’s why:  

### **Performance & Efficiency**  

- **Single Query vs. Multiple API Calls**: Instead of making multiple SDK calls to different Azure services (e.g., `@azure/arm-compute`, `@azure/arm-network`), Resource Graph allows you to query all resources at once.  
- **Optimized for Large-Scale Queries**: It uses an efficient query engine that is designed for retrieving data across large-scale Azure environments quickly.  

### **Reduced API Throttling**  
- Normal SDK calls are subject to **rate limits** when querying many resources.  
- Resource Graph is designed for bulk queries and is less prone to throttling.  

### **Lower Latency**  
- Instead of multiple sequential API calls, Resource Graph retrieves data in **parallel**, reducing latency. 

### **In-summary**  

For querying and reporting on Azure resources, `@azure/arm-resourcegraph` is **far superior** to making multiple service-specific SDK calls. However, for managing or modifying individual resources, the normal SDKs are still required.

src/commands/utils/clusterfilter.ts
```diff
@@ -32,29 +30,33 @@ export default async function aksClusterFilter(_context: IActionContext, target:
...
+            sessionProvider.result,
+            subscriptionNode.result.subscriptionId,
+        );
+        if (failed(aksClusters)) {
+            vscode.window.showErrorMessage(aksClusters.error);
+            return;
         }
+        clusterList = aksClusters.result.filter(
+            // only keep clusters for the cluster filter (remove fleets from the list)
+            (r) => r.type.toLowerCase() === "microsoft.containerservice/managedclusters",
```
Tatsinnit, line 46: 💡 We can always replace this with the type called `clusterResourceType`, but lets do it in next incremental way, I am commenting in case anyone notices.

JunyuQian: Sounds great! That improves the maintainability of the code.
I'll fix it together with the comment-related issues I mentioned above :)

src/commands/utils/clusterfilter.ts
```diff
@@ -32,29 +30,33 @@ export default async function aksClusterFilter(_context: IActionContext, target:
...
+        );
     });
 
     const filteredClusters = await getUniqueClusters();
 
-    const quickPickItems: ClusterQuickPickItem[] = clusterList.map((cluster) => {
+    const quickPickItems: ClusterQuickPickItem[] = clusterList.map((cluster: AksClusterAndFleet) => {
         return {
             label: cluster.name || "",
```
JunyuQian, line 54: remove the empty string as it is not acceptable -- david

JunyuQian:
Here is the VSIX file for the webview test. Please let me know if there is any issue. Thank you :)
[vscode-aks-tools-1.5.5-treeview-vsix.vsix.zip](https://github.com/user-attachments/files/18762081/vscode-aks-tools-1.5.5-treeview-vsix.vsix.zip)

dvadas (approved):

Tatsinnit (approved):

ReinierCC (approved):
Tested VSIX on Windows, Looks good!

tejhan (approved):
Tested v6 & everythings working great, LGTM 👍

src/azure-api-utils.ts
```diff
@@ -1,9 +1,17 @@
...
+    name: string | undefined;
+} {
+    // General armId Format: /subscriptions/{subid}/resourcegroups/{group}/providers/.../{name}
+    // armId format of a member cluster, given parentResource is a Fleet {parentResourceId}/members/{name}
     const bits = armId.split("/").filter((bit) => bit.length > 0);
     const resourceGroupName = bits[3];
+    const subscriptionId = bits[1];
     const name = bits[bits.length - 1];
-    return { resourceGroupName, name };
+    const parentResourceId = `/${bits.slice(0, bits.length - 2).join("/")}`; // take all except for the last two bits
```
hsubramanianaks, line 13: I feel this parseResource can be written better, given the PR size and time, we can do this as follow up or incremental.,

src/tree/subscriptionTreeItem.ts
```diff
@@ -73,43 +82,115 @@ class SubscriptionTreeItem extends AzExtParentTreeItem implements SubscriptionTr
...
+
+        // create tree nodes for filtered clusters and fleets
+        const filteredClusters = getFilteredClusters();
+
+        const fleetTreeNodes = new Map<string, FleetTreeNode>();
+        const clusterTreeItems = new Map<string, AzExtTreeItem>();
+        fleetResources.concat(clusterResources).forEach((r) => {
+            if (r.type?.toLocaleLowerCase() === fleetResourceType.toLowerCase()) {
+                const fleetTreeItem = createFleetTreeNode(this, this.subscriptionId, r);
+                fleetTreeItem.addCluster(fleetToMembersMap.get(r.id) || []);
```
hsubramanianaks, line 166: I have tested this using the VSIX  but for some of the subscriptions it doesn't show me the member clusters under the fleet tree view, which could be this line causing it or some role assignments are missing. Comparing it with portal, portal does few more steps which might be missing here. But that can be done as incremental PR. Good Job on this PR though @JunyuQian 
image

hsubramanianaks (approved):
LGTM, except few comments which can be handled as incremental PR's. Thank you @JunyuQian

PR #1240, opened 2025-02-13
Fix Cluster filter default behaviour

### LINKED ISSUE #1223: Old Filter behaviour: Lets Fix it.
#### **Issue Summary**  

When applying a filter on **Subscription (Sub)** or **Cluster**, and then unticking all selections, the expected behavior should be the same as **selecting all**—i.e., a "no-filter" state. However, currently, unticking all does nothing, which is an old existing behavior.  

#### **Steps to Reproduce**  

1. Apply a filter by selecting one or more Sub/Cluster options.  
2. Open the filter menu again and **untick all** options.  
3. Observe that instead of reverting to a "no-filter" state (showing all results), the filter remains inactive or behaves inconsistently.  

#### **Expected Behavior**  

- **Selecting all** = No filter applied (all results are shown).  
- **Unticking all** should also behave the same (no filter applied, all results shown).  

#### **Current Behavior**  

- Unticking all selections does **not** revert to a no-filter state, leading to confusion.  

#### **Why This Matters?**  

A user recently pointed out that from a practical **user-scenario perspective**, **"select all" and "select none" should behave the same way**—which makes sense for a more intuitive experience.  

#### **Screenshots**  

~~// ToDo: Tats will add this soon - (Attach the screenshots here)~~

Image

Image
  

#### **Proposed Fix**  

Ensure that **unticking all selections** in the filter behaves like a **no-filter state**, just as **selecting all** does.  

Thanks heaps all,

Tatsinnit:
Hiya all, ❤️ Thank you as always! The previous implementation had redundant reassignment of `filters`, multiple checks for the same condition, and slightly convoluted logic. This refactor makes the function more intuitive and performant while preserving its intended functionality.  

I think we can do this much better with fleet type filtering as well. But for the next release opening this for thoughts. 

As a key sidetone the reason currently `mergeFilter` is needed is because the tree fire event refreshes the whole node tree, we should avoid it. This kind of fixes #1223 

Rest here is the VSIX for quick spin as well please: 
[vscode-aks-tools-1.5.5-clusterfilterfix.vsix.zip](https://github.com/user-attachments/files/18777089/vscode-aks-tools-1.5.5-clusterfilterfix.vsix.zip)

### DIFF
--- a/src/commands/utils/clusterfilter.ts
+++ b/src/commands/utils/clusterfilter.ts
@@ -70,9 +70,11 @@ export default async function aksClusterFilter(_context: IActionContext, target:
     });
 
     // Set Cluster Instance
-    const newFilteredClusters = selectedItems ? [
-        ...selectedItems.map((item) => item.Cluster), // Retain filters in any other tenants.
-    ] : [];
+    const newFilteredClusters = selectedItems
+        ? [
+              ...selectedItems.map((item) => item.Cluster), // Retain filters in any other tenants.
+          ]
+        : [];
 
     await setFilteredClusters(newFilteredClusters, clusterList);
 }
--- a/src/commands/utils/clusterfilter.ts
+++ b/src/commands/utils/clusterfilter.ts
@@ -69,16 +69,12 @@ export default async function aksClusterFilter(_context: IActionContext, target:
         placeHolder: "Select Cluster",
     });
 
-    if (!selectedItems) {
-        return;
-    }
-
     // Set Cluster Instance
-    const newFilteredClusters = [
+    const newFilteredClusters = selectedItems ? [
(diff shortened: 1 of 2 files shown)


src/commands/utils/clusterfilter.ts
```diff
@@ -69,16 +69,14 @@ export default async function aksClusterFilter(_context: IActionContext, target:
         placeHolder: "Select Cluster",
     });
 
-    if (!selectedItems) {
```
hsubramanianaks, line 72: Nice catch :) Tats.

hsubramanianaks (approved):
LGTM, tested in ubuntu works well.

ReinierCC:
Looks good! Tested on windows.

ReinierCC (approved):

PR #1254, opened 2025-02-17
Treeview bug fix: showing members from different subscriptions

JunyuQian:
This PR resolves the issue where member clusters from different subscriptions were not displayed as part of the Fleet (**case 1 was not correct**). The Treeview now accurately displays clusters and Fleets following the rules illustrated in the picture below.

image

**Before this PR:**
- Cluster 1 was only displayed under Subscription 1 and not under Fleet 1 in Subscription 2.

**With this PR:**
- Cluster 1 is now displayed under both Subscription 1 and Fleet 1 in Subscription 2.

**What is the issue?**
Previously, the cluster filter applied to all member clusters within a Fleet, regardless of their subscription. However, the cluster filter only includes clusters within the current subscription. As a result, member clusters from other subscriptions were always filtered out, preventing users from seeing them under the Fleet.

**How is it fixed?**
The solution involves adding an if statement before applying the filter. The logic is: if a member cluster is from another subscription, it is added to the treeview immediately before the filter is applied.

### DIFF
--- a/src/tree/subscriptionTreeItem.ts
+++ b/src/tree/subscriptionTreeItem.ts
@@ -123,6 +123,12 @@ class SubscriptionTreeItem extends AzExtParentTreeItem implements SubscriptionTr
             // filter out members that do not satisfy the filter (so that they are not shown in the tree)
             const membersAfterFilt = members.result.filter((m) => {
                 // for each member cluster of the fleet
+                if (parseResource(m.clusterResourceId).subscriptionId !== this.subscriptionId) {
+                    // if the member is not in the same subscription as the fleet,
+                    // there is no way it can be in the filter, as cluster-filter only contains clusters from the same subscription
+                    // to avoid being filtered out, add it to the list before applying the filter
+                    return true;
+                }
                 const filteredClusters = getFilteredClusters();
                 return filteredClusters.some(
                     // check if the member is one of the clusters in the filter


src/tree/subscriptionTreeItem.ts
```diff
@@ -123,6 +123,12 @@ class SubscriptionTreeItem extends AzExtParentTreeItem implements SubscriptionTr
             // filter out members that do not satisfy the filter (so that they are not shown in the tree)
             const membersAfterFilt = members.result.filter((m) => {
                 // for each member cluster of the fleet
+                if (parseResource(m.clusterResourceId).subscriptionId !== this.subscriptionId) {
```
Tatsinnit, line 126: 💡Nice, Thank you so much fro this context and thoughts, will it be cool, if you can share small use case which we can do to test this, this is also for the wider folks who are contributing into his as well please. Thank you.

Tatsinnit: 💡 Please note: If you can please share VSIX in your description for quick test. Thanks!

JunyuQian: > 💡Nice, Thank you so much fro this context and thoughts, will it be cool, if you can share small use case which we can do to test this, this is also for the wider folks who are contributing into his as well please. Thank you.

Sure! Here are the steps to test the feature:

1. Go to the Azure portal.
2. Create a cluster in a subscription.
3. Create an AKS Fleet Manager in another subscription.
4. In the Fleet Manager's Overview page, click "Add Member Clusters" and assign the cluster to the current Fleet.
5. Once assigned, go to VS Code and check the Treeview.

If you can see the cluster under both the Fleet and its origin subscription, then the test is successful!

JunyuQian: > 💡 Please note: If you can please share VSIX in your description for quick test. Thanks!

And here is the VSIX file :) Thanks for reminding!
[vscode-aks-tools-1.5.5-treeview-bug-fix.vsix.zip](https://github.com/user-attachments/files/18834724/vscode-aks-tools-1.5.5-treeview-bug-fix.vsix.zip)

Tatsinnit (commented):
💡 ❤️  Looks good to me, I will test this sometime in an hour and report back. gentle fyi to @tejhan , @ReinierCC or @hsubramanianaks + @qpetraroia for quick test please.

Tatsinnit (approved):

ReinierCC:
After testing I'm able to accomplish case 1 where the cluster comes from a different subscription group but I'm unable to accomplish case 2. After adding two clusters in the same subscription as the fleet I'm unable to get them to pop up under the fleet.

Waited around ~10 minutes and multiple restarts.

Could we get a second test on this? @tejhan @hsubramanianaks

tejhan:
> After testing I'm able to accomplish case 1 where the cluster comes from a different subscription group but I'm unable to accomplish case 2. After adding two clusters in the same subscription as the fleet I'm unable to get them to pop up under the fleet.
> 
> Waited around ~10 minutes and multiple restarts.
> 
> Could we get a second test on this? @tejhan @hsubramanianaks

Interesting, I just tested it out myself & it worked as intended. I'm able to get case 1 & case 2, with clusters from both inside & outside the subscription showing up. Maybe it has to do with the subscriptions being used?

ReinierCC:
> > After testing I'm able to accomplish case 1 where the cluster comes from a different subscription group but I'm unable to accomplish case 2. After adding two clusters in the same subscription as the fleet I'm unable to get them to pop up under the fleet.
> > Waited around ~10 minutes and multiple restarts.
> > Could we get a second test on this? @tejhan @hsubramanianaks
> 
> Interesting, I just tested it out myself & it worked as intended. I'm able to get case 1 & case 2, with clusters from both inside & outside the subscription showing up. Maybe it has to do with the subscriptions being used?

Ahhh true. I think there was mentioning of having to onboard the subscription for something fleet related. Could you share the specific subscription to try it on my end?

tejhan:
> > > After testing I'm able to accomplish case 1 where the cluster comes from a different subscription group but I'm unable to accomplish case 2. After adding two clusters in the same subscription as the fleet I'm unable to get them to pop up under the fleet.
> > > Waited around ~10 minutes and multiple restarts.
> > > Could we get a second test on this? @tejhan @hsubramanianaks
> > 
> > 
> > Interesting, I just tested it out myself & it worked as intended. I'm able to get case 1 & case 2, with clusters from both inside & outside the subscription showing up. Maybe it has to do with the subscriptions being used?
> 
> Ahhh true. I think there was mentioning of having to onboard the subscription for something fleet related. Could you share the specific subscription to try it on my end?

Yep, I created the fleet & initial cluster in "Azure Container Service - Test (AKS Standalone)" sub, & created outside cluster in "Azure Container Service - Test - 2 (AKS Standalone)" sub

tejhan (approved):
Both cases worked as intended in my subscriptions 👍

PR #1259, opened 2025-02-19
[Webview UI toolkit deprecation] Inspektor Gadget, Retina Capture & TCP Dump component removal

tejhan:
This PR specifically removes elements from the  Inspektor Gadget, Retina Capture & TCP Dump commands.

**.vsix for testing:** [vscode-aks-tools-1.6.0-deprecate2.vsix.zip](https://github.com/user-attachments/files/18893162/vscode-aks-tools-1.6.0-deprecate2.vsix.zip)

This is broken up as part of incremental PR's to make review & testing smoother.

For future reference:
- .secondary-button replaces VSCodeButton's appearance="secondary"
- .icon-button replaces VSCodeButton's appearance="icon"
- input[type="text"] replaces VSCodeTextInput
- input[type="checkbox"] replaces VSCodeCheckbox
- input[type="radio"] replaces VSCodeRadio
- Default "< hr >" element now has identical styling to VSCodeDivider
-  Default "< a >" element also has identical styling to VSCodeLink

PR #1276, opened 2025-02-26
Full replacement of VSCodeDropdown & VSCodeOption with custom react elements [Webview-UI-toolkit deprecation]

tejhan:
***Note* : This PR builds on top of PR's #1259 - 1261, so this can be slated for merge/review after the preceding PR's are merged.**

**This should be the second to last PR, & once approved we can slate the final PR for removal of the entire package & leftover elements**

This PR fully removes the VSCodeDropdown & VSCodeOption elements, & replaces them with CustomDropdown & CustomDropdownOption, 2 new elements that can be imported and used as a react component in any of our panels. The elements are styled with respect to vscode's display & coloring settings. 

These elements have no external dependencies which will allow us to use fully functional vscode themed dropdowns in any of our commands without needing to import any toolkits. (Also allows further customization of our own in future)

For future reference, < CustomDropdownOption > can be used as a single tag, with the label attribute serving as the label. Other attributes like onchange, disabled, etc are also functional with these new elements. < CustomDropdown > works almost identically to the replaced VSCodeDropdown, where the values & selected elements are controlled with state variables & onchange functions. 

Commands affected: 
- TCP Dump
- Inspektor Gadget 
- Create Cluster
- Create Fleet
- Azure Service Operator

Dependent Components altered:
- NodeSelector.tsx
- ResourceSelector.tsx

**For anyone testing, you can utilize the inputs in each of the effected commands to ensure submission of data is correct & the ability to scroll through options, search in dropdown, etc. are all working as intended.**

**.vsix for testing:** [vscode-aks-tools-1.6.0-dep5-2.vsix.zip](https://github.com/user-attachments/files/19069938/vscode-aks-tools-1.6.0-dep5-2.vsix.zip)


ReinierCC (approved):
Functionality looks good on Windows!

Might be worth adding a page to our documentation with all the components we should using moving forward.

webview-ui/src/CreateCluster/CreateClusterInput.tsx
```diff
@@ -32,7 +32,7 @@ export function CreateClusterInput(props: CreateClusterInputProps) {
     const [isNewResourceGroupDialogShown, setIsNewResourceGroupDialogShown] = useState(false);
     const [newResourceGroupName, setNewResourceGroupName] = useState<string | null>(null);
     const [presetSelected, setPresetSelected] = useState<PresetType>(PresetType.Automatic);
-    const [selectedIndex, setSelectedIndex] = useState<number>(0);
+    const [selectedResourceGroup, setSelectedResourceGroup] = useState<string>("");
```
Tatsinnit, line 35: 💡 Note/question: Might be worth to check hence asking if this type change will cause any breaks, given its changed from `num` to `string` type.

webview-ui/src/CreateCluster/CreateClusterInput.tsx
```diff
@@ -51,26 +51,30 @@ export function CreateClusterInput(props: CreateClusterInputProps) {
...
-        const ele = e.currentTarget as HTMLSelectElement;
-        setSelectedIndex(ele.selectedIndex);
+    function handleValidationAndIndex(value: string) {
+        handleExistingResourceGroupChange(value);
+        setSelectedResourceGroup(value);
     }
 
-    function handleExistingResourceGroupChange(e: ChangeEvent) {
-        const elem = e.currentTarget as HTMLSelectElement;
-        const resourceGroup = elem.selectedIndex <= 0 ? null : allResourceGroups[elem.selectedIndex - 1];
```
Tatsinnit, line 69: 💡 Is this change no longer needed? (reason I ask is also to document what is changed here as comments.) Thank you

Tatsinnit (approved):
❤️ Thank you so much for this and really appreciate it, congrats on the final part of this deprecation. Left few comments, jsut to make sure we capture all points. Thank you!!

webview-ui/src/CreateCluster/CreateClusterInput.tsx
tejhan: Yes, with the usage of the custom dropdown, we can refer to the string values directly. The VSCodeDropdown had the option of identifying by index, but using string identification here works identically & allows for easier variable tracking. Have tested & works well

webview-ui/src/CreateCluster/CreateClusterInput.tsx
tejhan: Yes, this change handler function was used because the original VSCodeDropdown implementation didn't have the value accessible as an attribute, so it needed to check the DOM element.

With the CustomDropdown, value is simply an attribute that can be set to a state variable, so it can be checked & changed with only an onChange function. (Line 151 shows this). Thus changing the selectedResourceGroup state variable alone will automatically change the selected values in the dropdown menu as well.

Tatsinnit (approved):

PR #1277, opened 2025-02-28
Fix: Deploy button not appearing.

Tatsinnit:
Hiya, following fix is for a bug where `Deploy` button does not appear at-all for the first time and one way to enable that is via manual `kubectl` page deploy. 

* This PR also updates the binary version for inspector gadget tool.

The issue was that for the `server version` if there is no version available the string returns is `not available`

Screenshot 2025-02-28 at 9 20 03 PM

**Fix**

This PR now adds the `semver` version check, this is better than regex IMHO because it does strict parsing, battle testes and no need to craft regex.

Thank heaps and gentle fyi to ❤️  @tejhan, @ReinierCC , @bosesuneha and @davidgamero + @qpetraroia

**VSIX Package:**

[vscode-aks-tools-1.6.0-IG-Deploy-fix.vsix.zip](https://github.com/user-attachments/files/19032211/vscode-aks-tools-1.6.0-IG-Deploy-fix.vsix.zip)

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
                 },
                 "azure.kubectlgadget.releaseTag": {
                     "type": "string",
-                    "default": "v0.34.0",
+                    "default": "v0.36.1",
                     "title": "Kubectl-gadget repository release tag",
                     "description": "Release tag for the stable kubectl-gadget tool."
                 },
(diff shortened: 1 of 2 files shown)


ReinierCC (approved):
Looks good! semver addition really cleans up the check.

ReinierCC:
Tested VSIX on Mac ✅

PR #1278, opened 2025-03-03
Update: Retina Binary to recent release.

Tatsinnit:
Hiya, Lets make this to one of recent retina release seems like we have not updated this for a while. updating with this recent release: https://github.com/microsoft/retina/releases/tag/v0.0.26 

There are bunch of work we can do post this release: 

* Like blob update.

Thanks,


ReinierCC (approved):
LGTM

PR #1291, opened 2025-03-04
TCP Dump Bug Fix

tejhan:
Slight bug observed where if a user escapes the file selection menu after clicking the button to download a TCP dump capture, the button stays stuck in the disabled downloading state.
image
This PR adds an error response to correct for that.

(To reproduce, select Troubleshoot Network Health > Collect TCP Dumps, select a node, start & stop a capture, then click the download button that appears next to completed capture item)

.vsix: [vscode-aks-tools-1.6.1-tcpfix.vsix.zip](https://github.com/user-attachments/files/19071081/vscode-aks-tools-1.6.1-tcpfix.vsix.zip)

### DIFF
--- a/src/panels/TcpDumpPanel.ts
+++ b/src/panels/TcpDumpPanel.ts
@@ -419,6 +419,14 @@ spec:
         });
 
         if (!localCaptureUri) {
+            //  Reset download state with error message if no file path was selected
+            webview.postDownloadCaptureFileResponse({
+                node,
+                captureName,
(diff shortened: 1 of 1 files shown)


Tatsinnit (approved):
💡Thank you so much for this! That's really nice, thank you so much, I will test this soon. Code looks good.

ReinierCC (approved):
Good catch. Tested on Windows, worked well.

PR #1293, opened 2025-03-05
Automatically add newly created clusters to user's selected clusters

tejhan:
This PR addresses a case where if a user has filters enabled, clusters that  they create via the extension wouldn't appear in the tree view.

- The behavior has been changed so that once a cluster has been successfully created, it is automatically added to the list of selected clusters for the user so that it appears in the tree view.

Some code has been extracted into a couple helper functions to reuse some behavior in different scenarios.

(Note: Unexpected fleet behavior might be observed incidentally but this PR doesn't intersect with that functionality)

### DIFF
--- a/src/commands/utils/clusterfilter.ts
+++ b/src/commands/utils/clusterfilter.ts
@@ -1,6 +1,6 @@
 import { IActionContext } from "@microsoft/vscode-azext-utils";
 import { ClusterQuickPickItem, getAksClusterSubscriptionNode } from "../utils/clusters";
-import { failed } from "../utils/errorable";
+import { failed, Errorable } from "../utils/errorable";
 import * as vscode from "vscode";
 import * as k8s from "vscode-kubernetes-tools-api";
 import { getExtension } from "../utils/host";
@@ -10,46 +10,22 @@ import { AksClusterAndFleet, getFilteredClusters, setFilteredClusters } from "./
 import { clusterResourceType, getClusterAndFleetResourcesFromGraphAPI } from "./azureResources";
 
 export default async function aksClusterFilter(_context: IActionContext, target: unknown): Promise<void> {
-    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
-
-    const subscriptionNode = getAksClusterSubscriptionNode(target, cloudExplorer);
-    if (failed(subscriptionNode)) {
-        vscode.window.showErrorMessage(subscriptionNode.error);
-        return;
-    }
-
     const extension = getExtension();
     if (failed(extension)) {
         vscode.window.showErrorMessage(extension.error);
         return;
     }
 
-    const sessionProvider = await getReadySessionProvider();
-    if (failed(sessionProvider)) {
-        vscode.window.showErrorMessage(sessionProvider.error);
+    // all other usual pre-checks are in this function
+    const filteredClusters = await getUniqueClusters();
+
+    const clusterList = await getClusterList(target);
+    if (failed(clusterList)) {
+        vscode.window.showErrorMessage(clusterList.error);
         return;
     }
 
-    let clusterList: AksClusterAndFleet[] = [];
-
-    await longRunning(`Getting AKS Cluster list for ${subscriptionNode.result.name}`, async () => {
-        const aksClusters = await getClusterAndFleetResourcesFromGraphAPI(
-            sessionProvider.result,
-            subscriptionNode.result.subscriptionId,
-        );
-        if (failed(aksClusters)) {
-            vscode.window.showErrorMessage(aksClusters.error);
-            return;
-        }
-        clusterList = aksClusters.result.filter(
-            // only keep clusters for the cluster filter (remove fleets from the list)
-            (r) => r.type.toLowerCase() === clusterResourceType.toLowerCase(),
-        );
-    });
-
-    const filteredClusters = await getUniqueClusters();
-
-    const quickPickItems: ClusterQuickPickItem[] = clusterList.map((cluster: AksClusterAndFleet) => {
+    const quickPickItems: ClusterQuickPickItem[] = clusterList.result.map((cluster: AksClusterAndFleet) => {
         return {
             label: cluster.name,
             description: cluster.name,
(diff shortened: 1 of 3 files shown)


src/commands/utils/clusterfilter.ts
```diff
@@ -10,46 +10,22 @@ import { AksClusterAndFleet, getFilteredClusters, setFilteredClusters } from "./
...
         return;
     }
 
-    const sessionProvider = await getReadySessionProvider();
-    if (failed(sessionProvider)) {
-        vscode.window.showErrorMessage(sessionProvider.error);
+    // all other usual pre-checks are in this function
+    const filteredClusters = await getUniqueClusters();
+
+    const clusterList = await getClusterList(target);
```
Tatsinnit, line 22: 💡 Suggestions:

if we follow logic to bail early then we really don't need to pass `target` to another level deep here, i.e. I see you have passed the whole `target` inside new function, but rather I would do pre-check early and if we really want to refactor that we can extract that to common function for various places they are use. What do you think?

```*.ts

    const subscriptionNode = getAksClusterSubscriptionNode(target, cloudExplorer);
    if (failed(subscriptionNode)) {
        return { succeeded: false, error: subscriptionNode.error };
    }

    const sessionProvider = await getReadySessionProvider();
    if (failed(sessionProvider)) {
        return { succeeded: false, error: sessionProvider.error };
    }
```

And then create something like this for actually transiting the cluster list:


```*.ts

async function fetchClusterList(
    subscription: SubscriptionNode,
    session: SessionProvider
): Promise<Errorable<AksClusterAndFleet[]>> {
    return await longRunning(
        `Getting AKS Cluster list for ${subscription.name}`,
        async (): Promise<Errorable<AksClusterAndFleet[]>> => {
            const aksClusters = await getClusterAndFleetResourcesFromGraphAPI(session, subscription.subscriptionId);
            if (failed(aksClusters)) return { succeeded: false, error: aksClusters.error };

            const clusterList = aksClusters.result.filter(
                (r) => r.type.toLowerCase() === clusterResourceType.toLowerCase()
            );
            return { succeeded: true, result: clusterList };
        }
    );
}

```

src/commands/utils/clusterfilter.ts
```diff
@@ -10,46 +10,22 @@ import { AksClusterAndFleet, getFilteredClusters, setFilteredClusters } from "./
 import { clusterResourceType, getClusterAndFleetResourcesFromGraphAPI } from "./azureResources";
 
 export default async function aksClusterFilter(_context: IActionContext, target: unknown): Promise<void> {
-    const cloudExplorer = await k8s.extension.cloudExplorer.v1;
```
Tatsinnit, line 13: 💡 How about we leave this bit here, reason being:  "fail fast" (bail early on error) Avoids executing further logic when an error is already detected. This also makes debugging easier: Errors are handled immediately rather than propagating further.

Avoid sending unknown `target` too deep and only send well-typed objects layer below.

I've left a small refactor below for discussion please.

src/commands/utils/clusterfilter.ts
```diff
@@ -76,7 +52,58 @@ export default async function aksClusterFilter(_context: IActionContext, target:
           ]
         : [];
 
-    await setFilteredClusters(newFilteredClusters, clusterList);
+    await setFilteredClusters(newFilteredClusters, clusterList.result);
+}
+
+export async function addItemToClusterFilter(target: unknown, clusterName: string, subscriptionId: string) {
+    const clusterList = await getClusterList(target);
```
Tatsinnit, line 59: 💡 Please note: this is in association with previous 2 comment, if we keep pre-checks at the top level we don't need to pass the `unknown` type object this deep.

- Also, the only reason we need `target` is to revaluate it again which is not needed if we do it in 4/5 lines at top level, What do you think?

Happy to chat more = )

src/panels/CreateClusterPanel.ts
```diff
@@ -42,18 +44,20 @@ export class CreateClusterDataProvider implements PanelDataProvider<"createClust
     private readonly containerServiceClient: ContainerServiceClient;
     private readonly featureClient: FeatureClient;
     private readonly commandId: string;
-
+    target: unknown;
```
Tatsinnit, line 47: 💡 we are passing `target` all the way here only because we want to send it as parameter type `addItemToClusterFilter` but then, this consuming func sends it to `getClusterList` for this code execution which we can avoid by only checking at the top level function of consumption.

What I mean is if we move this code below outside of `getClsuterList` , we don't need to pass an `unknown` type object too deep. Totally happy to chat more around this.

```*ts

    const subscriptionNode = getAksClusterSubscriptionNode(target, cloudExplorer);
    if (failed(subscriptionNode)) {
        return { succeeded: false, error: subscriptionNode.error };
    }

    const sessionProvider = await getReadySessionProvider();
    if (failed(sessionProvider)) {
        return { succeeded: false, error: sessionProvider.error };
    }

```

Tatsinnit (commented):
❤️ Thank you so much for this PR @tejhan , left some comments, we can chat up around this, one key minor suggestion, rest looks great! 

Thanks heaps = )

tejhan:
> ❤️ Thank you so much for this PR @tejhan , left some comments, we can chat up around this, one key minor suggestion, rest looks great!
> 
> Thanks heaps = )

Thank you Tats, I've made the refactors so that the new helper functions use other variables that are already passed like sessionprovider & subscriptionId's instead of passing the target deeply. Thanks for the suggestions.

tejhan:
> Would this account for any new clusters created through the portal?

Nope, this would specifically be for clusters created in the extension, meaning that if new clusters appear in the subscription that weren't made in the extension, it won't populate their filtered selections. Only if they have filters unset.

ReinierCC:
Tested and a new cluster popped up nicely. Not sure if it was just my older config but when I go to select clusters for the first time, it has all the clusters unselected eventhough they are currently displayed.
Screenshot 2025-04-02 at 4 50 57 PM

Tatsinnit (approved):

Tatsinnit:
> > Tested and a new cluster popped up nicely. Not sure if it was just my older config but when I go to select clusters for the first time, it has all the clusters unselected eventhough they are currently displayed. Screenshot 2025-04-02 at 4 50 57 PM
> 
> This is a good catch. I believe the current functionality exists such that all clusters show when no filters are selected. This means I believe by default the quickpick will show up with all items unchecked if no filters are set (which would make all clusters show). We can change it however to start with them all checked, as this follows more with the design of our other quickpick features.

💡 Quick question please: Is this change of behaviour form the way it currently works? Just so that we can handle that scenario with more informed pinned fyi against the repo et. al.

PR #1295, opened 2025-03-07
Quick Refactor: Moving DiagnosticSetting function to common place Cluster.ts.

Tatsinnit:
Hiya, This code move around is specifically for the `diagnosticSetting` function to `cluster.ts`  which will come handy towards the Retina hook work for the `upload to blob scenario` 

The benefit is also that we should sit down to think if we want a testing framework for most of the `utils` function :) Jest or anything else?

Gentle fyi  and thanks to @tejhan, @ReinierCC , @bosesuneha , @davidgamero + @qpetraroia for reviews please ❤️ 

Thanks.


bosesuneha (approved):
thanks for the PR, lgtm

ReinierCC (approved):
Refactor looks good

tejhan (approved):

PR #1296, opened 2025-03-10
Upgrade Cluster Kubernetes Version

ReinierCC:
This PR adds the requested user functionality to easily upgrade an AKS cluster's kubernetes version.

If a cluster is elegible for upgrade, a dropdown next kubernetes version will appear on the 'Cluster Properties' page.
Screenshot 2025-03-11 at 9 58 47 AM

Created two additional frontend components: `ClusterUpgrade.tsx` and `ConfirmationDialog` to handle the upgrade process.

A similar warning message from the upgrade documentation is placed on a confirmation screen displayed to the user before the upgrade process begins:

Screenshot 2025-03-11 at 10 04 02 AM

Upgrading:
Screenshot 2025-03-17 at 11 11 15 AM

I've done testing on what I believe are the possible states the upgrade but please do give the vsix additional testing (Upgrading, aborting, closing page, ...).

 Latest:
[vscode-aks-tools-1.6.1-upgrade-cluster-3.vsix.zip](https://github.com/user-attachments/files/19289781/vscode-aks-tools-1.6.1-upgrade-cluster-3.vsix.zip)

⛹️‍♂️

### DIFF
--- a/src/commands/utils/clusters.ts
+++ b/src/commands/utils/clusters.ts
@@ -442,6 +442,22 @@ export async function getKubernetesVersionInfo(
     }
 }
 
+export async function getClusterUpgradeProfile(
+    containerClient: azcs.ContainerServiceClient,
+    resourceGroup: string,
+    clusterName: string,
+): Promise<Errorable<azcs.ManagedClusterUpgradeProfile>> {
+    try {
+        const upgradeProfile = await containerClient.managedClusters.getUpgradeProfile(resourceGroup, clusterName);
+        return { succeeded: true, result: upgradeProfile };
+    } catch (e) {
+        return {
+            succeeded: false,
+            error: `Failed to retrieve upgrade profile for cluster ${clusterName}: ${getErrorMessage(e)}`,
+        };
+    }
+}
+
 export async function getWindowsNodePoolKubernetesVersions(
     containerClient: azcs.ContainerServiceClient,
     resourceGroupName: string,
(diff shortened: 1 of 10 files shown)


Tatsinnit (dismissed):
❤️ Thank you so much for this PR, I've left few comments, I've also played around with the implementation, I will try more latter. Most of the recommendation are coding and not experience related.

As wider note for folks, I like the usage of `upgrading` highlighted state vs using `longRunning` task given this reside in the `show-properties` page. The `State` stays intact pretty nicely. 

Thank you once again!! We can give it a release once everyone are happy.

ReinierCC:
Thanks! @Tatsinnit for the great feedback. Comments should be addressed with the latest commit.

src/commands/utils/clusters.ts
```diff
@@ -442,6 +442,22 @@ export async function getKubernetesVersionInfo(
     }
 }
 
+export async function getClusterUpgradeProfile(
+    containerClient: azcs.ContainerServiceClient,
+    resourceGroup: string,
+    clusterName: string,
+): Promise<Errorable<azcs.ManagedClusterUpgradeProfile>> {
+    try {
```
bosesuneha, line 450: can you add a check for null/empty rg and cluster name?

tejhan (commented):
Tested out & works well in multiple scenarios👍

One thing I did notice is that the upgrade indicator is white so it doesn't show well in light themes like the one I use: 
image

bosesuneha:
Tested the extension, changes look good

Tatsinnit (approved):

ReinierCC:
> Tested out & works well in multiple scenarios👍
> 
> One thing I did notice is that the upgrade indicator is white so it doesn't show well in light themes like the one I use: image

That's a great catch. Will change.

src/commands/utils/clusters.ts
ReinierCC: Added!

Tatsinnit (approved):

Issue #1294, opened 2025-03-05
Fleet member clusters behavior issue


tejhan:
There is some inconsistent behavior with clusters that are members of fleets. In some scenarios, these clusters, oftentimes hub clusters will appear in the tree even after being filtered out, and they also sometimes do not appear under the fleet themselves. It also seems that in some scenarios, clusters won't appear in one subscription based on whether another sub is in the tree as well or not.

One Example: 
First, picking only 1 item for the filter.
![Image](https://github.com/user-attachments/assets/736b5417-090a-456b-b5f0-83fb8d7b6ef5)

These additional items show up. 
![Image](https://github.com/user-attachments/assets/8be792d2-88ac-4657-8490-b7c33812e35b)

When trying to filter again, we see these clusters added.
![Image](https://github.com/user-attachments/assets/2a7d3de3-667c-460f-80dc-de1da8a3d815)

When removing all but 1 cluster from another adjacent subscription, we are left without the hub clusters showing, but fleets there. This indicates that location/subscription of fleet related clusters is causing some hiccups.
![Image](https://github.com/user-attachments/assets/1201ecfd-0b81-41a6-98aa-88a6719583a2)

This behavior is interlinked with filter behavior that I've been recently working at, so I'm currently looking at a fix.

tejhan:
Fixed in PR #1307

PR #1307, opened 2025-03-12
Fix for similarly named clusters showing in adjacent subscription nodes

tejhan:
There was an issue where clusters who share names in different subscription nodes would always be selected for both subscriptions, despite removing it from the filters. It seems when matching values across subscriptions in the filter logic, we checked for name but didn't verify if the subscriptionID's matched.

(This scenario is most easily observed with fleet hub clusters, since you can very easily have clusters named "hub" across multiple subscriptions you have selected in your configuration.)

- This PR fixes the logic for that, ensuring a selected cluster from 1 sub doesn't populate the others. 
- This fixes issue #1294 

(This scenario is most easily observed with fleet hub clusters, since you can very easily have clusters named "hub" across multiple subscriptions you have selected in your configuration.)

.vsix: [vscode-aks-tools-1.6.1-filter5.vsix.zip](https://github.com/user-attachments/files/19214637/vscode-aks-tools-1.6.1-filter5.vsix.zip)

Note: Default behavior of fleet is to appear even when there are no member clusters, so it can show in the tree upon creation. But for future, maybe a toggle to show fleets or adding fleets to the clusterfilter can be considered, as currently no matter what your filters are, the fleets will always appear. Thoughts?

### DIFF
--- a/src/commands/utils/clusterfilter.ts
+++ b/src/commands/utils/clusterfilter.ts
@@ -53,7 +53,10 @@ export default async function aksClusterFilter(_context: IActionContext, target:
         return {
             label: cluster.name,
             description: cluster.name,
-            picked: filteredClusters.some((filtered) => filtered.clusterName === cluster.name),
+            picked: filteredClusters.some(
+                (filtered) =>
+                    filtered.clusterName === cluster.name && filtered.subscriptionId === cluster.subscriptionId,
+            ),
             Cluster: {
                 clusterName: cluster.name,
                 subscriptionId: cluster.subscriptionId,
--- a/src/commands/utils/config.ts
+++ b/src/commands/utils/config.ts
@@ -144,7 +144,9 @@ export async function setFilteredClusters(filters: ClusterFilter[], allClusters:
     // Check if filters have changed
     const filtersChanged =
         existingFilters.length !== finalFilters.length ||
-        !existingFilters.every(({ clusterName }) => finalFilters.some((f) => f.clusterName === clusterName));
+        !existingFilters.every(({ clusterName, subscriptionId }) =>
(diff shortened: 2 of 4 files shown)


src/commands/utils/clusterfilter.ts
```diff
@@ -53,7 +53,10 @@ export default async function aksClusterFilter(_context: IActionContext, target:
         return {
             label: cluster.name,
             description: cluster.name,
-            picked: filteredClusters.some((filtered) => filtered.clusterName === cluster.name),
+            picked: filteredClusters.some(
+                (filtered) =>
+                    filtered.clusterName === cluster.name && filtered.subscriptionId === cluster.subscriptionId,
```
Tatsinnit, line 58: 💡 Nice, lets chat more about this, about the use cases especially the case where can it be cross sub shared cluster in fleet cases?

* I also wonder if you open one tracking item agains this repo with the brain dump of idea spread across few PR this work is being delivered = ) jsut a thought and what do you think?

Thanks heaps.

tejhan: Yes, good point. When testing, clusters still properly showed from different subscriptions under the fleet dropdown, the only thing is that users can't "uncheck" a cluster that isn't part of the subscription, so if a cluster in a fleet comes from a different subscription, it will always show under a fleet, even if all clusters are unselected for the filter.

This leads me to think that maybe in the future, including the fleets themselves in the filter selection especially if we add more fleet related features could be one way to help with this, just a thought. 

Currently looking at the issue observed earlier with fleets not working properly depending on the subscription, which would also tie into that & the larger tracking item as well.

Tatsinnit (approved):

bosesuneha (approved):
lgtm

Tatsinnit (approved):

PR #1318, opened 2025-03-24
KAITO v4.4 support

tejhan:
This PR introduces support for KAITO version 4.4 alongside fixes that allow compatibility with clusters who've installed KAITO with helm instead of the add-on or have different naming configurations. 

- Added support for phi 3.5 & qwen 2.5 coder model
- Upgraded to use vLLM api instead of transformers API for inference calls
- Added helper functions to retrieve pods based on image & tags
- Modified workspace operative functions to work with v4 changes

Note: Version 4 has not been released to the KAITO add-on as of yet, so these updates work on alternatively installed KAITO clusters with v4 enabled, but but we will have to wait until the update is pushed to the add-on for full compatibility.  (All add-on versions get auto-updated to v4)

This should be merged once version 4 is released to the add-on, so I've marked it as do-not-merge right now, but it is still testable on v4 clusters.


src/commands/aksKaito/aksKaitoManage.ts
```diff
@@ -89,7 +89,7 @@ export default async function aksKaitoManage(_context: IActionContext, target: u
         const { resourceReady, inferenceReady, workspaceReady } = getConditions(conditions);
         // The data below is used to indicate the current progress of the active model deployments
         models.push({
-            name: item.inference?.preset?.name,
+            name: item.metadata?.name,
```
Tatsinnit, line 92: 💡 I am guessing that this property/attribute is now dropped from the newer version?

src/commands/aksKaito/akskaitoGenerateYaml.ts
```diff
@@ -39,9 +39,15 @@ export default async function aksKaitoGenerateYaml(_context: IActionContext, tar
...
-    // "kaito-" is assuming kaito pods are still named kaito-workspace & kaito-gpu-provisioner
     const filterKaitoPodNames = await longRunning(`Checking if KAITO is installed.`, () => {
-        return filterPodName(sessionProvider.result, kubectl, subscriptionId, resourceGroupName, clusterName, "kaito-");
+        return filterPodImage(
+            sessionProvider.result,
+            kubectl,
+            subscriptionId,
+            resourceGroupName,
+            clusterName,
+            "mcr.microsoft.com/aks/kaito",
```
Tatsinnit, line 49: 💡 Idea share: we can `Const` it at the top, easily identifiable

src/commands/utils/clusters.ts
```diff
@@ -663,6 +663,49 @@ export async function filterPodName(
...
+            vscode.window.showErrorMessage(output.error);
+            return [];
+        }
+
+        const strOutput = output.result.stdout;
+        const pods = strOutput
+            .trim()
+            .split(";")
+            .reduce<{ nameSpace: string; podName: string; imageName: string }[]>((acc, val, index, arr) => {
+                if (index % 3 === 0 && arr[index + 1] && arr[index + 2]) {
```
Tatsinnit, line 698: 💡 Idea only please: might be add a comment for the same of understanding :) and relevance of `index % 3 === 0 && arr[index + 1] && arr[index + 2]` expectation et. al. What do you think?

src/commands/utils/clusters.ts
```diff
@@ -663,6 +663,49 @@ export async function filterPodName(
...
+        return cluster;
+    }
+
+    const kubeconfig = await getKubeconfigYaml(sessionProvider, subscriptionId, resourceGroup, cluster.result);
+    if (failed(kubeconfig)) {
+        return kubeconfig;
+    }
+
+    const result = await withOptionalTempFile(kubeconfig.result, "yaml", async (kubeconfigPath) => {
+        const command = `get pods -A -o jsonpath="{range .items[*]}{.metadata.namespace};{.metadata.name};{.spec.containers[*].image};{end}"`;
```
Tatsinnit, line 686: Idea only: **Not under the scope of this PR,** but once this is merged we can shorten this function:


```.ts
// Helper type
type PodInfo = { nameSpace: string; podName: string; imageName: string };

// Extracts pod info from kubectl output
function parseKubectlOutput(output: string): PodInfo[] {
    return output
        .trim()
        .split(";")
        .reduce<PodInfo[]>((acc, val, index, arr) => {
            if (index % 3 === 0 && arr[index + 1] && arr[index + 2]) {
                acc.push({
                    nameSpace: val,
                    podName: arr[index + 1],
                    imageName: arr[index + 2],
                });
            }
            return acc;
        }, []);
}

// Runs kubectl to get pod info
async function getPodsFromCluster(
    kubectl: APIAvailable<KubectlV1>,
    kubeconfig: string
): Promise<Errorable<PodInfo[]>> {
    const command =
        `get pods -A -o jsonpath="{range .items[*]}{.metadata.namespace};{.metadata.name};{.spec.containers[*].image};{end}"`;

    const output = await invokeKubectlCommand(kubectl, kubeconfig, command);
    if (failed(output)) {
        vscode.window.showErrorMessage(output.error);
        return { succeeded: true, result: [] };
    }

    return { succeeded: true, result: parseKubectlOutput(output.result.stdout) };
}

// Main exported function
export async function filterPodImage(
    kubectl: APIAvailable<KubectlV1>,
    subscriptionId: string,
    resourceGroup: string,
    clusterName: string,
    imageNameStartsWith: string
): Promise<Errorable<PodInfo[]>> {
    const cluster = await getManagedCluster(sessionProvider, subscriptionId, resourceGroup, clusterName);
    if (failed(cluster)) {
        return cluster;
    }

    const kubeconfig = await getKubeconfigYaml(sessionProvider, subscriptionId, resourceGroup, cluster.result);
    if (failed(kubeconfig)) {
        return kubeconfig;
    }

    const result = await withOptionalTempFile(kubeconfig.result, "yaml", (kubeconfigPath) =>
        getPodsFromCluster(kubectl, kubeconfigPath)
    );

    if (failed(result)) {
        return result;
    }

    const filteredPods = result.result.filter((pod) => pod.imageName.startsWith(imageNameStartsWith));
    return { succeeded: true, result: filteredPods };
}

```

Tatsinnit (dismissed):
👍 ❤️ Thank you so much for this @tejhan, looks good to me. I've tested this by switching to this branch as well, please note only comment is for us to tackle things latter. Please let me know and if sounds good we can align this for a release. (_I know you have added label of donotmrege_) = )

Thank you!!

Tatsinnit (approved):

Tatsinnit (approved):

Issue #1344, opened 2025-04-03
Unable to register extension for Windows on non-wsl Scenario.


Tatsinnit:
**Describe the bug**

Currently the AKS extension is not getting registered for the extension run on windows with non-wsl under the hood, the developer tool error captured is as follows:

I am on Mac and it works perfectly fine in Mac which tells me it has to be something underlying the way a package which has non-wsl issue got update and we are inheriting this underlying behaviour from it.

![Image](https://github.com/user-attachments/assets/0e051439-7b94-462e-937a-003eefd920d1)

### Temporary Solution:

Downgrade your VSCE extension version to `1.6.1` and things will be back to normal, while we artily zero in the really cause of this issue. (Fyi:  @kejatura-dev, @tejhan, @bosesuneha and @ReinierCC ) 

**To Reproduce**

make sure Mahican is windows with non-wsl scenario

1. Go to 'VSC Extension Marketplace;
2. Get latest 1.6.2 and above and **only on widows - non-wsl ** case this issue will happen
3. Visually cloud explorer tree will be unreachable but in vscode Deve tool you will see error as above screenshot.

**Expected behavior**

Tree should load as expected.

PR #1345, opened 2025-04-03
Reverting @azure/identity update to fix issues in non-wsl Windows

tejhan:
This pull request reverts recent updates to the `@azure/identity` package and its dependency `open`, due to critical issues affecting Windows users. The updates introduced breaking changes that prevented Windows users from loading and viewing subscriptions or clusters within the VSCode AKS Tools extension.

- Addresses issue https://github.com/Azure/vscode-aks-tools/issues/1344
- Reverted update from `@azure/identity` from 4.8 back to 4.6
- Reverted `open` from 10.1 to 8.0 (This package specifically or (it's depdendencies)) are what introduced the breaking changes.

For a bit of detail - it seems to be that the update to  the open package included some new dependencies such as `default-browser` & `is-inside-container` package, which also came along with additional subdependencies of `bundle-name` & `run-applescript` alongside updates to other OS-related packages like `is-wsl` & removal of previous sub dependency `is-docker`. One of these packages caused the issue & since they all mutually need eachother for this update, we will have to exclude them all for now to provide functionality for windows users. 

All of these packages are required for version 10 of `open`, which is required for version 4.8 of `@azure/identity`, so for now we can revert this major change & prevent dependabot from auto-updating the `@azure/identity` or `open` package as of now.

.vsix for testing -> [vscode-aks-tools-1.6.3-windowsfix.vsix.zip](https://github.com/user-attachments/files/19591011/vscode-aks-tools-1.6.3-windowsfix.vsix.zip)

We can do further investigation to pin down the specific individual package causing the issue for further decisions, but for now this serves to fix the issue by reverting the major upgrades to the parent packages while maintaining full functionality.

### DIFF
--- a/package.json
+++ b/package.json
@@ -715,7 +715,7 @@
         "@azure/arm-subscriptions": "^5.1.0",
         "@azure/container-registry": "^1.1.0",
         "@azure/core-auth": "^1.8.0",
-        "@azure/identity": "^4.8.0",
+        "@azure/identity": "4.6",
         "@azure/ms-rest-azure-env": "^2.0.0",
         "@azure/storage-blob": "^12.26.0",
         "@eslint/compat": "^1.2.7",
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
                 "@azure/arm-subscriptions": "^5.1.0",
                 "@azure/container-registry": "^1.1.0",
                 "@azure/core-auth": "^1.8.0",
-                "@azure/identity": "^4.8.0",
+                "@azure/identity": "4.6",
                 "@azure/ms-rest-azure-env": "^2.0.0",
                 "@azure/storage-blob": "^12.26.0",
                 "@eslint/compat": "^1.2.7",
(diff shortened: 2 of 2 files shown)


kejatura-dev:
Verified the VSX package, reverting @azure/identity package to 4.6 works. Can now load AKS extension properly on Windows

bosesuneha:
tested on Windows, AKS extension is getting loaded as expected.

bosesuneha (approved):

Tatsinnit:
❤️🥂 thank you so much for this!! Indeed let’s open any tracking bug because the pancake update was minor version which added breaking change that’s disruptive. 

❤️☕️ merging it and cutting a release.

Tatsinnit (approved):

package.json
```diff
@@ -715,7 +715,7 @@
         "@azure/arm-subscriptions": "^5.1.0",
         "@azure/container-registry": "^1.1.0",
         "@azure/core-auth": "^1.8.0",
-        "@azure/identity": "^4.8.0",
+        "@azure/identity": "4.6",
```
Tatsinnit, line 718: Minor do we want to keep the whole 4.6.0 but just a bit and if it works I am merging it 😊☕️

PR #1349, opened 2025-04-07
Fix for viewing downloaded files when using WSL

tejhan:
Observed small issue where attempting to view downloaded files didn't work in TCP Dump & Retina panels - specifically in WSL.

It seems that the `revealFileInOS` & `open` functions both don't appear to work in WSL, so this PR implements a small fix to check for WSL environments & if detected, use the `remote-wsl.revealInExplorer` command instead.

.vsix for testing: 
[vscode-aks-tools-1.6.4-wslfolderfix.vsix.zip](https://github.com/user-attachments/files/19635675/vscode-aks-tools-1.6.4-wslfolderfix.vsix.zip)


bosesuneha:
tested on wsl and windows, works fine

bosesuneha (approved):

Tatsinnit (approved):

PR #1408, opened 2025-05-07
Add the enable copilot flag for gh copilot feature scenario.

Tatsinnit:
This PR is to add an enable flag which can be set using `setAIRecommendationsInfoState` or get the state by using `getAIRecommendationsInfoState` for the case of enablement, one such way is like this below as sample:

Below is **just sample how this flag could be used**. cc: @tejhan and @kejatura-dev fyi please. ❤️ 

```*.ts
       // Sample user prompt to enable the Azure Kubernetes Service GH CoPilot Feature
        const value = await vscode.window.showInformationMessage(
            `Do you want to enable the Azure Kubernetes Service GH CoPilot Feature?`,
            "Yes",
            "No",
        );
        if (value === "Yes") {
            setAIRecommendationsInfoState(true);
        }
        if (value === "No") {
            setAIRecommendationsInfoState(false);
        }
```

VSIX to test:

* [vscode-aks-tools-1.6.6-copilotEnabled-test11.vsix.zip](https://github.com/user-attachments/files/20113174/vscode-aks-tools-1.6.6-copilotEnabled-test11.vsix.zip)

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,6 +164,11 @@
                     "default": "v0.0.26",
                     "title": "Retina repository release tag",
                     "description": "Release tag for the stable Retina tool release."
+                },
+                "vs-kubernetes.enable-gh-copilot": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Set to true to enable GH Copilot hook."
                 }
             }
         },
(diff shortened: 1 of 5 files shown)


tejhan (approved):

kejatura-dev (approved):
Tested latest VSIX and is working as expected!

PR #1409, opened 2025-05-09
Documentation for the copilot flag.

Tatsinnit:
This is the addition as to how to add the flag as false for the copilot.

Thanks heaps.

### DIFF
--- a/docs/book/src/features/aks-plugins-github-copilot.md
+++ b/docs/book/src/features/aks-plugins-github-copilot.md
@@ -6,6 +6,15 @@ The AKS plugins (or skills) for GitHub Copilot for Azure (@azure) extension enab
 
 ## **Features**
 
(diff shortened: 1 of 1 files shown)


kejatura-dev (approved):
lgtm

kejatura-dev:
Is there a way we can link this documentation in the dialog box as well?

Tatsinnit:
> > Is there a way we can link this documentation in the dialog box as well?
> 
> I guess if they have already explicitly set this to be disabled, there is no need for linking as they should already be aware where to change this

Thank you for this comment, I have sorted this out in separate branch, we could add this latter in the day post release with the link directly to the doc as well in case of need. But yes what you just said is exactly correct, if you know how to disable it then you know how to enable it = )  🥷

tejhan (approved):

PR #1444, opened 2025-05-29
Adding namespace awareness & different runtime support for KAITO models

tejhan:
This PR enables namespace awareness amongst KAITO features, allowing workspaces deployed in non-standard namespaces to be recognized & used for deleting models, viewing them, and querying the inference server properly. 

Additionally, I've added a check for the runtime of the KAITO model, which allows selection of a proper query to the inference server based on the runtime that's being utilized. 

.vsix for testing: [Uploading vscode-aks-tools-1.6.8-runtimefix.vsix.zip…]()

Thank you to @Tatsinnit and @lyantovski or contributions & insights.

### DIFF
--- a/src/panels/KaitoManagePanel.ts
+++ b/src/panels/KaitoManagePanel.ts
@@ -154,7 +154,7 @@ export class KaitoManagePanelDataProvider implements PanelDataProvider<"kaitoMan
 
     // Updates the current state of models on the cluster
     private async updateModels(webview: MessageSink<ToWebViewMsgDef>) {
-        const command = `get workspace -o json`;
+        const command = `get workspace -A -o json`;
         const kubectlresult = await invokeKubectlCommand(this.kubectl, this.kubeConfigFilePath, command);
         if (failed(kubectlresult)) {
             webview.postMonitorUpdate({
--- a/src/panels/KaitoManagePanel.ts
+++ b/src/panels/KaitoManagePanel.ts
@@ -68,23 +68,27 @@ export class KaitoManagePanelDataProvider implements PanelDataProvider<"kaitoMan
                 this.handleMonitorUpdateRequest(webview);
             },
             deleteWorkspaceRequest: (params) => {
-                this.handleDeleteWorkspaceRequest(params.model, webview);
+                this.handleDeleteWorkspaceRequest(params.model, params.namespace, webview);
             },
             redeployWorkspaceRequest: (params) => {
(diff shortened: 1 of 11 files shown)


Tatsinnit (dismissed):
❤️ Good catch and its not a breaking change, happy to get this released once we get go-ahead form quick test.

src/panels/KaitoManagePanel.ts
```diff
@@ -177,6 +188,7 @@ export class KaitoManagePanelDataProvider implements PanelDataProvider<"kaitoMan
                 inferenceReady: inferenceReady,
                 workspaceReady: workspaceReady,
                 age: convertAgeToMinutes(item.metadata?.creationTimestamp),
+                namespace: item.metadata?.namespace,
```
Tatsinnit, line 191: 💡 Thought share please: if nothing all we default to `default` namespace or is it intrinsically taken care here? (Came to my mind, I thought I will explicitly ask) = )

src/panels/KaitoModelsPanel.ts
```diff
@@ -103,7 +104,11 @@ export class KaitoModelsPanelDataProvider implements PanelDataProvider<"kaitoMod
     private async handleKaitoManageRedirectRequest() {
         vscode.commands.executeCommand("aks.aksKaitoManage", this.newtarget);
     }
-    private async handleGenerateCRDRequest(yaml: string) {
+    private async handleGenerateCRDRequest(yaml: string | undefined) {
```
Tatsinnit, line 107: 💡 Note for reader: I see, its a private func and scope is KAITO if it was public - I would have asked about the `undefined` but since its specific to Kaito use letting this for the scenario case reasoning.

src/panels/KaitoManagePanel.ts
tejhan: Namespace is required for workspaces, so it should always return a value unless the workspace definition changes in the future. I will put a defensive default here though to be safe, thanks for pointing it out.

src/panels/KaitoModelsPanel.ts
tejhan: Yes, this undefined option was added alongside other logic as an additional safeguard - since there's a chain of multiple functions called before this one is reached, there has been some modifications to other functions that will catch if the yaml file isn't properly loaded & return undefined, which will then trigger an error catch - rather than return & further use an incompleted/corrupted yaml.

PR #1465, opened 2025-06-11
Updating supported models for v0.4.6 of KAITO

tejhan:
This PR adds the Llama3 and Qwen32b models & removes the Llama2 model which has now been deprecated in KAITO.
- Added Llama3 & Qwen32b
- Bumped apiversion from v1aplha1 to v1beta1
- Removed unused llama2 logic


Tatsinnit (approved):
💡 ❤️  Thank you for this update and for what it's worth we can also get Lior aka @lyantovski  for a quick eyes & fyi as well.

for readers regarding use of `beta1` here is the more detail in Kaito's original page: https://github.com/kaito-project/kaito?tab=readme-ov-file#quick-start 

THanks

PR #1466, opened 2025-06-12
IG menu restructure:  Investigate DNS and Real-time TCP Monitoring

bosesuneha:
This PR introduces two new commands under the Troubleshoot Network Health menu: `Investigate DNS` and `Real-time TCP Monitoring`. These options allow users to directly launch the respective IG tools from the menu, streamlining access to common troubleshooting actions. The `Show Inspektor Gadget` command has now been moved up a level in the menu hierarchy to make it more accessible.

If IG is not currently deployed on the cluster, the user will be prompted to deploy it. Once running, the commands will open the New Trace Dialog with the appropriate tool pre-selected.

This update improves usability by enabling quick, use-case-driven access to IG tools without requiring users to navigate through the full IG UI. It’s the first step in a series of PRs aimed at restructuring the IG menu to be more intuitive and user-friendly. 

The goal is to shift toward a UX where users can troubleshoot issues based on symptoms or use cases—without needing to know which specific IG gadget to choose. This will make the tool more approachable for a broader range of users, especially those less familiar with Kubernetes internals.

VSIX for testing: 
[vscode-aks-tools-1.6.10-ig-menu-restrutcure-1.0.vsix.zip](https://github.com/user-attachments/files/21106349/vscode-aks-tools-1.6.10-ig-menu-restrutcure-1.0.vsix.zip)


src/commands/aksInspektorGadget/inspektorGadgetHelper.ts
```diff
@@ -0,0 +1,160 @@
...
+    const isRunning = await clusterOperations.isInspektorGadgetRunning();
+    if (failed(isRunning)) {
+        vscode.window.showErrorMessage(`Failed to check Inspektor Gadget status: ${isRunning.error}`);
+        return;
+    }
+
+    if (!isRunning.result) {
+        const items = [
+            {
+                label: "$(rocket) Deploy Inspektor Gadget",
```
Tatsinnit, line 93: 💡 What's this `rocket` will represent as, please? = )

bosesuneha: Yes this the 🚀 icon

Tatsinnit (approved):
❤️ 👍  Thank you so much for this and all looks great!! Usability is nice for me at least. I am totally onboard with merging this PR but with key undertaking that we either land other command soon after any important release we need this week or as part of one release with this item but different PR. Gentle fyi to @qpetraroia as well 

What do you think?

One observation I would share is to might be do the whole menu readjustment as one release so that we don't loose the other commands in play like TOP etc, 

Screenshot 2025-07-09 at 1 46 23 PM


So with the new changes the menu looks like this: 

Screenshot 2025-07-09 at 1 42 10 PM

Screenshot 2025-07-09 at 1 43 08 PM


Thanks once again.

PR #1525, opened 2025-07-30
Added option to port-forward KAITO workspace service

tejhan:
This PR introduces a new suggested feature that allows users to easily port-forward any KAITO workspace with a single click. On the KAITO manage page, deployed models will now have an option to initiate port-forwarding. The user will be prompted to specify a port, which will automatically launch an interactive shell with the port-forward process, streamlining the workflow. 

This eliminates the need for users to manually identify the relevant service port and start port-forwarding via the CLI.

.vsix for testing here -> [vscode-aks-tools-1.6.11-port1.vsix.zip](https://github.com/user-attachments/files/21516327/vscode-aks-tools-1.6.11-port1.vsix.zip)

### DIFF
--- a/src/commands/utils/tempfile.ts
+++ b/src/commands/utils/tempfile.ts
@@ -18,7 +18,7 @@ export async function withOptionalTempFile<T>(
 
 export async function createTempFile(content: string, extension: string): Promise<TempFile> {
     // TODO: try/catch and return errorable?
-    const tempFile = fileSync({ prefix: "aks-periscope-", postfix: `.${extension}` });
+    const tempFile = fileSync({ postfix: `.${extension}` });
     await fs.writeFile(tempFile.name, content);
     return new TempFile(tempFile);
 }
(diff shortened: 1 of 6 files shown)


src/commands/utils/tempfile.ts
```diff
@@ -18,7 +18,7 @@ export async function withOptionalTempFile<T>(
 
 export async function createTempFile(content: string, extension: string): Promise<TempFile> {
     // TODO: try/catch and return errorable?
-    const tempFile = fileSync({ prefix: "aks-periscope-", postfix: `.${extension}` });
+    const tempFile = fileSync({ postfix: `.${extension}` });
```
tejhan, line 21: Removed this periscope prefix - it seems this was previously used exclusively for periscope functionality, but usage has expanded for now, so we can remove this prefix. (Safe removal)

Tatsinnit: 💡 I think there is more to it then just this change hence for a long time this part of the code was not touched, I think if I recall correctly, the idea was to replace the `prefix` with more relevant string with `purpose` this is used for - which can then latter be used for debugging purposes and what file was used for what purposes.

tejhan: I see. I did notice that in many instances of temp file generation that would have the aks-periscope descriptor from an unrelated tool. 

I think that I can open a separate PR to to add an optional `prefix` argument to `createTempFile` and modify the instances where it's called in the code to pass in specific prefixes where needed.

Tatsinnit (approved):

PR #1564, opened 2025-08-15
Fix the client using sessionProvider instead of default.

Tatsinnit:
💡 This PR is to fix avoid using `DefaultAzureCredential` and user properly formed object via  `sessionProvider` .

Instead of using DefaultAzureCredential (which is designed for development/testing), the code now uses the existing VS Code authentication session through. Now it uses the `[ReadyAzureSessionProvider]` that manages the user's sign-in state.

Tested locally.

* **VSIX:**  [vscode-aks-tools-1.6.12-storage-test.vsix.zip](https://github.com/user-attachments/files/21801714/vscode-aks-tools-1.6.12-storage-test.vsix.zip)


tejhan (approved):
Tested & working well, lgtm

bosesuneha (approved):
tested, lgtm

PR #1689, opened 2025-10-21
Fix/dependbot weuui eslint

Tatsinnit:
Hiya, This PR will help in fixing: #1674 the new major update causes following:

❤️ Thanks and gentle FYI to cc: @tejhan , @bosesuneha 

```
Effects are intended to synchronize state between React and external systems such as manually updating the DOM, state management libraries, or other platform APIs. In general, the body of an effect should do one or both of the following:
* Update external systems with the latest state from React.
* Subscribe for updates from some external system, calling setState in a callback function when external state changes.

Calling setState synchronously within an effect body causes cascading renders that can hurt performance, and is not recommended. (https://react.dev/learn/you-might-not-need-an-effect).

```


tejhan (approved):
Tested branch locally & working well, LGTM.

PR #1818, opened 2026-01-08
REmove unwanted az call

Tatsinnit:
This PR get rid of az call which is no longer in need. other improvement thank you @ashu8912

❤️ Gentle fyi cc: @tejhan, @ashu8912 , @bosesuneha , @davidgamero or @gambtho


ashu8912 (approved):
LGTM!!

Issue #1847, opened 2026-01-27
Feature Request: Feature Flag for Controlled Rollout of New Capabilities


Tatsinnit:
### Feature Request: Feature Flag for Controlled Rollout of New Capabilities

**Summary**

Introduce a feature flag mechanism that allows new functionality to be merged and released continuously, without being visible or enabled for end users until the feature is fully ready.

**Problem Statement**

Today, larger changes often get bundled into major releases to avoid exposing incomplete or experimental functionality to users. This can:

* Increase release risk
* Slow down iteration
* Create long-lived branches or large PRs
* Make rollback and testing harder

**Proposed Solution**

Add a configurable feature flag (opt-in) that:

* Is disabled by default for all users
* Can be enabled explicitly via user settings (or internal config)
* Allows features to ship progressively behind the flag
* Can be removed once the feature is fully released and stable

**How This Helps**

* Enables **continuous merging** instead of large, risky releases
* Keeps main branches healthy and up to date
* Allows early validation by selected users or internal teams
* Reduces the need for long-running feature branches
* Makes rollouts safer and more predictable

**End State**

Once the feature is validated and ready for general availability:

* The feature flag is removed
* The capability becomes enabled by default
* No user action is required

**Additional Considerations**

* Flags should be clearly documented and time-bound
* Telemetry (if applicable) can help validate readiness
* Avoid flag sprawl by enforcing cleanup after GA

**Open Questions for Discussion**

* Should flags live in user settings, internal config, or both?
* How do we ensure flags are retired after release? (I am personally inclined to this and only for testing use the feature flag)

PR #1861, opened 2026-01-29
Feat: add aks app-deployment feature flag.

Tatsinnit:
## Summary of Changes

This PR intends to take care of #1847 ❤️ with this we have a feature flag which then can be maintained and we can continue to merge the container assis small implementation e2e to main without worrying releasing something which is available to end user. Further thoughts are welcome.

### 1. Configuration Setting (package.json)

Added `aks.containerAssistEnabledPreview` setting:
- **Type**: boolean
- **Default**: false
- **Description**: Controls Container Assist recommendations (Preview feature)

### 2. Getter Function (config.ts)

Added `getContainerAssistRecommendationsInfoState()` function that reads the configuration value.

## Usage will be as follows

Commands can now use this feature flag exactly like the existing similar functionality of `getAIRecommendationsInfoState()` , we have added `getContainerAssistRecommendationsInfoState` :

```typescript
import { getContainerAssistRecommendationsInfoState } from "../utils/config";

// In your command
const containerAssistFlag = getContainerAssistRecommendationsInfoState();
if (!containerAssistFlag) {
    vscode.window.showWarningMessage("Container Assist is disabled. Enable it in settings.");
    return;
}

// Proceed with container assist features
```

The implementation follows the same pattern as the existing copilot feature flag we have against the repo, making it consistent with the codebase. We can enable/disable it via VS Code settings: `aks.containerAssistEnabledPreview`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -174,7 +174,7 @@
                 "aks.containerAssistEnabledPreview": {
                     "type": "boolean",
                     "default": false,
-                    "description": "Set to true to enable Container Assist recommendations. (Preview feature)"
+                    "description": "Set to true to enable deployment related file generation capability using AI tool. (Preview feature)"
                 },
                 "aks.aksmcpserver.releaseTag": {
                     "type": "string",


package.json
```diff
@@ -171,6 +171,11 @@
                     "default": true,
                     "description": "Set to true to enable GH Copilot hook. (Preview feature)"
                 },
+                "aks.containerAssistEnabledPreview": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Set to true to enable Container Assist recommendations. (Preview feature)"
```
bosesuneha:
```suggestion
                    "description": "Set to true to enable deployment related file generation capability using AI tool. (Preview feature)"
```

PR #1914, opened 2026-02-23
Enhance namespace selection with type categorization

Tatsinnit:
This PR enhances the namespace selection experience by adding type categorization to Kubernetes namespaces in QuickPick selections. Users can now see whether a namespace is a System, Managed, or user namespace, making it easier to understand the purpose and origin of each namespace.

## What Changed

- **New Interface**: Added `NamespaceWithType` interface to represent namespaces with type classification
- **Enhanced Function**: Added `getClusterNamespacesWithTypes()` to retrieve namespaces with their types and metadata
- **Namespace Classification**: Implemented helper functions to categorize namespaces as:
  - **System**: Core Kubernetes namespaces (`kube-system`, `default`, etc.)
  - **Managed**: Azure-managed or operator-managed namespaces (identified by labels like `app.kubernetes.io/managed-by`)
  - **User**: User-created namespaces, including test namespaces
- **AKS Desktop Support**: Added detection for AKS Desktop managed namespaces via `headlamp.dev/project-managed-by=aks-desktop` label
- **Container Assist Integration**: Updated Container Assist to use typed namespace selections with descriptive labels
- **Bug Fix**: Resolved `localeCompare` error by using correct property names in sort operations

## Why This Change?

Previously, namespace selections showed only names without context about their purpose or management status. This made it difficult for users to:

- Distinguish between system and user namespaces
- Identify managed services and their namespaces
- Understand which namespaces are safe to modify
- Locate test or development namespaces

## Benefits

- **Improved UX**: QuickPick selections now show descriptive labels (e.g., "System namespace", "Managed namespace")
- **Better Decision Making**: Users can make informed choices about which namespaces to use
- **Reduced Errors**: Clear indication of system/managed namespaces helps prevent accidental modifications
- **Enhanced Discoverability**: Test namespaces are identified by name patterns and labels

## Technical Details

- Uses `kubectl get namespace -o json` to retrieve full namespace metadata
- Implements pattern matching for test namespaces (`test-*`, `*-test`, `e2e-*`, etc.)
- Supports various managed service detection patterns including Azure Workload Identity
- Maintains backward compatibility with existing namespace functions

## Screenshots

**Before** 

Screenshot 2026-02-23 at 12 23 22 PM

**After**

Screenshot 2026-02-23 at 1 06 20 PM

### DIFF
--- a/src/commands/utils/clusters.ts
+++ b/src/commands/utils/clusters.ts
@@ -630,7 +630,10 @@ export async function getClusterNamespacesWithTypes(
                     },
                 );
             } catch (e) {
-                throw new Error(`Failed to parse namespace JSON: ${e}`);
+                return {
+                    succeeded: false,
+                    error: `Failed to parse namespace JSON: ${clusterName}: ${getErrorMessage(e)}`,
+                };
             }
         });
     });


src/commands/utils/clusters.ts
```diff
@@ -568,6 +576,100 @@ export async function getClusterNamespaces(
...
+}
+
+function isSystemNamespace(name: string): boolean {
+    const systemNamespaces = ["kube-system", "kube-public", "kube-node-lease", "default", "kube-lease"];
+    return systemNamespaces.includes(name);
+}
+
+function isManagedNamespace(labels: { [key: string]: string }, annotations: { [key: string]: string }): boolean {
+    // Check for common managed service labels and annotations
+    const managedLabels = [
```
bosesuneha: I'm wondering if we're showing all managed namespaces, should we use the sdk api for that? and on top of that we can check for the aks-desktop labels if we want to show them as aks-desktop projects.

bosesuneha: leaving this for future changes, for showing the managed ns we can use the sdk call for `managednamespaces` and then specifically for aks-desktop project we can use the label matching logic. resolving this for now.

bosesuneha (approved):

PR #1940, opened 2026-02-27
Harden error handling for setGitHubActionsSecrets

bosesuneha:
## Summary

Improves error handling in `setGitHubActionsSecrets` so that every failure path shows a single, actionable error dialog with "Copy Secrets" / "View Output" fallback button to ensure users can always retrieve their secret values even when the API call fails.

### Error handling (`oidcSetup.ts`)

- **SAML SSO detection**: Parse the `X-GitHub-SSO` response header on 403 errors from both `repos.get` and `getRepoPublicKey`. Show an "Authorize Token" button that opens the SSO authorization URL.
- **Archived repo check**: Detect archived repositories and show a clear error instead of failing silently on the secrets API.
- **Single error dialog**: Replace the previous pattern where both `setGitHubActionsSecrets` and `displayOIDCResults` could each show an error dialog. Now only one dialog is shown per failure, with fallback buttons.
- **Boolean return type**: Changed from `Promise<void>` to `Promise<boolean>` so callers know whether secrets were set successfully.
- **`createIfNone` auth**: Switched from `forceNewSession` to `createIfNone: true` to silently reuse existing GitHub sessions with sufficient scope.
- **Removed "Contact Admin" / "Open Repo Settings" buttons**: The `/settings/access` URL 404s for non-admins, so the advice is now included in the error message text instead.
- **Extracted `handleReposGetError`**: Keeps the main function readable by handling all `repos.get` error cases in a dedicated function.
- **Exported helpers**: `createOctokitClient`, `isOctokitError`, `encryptSecret`, `getSAMLSSOUrl` are now exported for direct unit testing.

### DIFF
--- a/src/tests/suite/containerAssist/tools.test.ts
+++ b/src/tests/suite/containerAssist/tools.test.ts
@@ -164,7 +164,8 @@ describe("tools", () => {
         });
     });
 
-    describe("handleListDirectory", () => {
+    describe("handleListDirectory", function () {
+        this.timeout(10_000);
         it("rejects path traversal", async () => {
             const result = await handleListDirectory({ path: "../../" }, tempDir);
             assert.ok(result.includes("Refused"));
@@ -200,10 +201,12 @@ describe("tools", () => {
         });
 
         it("caps maxDepth at hard limit", async () => {
-            const resultAtCap = await handleListDirectory({ path: ".", maxDepth: 3 }, tempDir);
-            const resultOverCap = await handleListDirectory({ path: ".", maxDepth: 100 }, tempDir);
+            const [resultAtCap, resultOverCap] = await Promise.all([
+                handleListDirectory({ path: ".", maxDepth: 3 }, tempDir),
+                handleListDirectory({ path: ".", maxDepth: 100 }, tempDir),
+            ]);
             assert.strictEqual(resultAtCap, resultOverCap);
-        });
+        }).timeout(10_000);
     });
 
     describe("handleToolCall", () => {


src/commands/aksContainerAssist/oidcSetup.ts
```diff
@@ -498,104 +498,118 @@ AZURE_SUBSCRIPTION_ID: ${result.subscriptionId}`;
...
+     * Shows an SSO error with "Authorize Token" plus fallback buttons.
+     * Opens the SSO authorization URL if the user clicks "Authorize Token".
+     */
+    async function showSSOError(message: string, ssoUrl: string): Promise<void> {
+        const authorize = l10n.t("Authorize Token");
+        const copySecrets = l10n.t("Copy Secrets");
+        const viewOutput = l10n.t("View Output");
+        const selection = await vscode.window.showErrorMessage(message, authorize, copySecrets, viewOutput);
+        if (selection === authorize) {
+            vscode.env.openExternal(vscode.Uri.parse(ssoUrl));
```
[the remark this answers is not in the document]
Tatsinnit: Please note we have sanitisation function for this, let's use that, there is one PR where I used that. (SO what is this ssoURL looks like? please. thanks!

src/commands/aksContainerAssist/oidcSetup.ts
```diff
@@ -646,56 +677,131 @@ export async function setGitHubActionsSecrets(
...
+async function handleReposGetError(
+    error: unknown,
+    owner: string,
+    repo: string,
+    showError: (message: string) => Promise<void>,
+    showSSOError: (message: string, ssoUrl: string) => Promise<void>,
+): Promise<void> {
+    const statusCode = isOctokitError(error) ? error.status : undefined;
+
+    if (statusCode === 401) {
```
Tatsinnit: 💡 Nice, thanks for moving this here.

can we make tiny adjustment please:

Using `switch` to make it more structured et. al. 

```*.ts
switch (statusCode) {
    case 401:
        await showError(
            l10n.t("GitHub authentication token is invalid or expired. Please try again to re-authenticate."),
        );
        return;

    case 403: {
        const ssoUrl = getSAMLSSOUrl(error);
        // handle SSO case here
        break;
    }
}

```

src/commands/aksContainerAssist/oidcSetup.ts
```diff
@@ -498,104 +498,118 @@ AZURE_SUBSCRIPTION_ID: ${result.subscriptionId}`;
...
+ * Returns `true` if all secrets were set successfully, `false` otherwise.
+ * On failure, shows an error dialog with "Copy Secrets" / "View Output"
+ * fallback buttons so the user can still retrieve their secret values.
  */
 export async function setGitHubActionsSecrets(
     owner: string,
     repo: string,
     secrets: Record<string, string>,
-): Promise<void> {
+): Promise<boolean> {
```
Tatsinnit: 💡 Question please: We changed the return type signature of this function because of specific use case? Thanks.

PR #1953, opened 2026-03-04
CA: namespace selection, exit confirmation, and LM availability check

bosesuneha:
**Namespace always prompted from cluster tree (azureSelections.ts)**
- When invoking the command from the AKS cluster tree with "Generate Deployment Files only", namespace was silently skipped and defaulted to "default" in generated manifests. It is now always prompted, consistent with the file explorer entry point.
- Switched the deployment-only path from selectAcr (subscription-wide, no cluster context) to selectClusterAcr (filters to ACRs already attached to the cluster).
- Removed the now-dead selectAcr export.
- Extracted a shared collectAzureContextForCluster helper to eliminate duplicated ACR → namespace → workflow name logic between collectAzureContext and collectAzureContextFromTree.

**Exit confirmation on all wizard steps (aksContainerAssist.ts)**
- The action picker ("Generate Deployment Files" / "Generate GitHub Workflow") and the workspace folder picker (cluster tree entry point) were missing exit confirmation. Cancelling either would silently drop the user out of the flow. Both now show the "Go Back / Exit Container Assist" modal, consistent with all other wizard steps.
Early LM availability check (aksContainerAssist.ts)
- When "Generate Deployment Files" is selected, selectModel(false) is now called immediately — before the Azure context prompts — so a missing Copilot model is surfaced with a clear error right away rather than failing silently deep in the generation flow.

**Bugfix in preferred model fallback warning (lmClient.ts)**
- The condition if (selectedModel !== preferredModels[0] && preferredModels.length === 0) was always false (if preferredModels is empty, preferredModels[0] is undefined and selectedModel would be allModels[0], never undefined). Simplified to if (preferredModels.length === 0).

### DIFF
--- a/src/commands/aksContainerAssist/aksContainerAssist.ts
+++ b/src/commands/aksContainerAssist/aksContainerAssist.ts
@@ -124,7 +124,7 @@ export async function runContainerAssistFromTree(_context: IActionContext, targe
  * Prompts the user to pick a workspace folder if multiple are open,
  * or uses the single workspace folder automatically.
  */
-async function pickWorkspaceFolder(): Promise<vscode.WorkspaceFolder | undefined> {
+export async function pickWorkspaceFolder(): Promise<vscode.WorkspaceFolder | undefined> {
     const folders = vscode.workspace.workspaceFolders;
     if (!folders || folders.length === 0) {
         vscode.window.showErrorMessage(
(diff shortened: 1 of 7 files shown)


Tatsinnit (approved):

src/commands/aksContainerAssist/azureSelections.ts
```diff
@@ -410,45 +410,8 @@ export async function promptForWorkflowName(appName: string): Promise<string | u
     return workflowName;
 }
 
-/**
- * Prompts user to select an ACR from the subscription (no cluster required).
- * Returns the AzureResource for the selected ACR.
- */
-export async function selectAcr(
```
Tatsinnit: 💡 this function is moved to another file for implementation or removal is for specific reasons

Tatsinnit: I guess this part: selectClusterAcr (filters to ACRs already attached to the cluster). Is know understanding form PM then. Making sure I double checked ❤️💡

PR #1960, opened 2026-03-06
Update isManagedNamespace function

bosesuneha:
This pull request simplifies the logic used to determine if a Kubernetes namespace is "managed" within the `clusters.ts` utility. The main change is narrowing the criteria for managed namespaces to a single label check, which streamlines the code and reduces potential sources of error.

Namespace classification logic simplification:

* The `isManagedNamespace` function now only checks if the `labels["kubernetes.azure.com/managedByArm"]` is set to `"true"` instead of evaluating multiple labels and annotations.
* The call to `isManagedNamespace` in `getClusterNamespacesWithTypes` was updated to pass only the `labels` argument, removing the unnecessary `annotations` parameter.

### DIFF
--- a/src/commands/utils/clusters.ts
+++ b/src/commands/utils/clusters.ts
@@ -615,7 +615,7 @@ export async function getClusterNamespacesWithTypes(
                         let type: "system" | "user" | "managed" | "unknown";
                         if (isSystemNamespace(name)) {
                             type = "system";
-                        } else if (isManagedNamespace(labels, annotations)) {
+                        } else if (isManagedNamespace(labels)) {
                             type = "managed";
                         } else {
                             type = "user";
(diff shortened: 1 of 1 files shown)


src/commands/utils/clusters.ts
```diff
@@ -644,30 +644,8 @@ function isSystemNamespace(name: string): boolean {
...
-    ];
-
-    return (
-        managedLabels.some((label) => labels[label]) ||
-        managedAnnotations.some((annotation) => annotations[annotation]) ||
-        // Check for Azure-specific managed namespaces
-        labels["app.kubernetes.io/component"] === "azure-workload-identity" ||
-        labels["control-plane"] === "controller-manager" ||
-        // Check for AKS Desktop managed namespaces
-        labels["headlamp.dev/project-managed-by"] === "aks-desktop"
```
Tatsinnit: 💡 Thank you so much for this, I wonder if we check that this does not negate the label logic for `aks-desktop`? can we check please <3

PR #2069, opened 2026-04-17
feat: split tree-view Container Assist into discrete cluster actions

bosesuneha:
- Replaces the single `aks.runContainerAssistFromTree` tree-view entry with three discrete actions in both the legacy and simplified cluster context menus: **Migrate Application to AKS**, **Generate Dockerfiles and K8s Manifests for App**, and **Deploy App with Automated Pipeline**. Each tree-variant command pre-selects its corresponding `ContainerAssistAction`, removing the extra multi-select picker step.
- Renames the `containerizationApp` command to `containerizeApp` for naming consistency and fixes the previously broken manifest state where `aks.runContainerAssistFromTree` was still declared in `package.json` but no longer registered in `extension.ts`.
- Renames the user-visible "Draft Validate" strings to "Run Deployment Safeguards" so the webview panel title and heading match the already-renamed `aks.aksDraftValidate` command title.
- Reorganizes `appModernizationBridge.ts` so it only contains the actual bridge to the external app-modernization extension; the four container-assist command adapters now live next to the `runContainerAssist` dispatcher they wrap.

### DIFF
--- a/l10n/bundle.l10n.json
+++ b/l10n/bundle.l10n.json
@@ -419,7 +419,7 @@
   "Attach ACR to Cluster": "Attach ACR to Cluster",
   "Draft GitHub Workflow in {0}": "Draft GitHub Workflow in {0}",
   "GitHub repository {0}/{1} not found.": "GitHub repository {0}/{1} not found.",
-  "Draft Validate in {0}": "Draft Validate in {0}",
+  "Run Deployment Safeguards in {0}": "Run Deployment Safeguards in {0}",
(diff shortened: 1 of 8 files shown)


Tatsinnit (dismissed):

Tatsinnit:
💡 Just need `verified` comment please, aka `git commit -s -S -m "some helpful message"` of preferred thing. ❤️

Tatsinnit (approved):

PR #2071, opened 2026-04-20
fix(containerAssist): detect manifests via module paths in workflow generator

bosesuneha:
This pull request enhances the detection of Kubernetes manifests for workflow generation by introducing a module-aware scanning approach. The new logic attempts to find manifests within each detected module's manifest folder, improving alignment with where deployment generation writes these files. If no manifests are found or an error occurs, the system gracefully falls back to scanning the workspace root.

The most important changes are:

**Kubernetes Manifest Detection Improvements:**

* Added a new `detectK8sManifests` function that first analyzes the repository to find module paths and scans each module's manifest folder for Kubernetes manifests, falling back to a workspace-root scan if necessary. (Fc28be85L358)
* Updated the manifest detection call in `collectWorkflowConfiguration` to use the new `detectK8sManifests` function, enabling module-aware detection logic.

**Dependency and Import Updates:**

* Imported `getK8sManifestFolder` and `ContainerAssistService` to support the new detection logic.

### DIFF
--- a/src/commands/aksContainerAssist/workflowGenerator.ts
+++ b/src/commands/aksContainerAssist/workflowGenerator.ts
@@ -166,8 +166,13 @@ async function collectWorkflowConfiguration(
             relativeManifests,
         );
     } else {
-        const detectedManifests = await detectK8sManifests(workspaceRoot, projectRoot);
-        relativeManifests = detectedManifests.map((p) => toPosixPath(path.relative(workspaceRoot, p)));
+        const detection = await detectK8sManifests(workspaceRoot, projectRoot);
+        if (!detection.succeeded) {
+            logger.error("Manifest detection failed", detection.error);
+            void vscode.window.showErrorMessage(l10n.t("Failed to detect Kubernetes manifests: {0}", detection.error));
+            return undefined;
+        }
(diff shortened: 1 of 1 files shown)


Tatsinnit (dismissed):

src/commands/aksContainerAssist/workflowGenerator.ts
```diff
@@ -356,6 +358,41 @@ function formatManifestPathForYamlBlock(manifests: string[]): string {
...
+                const hits = await scanForK8sManifests(root);
+                hits.forEach((p) => found.add(p));
+            }
+
+            if (found.size > 0) {
+                return Array.from(found);
+            }
+        }
+    } catch (error) {
+        logger.error("Module-aware manifest detection failed; falling back to workspace-root scan", error);
```
Tatsinnit: 💡Thought share and idea: If this is legitimate error which user should see it for any further action then we should bumble it for the user using notification, logger is a silent thing for devs to debug. 

If this is meant for end user we possibly need to record for more clarity.

bosesuneha: Thanks, propagtaing the error to the user now.

Tatsinnit (approved):

Issue #2109, opened 2026-05-01
Add helper to scan manifests per module path in `fileOperations.ts`


Tatsinnit:
## Description

Add a function in `fileOperations.ts` that, given a module directory path, scans `<moduleDir>/k8s/` for `.yaml` and `.yml` files and returns their paths. This helper is used by the per-service configuration loop in multi-container mode to auto-detect manifests for each selected Dockerfile's module.
- **Labels:** `enhancement`, `container-assist`
- **Acceptance Criteria:**
  - [ ] Function accepts a module directory path and workspace root
  - [ ] Scans `<moduleDir>/k8s/` for `.yaml` and `.yml` files
  - [ ] Returns array of relative paths (empty array if directory doesn't exist)
  - [ ] Handles edge cases: no k8s directory, empty k8s directory, nested manifests
- **Testing Plan (estimated — adjust based on current test setup):**
  - [ ] **Unit:** Create temp directory with `frontend/k8s/deployment.yaml` and `frontend/k8s/service.yml` → assert both files returned
  - [ ] **Unit:** Create temp directory with empty `frontend/k8s/` → assert empty array returned
  - [ ] **Unit:** Create temp directory with no `frontend/k8s/` directory → assert empty array returned (no error thrown)
  - [ ] **Unit:** Create temp directory with `frontend/k8s/subfolder/nested.yaml` → assert nested file is **not** returned (flat scan only) or **is** returned (document which behavior is expected)
  - [ ] **Unit:** Create temp directory with `frontend/k8s/readme.md` and `frontend/k8s/deploy.yaml` → assert only `.yaml`/`.yml` files returned
  - [ ] **Unit:** Verify returned paths are relative to workspace root
  - [ ] **Unit:** Verify function handles symlinks gracefully (no infinite loops)
  - [ ] **Edge case:** Module directory path contains spaces (e.g., `my service/k8s/`) → assert correct paths returned

PR #2126, opened 2026-05-07
refactor: consolidate filesystem scanners (#2109)

bosesuneha:
## Summary

Consolidates 5 filesystem scanning functions into a single generic `walkDirectory` helper with two focused callers (`scanForDockerfiles`, `scanForK8sManifests`), and fixes the DOCKER_FILE path resolution in generated workflows.

### fileOperations.ts
- **Replace** `scanForDockerfilesRecursive`, `scanDirectory`, `scanDirectoryRecursive`, and two separate scanner implementations with a single `walkDirectory` helper
- **Unify** `EXCLUDED_DIRS` set — add Java-specific entries (`.gradle`, `.idea`, `.settings`, `.mvn`)
- **Extract** `SCAN_MAX_DEPTH` constant (depth 3 for both scanners)
- **Use bitwise** `FileType` checks (`!== 0`) so symlinked files/directories are correctly traversed
- **Add CRLF handling** in K8s manifest `kind:` regex (`\r?` before `$`)
- **Remove** unused debug logging and the `scanManifestsForModule` wrapper (zero callers)

### workflowGenerator.ts
- **Fix DOCKER_FILE path**: resolve relative to build context, not workspace root — `az acr build -f` expects this

## Testing
- `npx tsc --noEmit` — clean
- `npm run webpack` — compiles successfully

Contributes to #2109

### DIFF
--- a/src/commands/aksContainerAssist/fileOperations.ts
+++ b/src/commands/aksContainerAssist/fileOperations.ts
@@ -55,23 +55,8 @@ export async function checkExistingFiles(folderPath: string): Promise<ExistingFi
     return result;
 }
 
-/**
- * Searches recursively (up to maxDepth levels) for Dockerfiles under rootPath.
- * Returns absolute paths sorted shallowest-first; excluded dirs are skipped.
- */
-export async function scanForDockerfiles(rootPath: string, maxDepth: number = 3): Promise<string[]> {
-    const found: string[] = [];
-    await scanForDockerfilesRecursive(rootPath, found, maxDepth, 0);
-    // Sort shallowest first
-    found.sort((a, b) => {
-        const aDepth = a.split(path.sep).length;
-        const bDepth = b.split(path.sep).length;
-        return aDepth - bDepth;
-    });
-    return found;
-}
-
-const DOCKERFILE_EXCLUDED_DIRS = new Set([
+/** Directories to skip during recursive scans (build artifacts, dependencies, etc.) */
+const EXCLUDED_DIRS = new Set([
     "node_modules",
     ".git",
     "dist",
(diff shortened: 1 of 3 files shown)


src/commands/aksContainerAssist/fileOperations.ts
```diff
@@ -55,23 +55,8 @@ export async function checkExistingFiles(folderPath: string): Promise<ExistingFi
...
-        const aDepth = a.split(path.sep).length;
-        const bDepth = b.split(path.sep).length;
-        return aDepth - bDepth;
-    });
-    return found;
-}
-
-const DOCKERFILE_EXCLUDED_DIRS = new Set([
+/** Directories to skip during recursive scans (build artifacts, dependencies, etc.) */
+const EXCLUDED_DIRS = new Set([
```
Tatsinnit, line 59: 💡 Nice, thank you for these changes, I wonder if worth to add the reason of rename of this `const` from `DOCKERFILE_EXCLUDED_DIRS` vs `EXCLUDED_DIRS`? I guess mainly thinking if the larger context was there on previous name.  Might be add as comment and we remove it ad we progress?

src/commands/aksContainerAssist/fileOperations.ts
```diff
@@ -89,139 +74,75 @@ const DOCKERFILE_EXCLUDED_DIRS = new Set([
...
+export async function scanForDockerfiles(rootPath: string): Promise<string[]> {
+    const found: string[] = [];
+    await walkDirectory(rootPath, SCAN_MAX_DEPTH, 0, (filePath, name) => {
+        if (name === "Dockerfile") {
+            found.push(filePath);
+        }
+    });
+    // Sort shallowest first
+    found.sort((a, b) => a.split(path.sep).length - b.split(path.sep).length);
+    return found;
```
Tatsinnit:
```suggestion
export async function scanForDockerfiles(rootPath: string): Promise<string[]> {
    const dockerfiles: string[] = [];

    const isDockerfile = (_filePath: string, fileName: string): boolean =>
        fileName === "Dockerfile";

    await walkDirectory(
        rootPath,
        SCAN_MAX_DEPTH,
        0,
        (filePath, fileName) => {
            if (isDockerfile(filePath, fileName)) {
                dockerfiles.push(filePath);
            }
        }
    );

    return dockerfiles.sort(
        (left, right) =>
            left.split(path.sep).length - right.split(path.sep).length
    );
}
```

src/commands/aksContainerAssist/fileOperations.ts
```diff
@@ -89,139 +74,75 @@ const DOCKERFILE_EXCLUDED_DIRS = new Set([
...
-                    manifestSet.add(fullPath);
-                }
-            });
-
-        await Promise.all(fileChecks);
-    } catch {
-        // Directory doesn't exist or can't be read - skip silently
-    }
+    });
+    return manifests;
```
Tatsinnit, line 124: how about this

```suggestion
export async function scanForK8sManifests(rootPath: string): Promise<string[]> {
    const manifests: string[] = [];

    const isYamlFile = (fileName: string): boolean =>
        fileName.endsWith(".yaml") || fileName.endsWith(".yml");

    await walkDirectory(
        rootPath,
        SCAN_MAX_DEPTH,
        0,
        async (filePath, fileName) => {
            if (!isYamlFile(fileName)) {
                return;
            }

            if (await isKubernetesManifest(filePath)) {
                manifests.push(filePath);
            }
        }
    );

    return manifests;
}
```

Tatsinnit (dismissed):

src/commands/aksContainerAssist/fileOperations.ts
bosesuneha: It was just misnamed earlier. We were/are using it for scannning both Dockerfiles and Manifests, so renamed it.

src/commands/aksContainerAssist/fileOperations.ts
bosesuneha: updated the code

src/commands/aksContainerAssist/fileOperations.ts
bosesuneha: updated the code

Tatsinnit (approved):

PR #2144, opened 2026-05-14
Kickstart enhancements

bosesuneha:
fix(kickstart): correct run_in_terminal tool invocation — was calling "runInTerminal" with {command, cwd}; the real tool is "run_in_terminal" with {command, explanation, goal, mode} (no cwd). Probe vscode.lm.tools first and prepend cd <cwd> && to the command.

feat(kickstart): clone sample repos into existing workspace folder — in vscode.dev / Cloud Shell, samples must land on the Cloud-Shell-backed folder, not os.tmpdir(). useSample() now takes an optional parentPath and the command passes the open workspace folder when one exists.

feat(kickstart): support monorepos with per-module Dockerfile and k8s manifests

Stage per-module at <moduleRel>/Dockerfile and <moduleRel>/k8s/<file> via a new moduleStagePrefix() helper; single-module projects keep the flat layout.
ExistingFilesCheckResult.dockerfilePaths: string[] (was a single path); AnalysisData carries existing Dockerfile / manifest paths.
Build phase: one target per Dockerfile with sanitized image names, workspace as build context, --file for staged Dockerfiles, --subscription on az acr calls.
Deploy phase: collects every <module>/k8s/ dir and applies them all with multiple -f flags in one kubectl apply.
Generic buildNestedFileTree() for chat file-tree rendering of nested paths.
Resolve modulePath to absolute before SDK calls. Tests updated.
feat(kickstart): enhance existing Dockerfiles and k8s manifests instead of overwriting

prepare.ts reads existing files into Map<modulePath, …> via longest-prefix match (pickModuleForRelDir) so each module only sees its own.
Streams a "🔄 Enhancing N existing …" message when applicable.
System prompts gain an ENHANCEMENT MODE section: preserve base image family, custom RUN/COPY/ENV, metadata, env vars, volumes, probes; improve only missing best practices (multi-stage, non-root, HEALTHCHECK, securityContext, resources, image refs); return the COMPLETE artifact, not a diff.
buildDockerfileUserPrompt / buildK8sManifestUserPrompt accept the existing artifacts and append them as guidance.

### DIFF
--- a/src/chatParticipants/kickstart/orchestrator.ts
+++ b/src/chatParticipants/kickstart/orchestrator.ts
@@ -182,19 +182,8 @@ export async function handleStart(
 
     stream.markdown("\n✅ **Files generated** — review in the panel, then click **Save to project**:\n\n");
 
-    // Build file tree for native chat rendering
-    const fileTree: vscode.ChatResponseFileTree[] = [];
-    const k8sChildren: vscode.ChatResponseFileTree[] = [];
-    for (const sf of stagedSoFar) {
-        if (sf.filename.startsWith("k8s/")) {
-            k8sChildren.push({ name: sf.filename.replace("k8s/", "") });
-        } else {
-            fileTree.push({ name: sf.filename });
-        }
-    }
-    if (k8sChildren.length > 0) {
-        fileTree.push({ name: "k8s", children: k8sChildren });
-    }
+    // Build file tree for native chat rendering (supports nested per-module paths).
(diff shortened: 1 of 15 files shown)


Tatsinnit (approved):
❤️ Thank you, this is a demo in progress and beta work, in order to expedite and this merge is not going to `main` directly, I think we can get this in, gently fyi to @davidgamero  for eyes.

davidgamero (approved):
lgtm

PR #2165, opened 2026-05-20
Kickstart: pre-flight check for user deploy RBAC (cluster + ACR)

bosesuneha:
## Summary

Adds a centralized pre-flight check that verifies the **signed-in user** holds the Azure role assignments required for a successful Kickstart deploy on the Normal Namespace path, and surfaces the result as a markdown table in the CONFIGURE phase.

**Cluster scope** (only when Azure RBAC for Kubernetes is enabled — always on for AKS Automatic):
- AKS RBAC Writer / RBAC Admin / RBAC Cluster Admin → required for `kubectl apply`

**ACR scope:**
- AcrPush → required for `docker push` during BUILD
- Container Registry Tasks Contributor → required for `az acr build`

The check uses `roleAssignments.listForScope` with `atScope()`, so RG / subscription / management-group-inherited assignments count.

## Behavior

- **CONFIGURE phase** — displays a single pre-flight table with five always-rendered checks (Cluster SKU, Kubeconfig access, ACR pull (cluster→registry), ACR push (you→registry), ACR Tasks (you→registry)) plus the AKS RBAC row when `azureRbacEnabled` is true.
- **DEPLOY phase** — blocks only when the user definitively lacks an AKS RBAC Writer-tier role on an Azure-RBAC-enabled cluster. All other misses (ACR roles, kubeconfig) are display-only; they fail naturally with a clearer 403 in BUILD if missing.
- **403 on enumeration** → `*Inconclusive` flag is set → warning rendered, no block.

## Out of scope (deferred to follow-up PR)

Self-assign / fix-it commands for missing roles. The check results carry all the data needed; this PR is checks-only.

## Files

- **New** `src/commands/utils/aksRbacHelpers.ts` — role IDs, `isAzureRbacEnabled`, `isAuthorizationError`, `checkUserDeployRbac`
- `src/commands/aksContainerAssist/oidcSetup.ts` — dedup constants, delegate to shared helper
- `src/commands/aksKickstart/configure.ts` — fetch cluster once, parallel checks, `userDeployRbac` on `KickstartConfiguration`
- `src/chatParticipants/kickstart/state.ts` — `ConfigData` extended with RBAC fields
- `src/chatParticipants/kickstart/phases/configure.ts` — markdown-table renderer
- `src/chatParticipants/kickstart/phases/deploy.ts` — pre-deploy RBAC gate

## Testing

- `tsc --noEmit` clean
- `eslint` clean on all touched files
[...]

PR #2181, opened 2026-05-26
kickstart: feat/kickstart rbac checks

davidgamero:
continuing work from https://github.com/Azure/vscode-aks-tools/pull/2165


davidgamero:
@Tatsinnit for visibility

PR #2222, opened 2026-06-10
feat(argo): enable Argo CD feature flag by default and update related docs

Tatsinnit:
## Summary

This PR enables the existing Argo CD feature flag by default so users can discover and use Argo CD GitOps commands out of the box, while still allowing opt-out via settings.

## Changes

1. Changed the default value of the Argo CD feature flag from false to true.
2. Updated activation-time config fallback to align with the new default.
3. Updated documentation to reflect the new behavior and how to disable the feature if needed.

## Why
Argo CD functionality was already feature-gated, but defaulting to disabled made it easy to miss. Enabling it by default improves discoverability and reduces setup friction, while preserving control for users who want to hide it.

## User Impact

- New default: Argo CD commands are visible/enabled by default.
- Users can disable them with:
  - aks.argoCDEnabled = false
- Reload is still required after changing the setting.

## Validation

1. Fresh install / default settings: Argo CD commands appear.
2. Set aks.argoCDEnabled to false and reload: Argo CD commands are hidden.
3. Re-enable and reload: commands return as expected.

## Risk

Low risk. Behavior is controlled by an existing feature flag and remains fully reversible through settings.

### DIFF
--- a/docs/book/src/features/argocd-gitops-integration.md
+++ b/docs/book/src/features/argocd-gitops-integration.md
@@ -9,15 +9,17 @@ The Argo CD integration brings a complete GitOps workflow to AKS clusters direct
 
 ## Feature Flag
 
-All Argo CD commands are gated behind a feature flag and **disabled by default**. To enable:
+All Argo CD commands are gated behind a feature flag and **enabled by default**.
+
+If you want to hide Argo CD commands, set:
 
 ```json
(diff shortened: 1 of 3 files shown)


tejhan (approved):

bosesuneha (approved):
lgtm

PR #2226, opened 2026-06-11
Updating marketplace display name and description for discoverability

tejhan:
***Description***

This PR updates the `displayName` and `description` to improve discoverability of the extension, primarily for the 'AKS' query. Currently when searching up 'Aks', our extension ranks #9 even though we have over 10x the downloads of the second most popular result, which funnily enough is our deprecated older version. The biggest factor is most likely the lack of AKS in the `displayName` which is a highly weighed factor for search results. We can see that the higher ranking results include the `Aks` substring even if they aren't aks related. 

***Changes***

- Adds 'AKS' to the display name (`Azure Kubernetes Service (AKS)`) so users searching the marketplace for 'AKS' get a direct hit.
- Replaces the generic description with a more actionable one that highlights core capabilities: create, manage, deploy, and diagnose.

***Image of search results for the 'aks' query:***

Screenshot 2026-06-08 141351


package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "vscode-aks-tools",
-    "displayName": "Azure Kubernetes Service",
-    "description": "Display Azure Kubernetes Services within VS Code",
+    "displayName": "Azure Kubernetes Service (AKS)",
+    "description": "Create, manage, deploy, and diagnose AKS clusters directly from VS Code",
```
Tatsinnit, line 4: 💡 thank you for this!! we should find out what will be the affect of this change will this have any breaking impact. (Generic fyi)

bosesuneha (dismissed):
lgtm

Tatsinnit (dismissed):

Tatsinnit (approved):

PR #2259, opened 2026-06-23
fix: handle `err.code` as string | number in `shell.ts`

Tatsinnit:
### Summary

This PR updates `@types/node` from `^25.7.0` to `^26.0.0` and fixes the resulting TypeScript compilation error in the webpack build. Gentle FYI ❤️  @tejhan and @bosesuneha + @davidgamero 

[Fixes](https://github.com/Azure/vscode-aks-tools/pull/2256)

#### Dependency Update
- **@types/node**: `^25.7.0` → `^26.0.0`
  - Updates to latest Node.js type definitions

#### Code Fix
- **src/commands/utils/shell.ts**: Fixed type incompatibility with `err.code`
  - In `@types/node@26`, the `ExecException.code` property type changed to `string | number` (was previously just `number`)
  - Added type guard to safely handle both types:
    ```typescript
    const code = typeof err.code === "number" ? err.code : parseInt(String(err.code), 10) || 1;
    ```

### 🐛 Issue Fixed

**Before**: Webpack compilation failed with:
```
ERROR in shell.ts(58,27)
TS2322: Type 'string | number' is not assignable to type 'number'.
  Type 'string' is not assignable to type 'number'.
```

**After**: ✅ Webpack compiles successfully

### ✅ Verification

- ✅ `npm run webpack`: Compiles successfully in production mode
- ✅ Type safety maintained with runtime type checking
- ✅ Fallback to exit code `1` if parsing fails

### 📝 Implementation Details

The fix uses a defensive approach to handle the `err.code` property:
1. If `err.code` is already a number, use it directly
2. If it's a string, parse it to an integer using `parseInt()`
3. If parsing fails (returns `NaN`), default to exit code `1`

This ensures backward compatibility with both old and new versions of `@types/node` and handles edge cases gracefully.

### 🔗 Related

- Updates Node.js type definitions to match Node.js 26.x LTS
- Maintains compatibility with existing shell execution logic

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,7 +54,7 @@
                 "@types/js-yaml": "^4.0.9",
                 "@types/libsodium-wrappers": "^0.8.2",
                 "@types/mocha": "^10.0.10",
-                "@types/node": "^25.7.0",
+                "@types/node": "^26.0.0",
                 "@types/semver": "^7.7.1",
                 "@types/sinon": "^21.0.1",
                 "@types/tmp": "^0.2.6",
@@ -1102,7 +1102,7 @@
             "version": "4.9.1",
             "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.9.1.tgz",
             "integrity": "sha512-phrYmNiYppR7znFEdqgfWHXR6NCkZEK7hwWDHZUjit/2/U0r6XvkDl0SYnoM51Hq7FhCGdLDT6zxCCOY1hexsQ==",
-            "devOptional": true,
+            "dev": true,
             "license": "MIT",
             "dependencies": {
                 "eslint-visitor-keys": "^3.4.3"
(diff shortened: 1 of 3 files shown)


Tatsinnit:
## ✅ Fixed Exit Code 0 Handling

Great catch @copilot! This was a **legitimate bug** - the `|| 1` fallback would incorrectly convert exit code 0 (success) to 1.

### 🐛 The Problem
```typescript
// ❌ WRONG: coerces 0 to 1 because 0 is falsy
const code = parseInt(String(err.code), 10) || 1;
```

If `err.code` was the string `"0"`, this would:
1. Parse to numeric `0`
2. Evaluate as falsy
3. Incorrectly return `1` instead of `0`

### ✅ The Fix
```typescript
// ✅ CORRECT: explicitly checks for NaN
const parsed = Number.parseInt(String(err.code), 10);
const code = typeof err.code === "number" ? err.code : Number.isNaN(parsed) ? 1 : parsed;
```

Now properly handles all valid exit codes including 0.

### 🧪 Verification
- ✅ Webpack compiles successfully
- ✅ Exit code 0 preserved correctly
- ✅ Invalid codes still default to 1

Fixed in commit [11dccaf5](https://github.com/Azure/vscode-aks-tools/pull/2259/commits/11dccaf5). Thank you for the review! 🙏

bosesuneha (approved):
lgtm

PR #2277, opened 2026-06-30
fix(container-assist): stop managed namespace update command from overwriting existing annotations/labels

bosesuneha:
## Problem

The `Annotate namespace` step in the CA-generated managed-namespace deploy workflows ran `az aks namespace update --annotations …` on every push. That command is an ARM PUT (full replacement), so it silently wiped every existing annotation **and** every label on the managed namespace, breaking AKS desktop project tracking and any other tooling that relied on namespace metadata.

The two annotations being stamped are static:

- `aks-project/workload-identity-id`
- `aks-project/workload-identity-tenant`

They're determined when the federated identity is created and never change between deploys — so there is no reason to re-stamp them on every push.

## Fix

- **Delete the destructive `Annotate namespace` step** from both managed-ns workflow templates (single-container `aks-deploy-managed-ns.template.yaml` and multi-container `workflow-multi-deploy-job-managed-ns.template.yaml`).
- **Move the one-time write into `oidcSetup.ts`** as a new exported function `annotateManagedNamespaceWithIdentity`, using the `@azure/arm-containerservice` SDK for a typed read-modify-write that preserves all existing annotations and labels via object spread.
- **Fail closed:** if the SDK read fails, log a warning and skip rather than risk a destructive PUT.
- The per-run `Annotate deployment` step is unchanged — `kubectl annotate --overwrite` is per-key and not destructive.

## Why move it out of CI

| Concern | Before (per-deploy CI) | After (one-time OIDC setup) |
|---|---|---|
| How often it runs | Every push | Once, at setup |
| CI principal RBAC needed | `managedNamespaces/write` | None |
| Data-loss risk | Wipes existing annotations + labels | Typed RMW preserves them |
| Per-deploy latency | +1 ARM read + 1 ARM write | 0 |

## Tests

- **5 new unit tests** in `oidcSetup.test.ts` covering:
  - Merge: existing annotations + labels + non-annotation properties + outer ARM envelope all survive the PUT
  - Empty-namespace edge case
  - Idempotency: re-running with new identity values overwrites only the workload-identity keys
  - Fail-closed: a failed `get()` does not result in a PUT
  - Best-effort: a failed PUT logs a warning and does not throw
- **2 inverted assertions** in `workflowTemplate.test.ts` confirm the destructive step is gone from both managed-ns templates and that the per-run deployment annotation step still renders.

`npm test` — **210 passing**.

## Manual verification recommended before merge

On a real cluster with a managed namespace that has pre-existing user annotations and labels:

```bash
# Pre: seed user metadata
az aks namespace update -g $RG --cluster-name $C -n $NS \
  --annotations user/keep-annotation=A --labels user/keep-label=L

# Run "AKS: Setup OIDC for GitHub" from the extension

# Post: both your keys AND the new workload-identity keys must be present
az aks namespace show -g $RG --cluster-name $C -n $NS \
  --query 'properties.{annotations:annotations, labels:labels}'
```

### DIFF
--- a/src/commands/aksContainerAssist/oidcSetup.ts
+++ b/src/commands/aksContainerAssist/oidcSetup.ts
@@ -38,6 +38,11 @@ const AKS_NAMESPACE_CONTRIBUTOR_ROLE_ID = "289d8817-ee69-43f1-a0af-43a45505b488"
 const ACR_PUSH_ROLE_ID = "8311e382-0749-4cb8-b61a-304f252e45ec";
 const ACR_TASKS_CONTRIBUTOR_ROLE_ID = "fb382eab-e894-4461-af04-94435c366c3f";
 
+// Managed-namespace annotation keys for AKS desktop / project tooling to
+// discover which federated workload identity is wired up for the namespace.
+const WORKLOAD_IDENTITY_CLIENT_ID_ANNOTATION = "aks-project/workload-identity-id";
+const WORKLOAD_IDENTITY_TENANT_ID_ANNOTATION = "aks-project/workload-identity-tenant";
+
 interface OIDCSetupResult {
     clientId: string;
     tenantId: string;
@@ -152,6 +157,23 @@ export async function setupOIDCForGitHub(
                     azureConfig.identityName,
                     repoInfo,
                 );
+
+                if (
+                    azureContext?.isManagedNamespace &&
+                    azureContext.namespace &&
+                    azureContext.clusterName &&
+                    azureContext.clusterResourceGroup
+                ) {
+                    progress.report({ message: l10n.t("Annotating managed namespace...") });
+                    await annotateManagedNamespaceWithIdentity(
+                        new ContainerServiceClient(credential, azureConfig.subscriptionId),
+                        azureContext.clusterResourceGroup,
+                        azureContext.clusterName,
+                        azureContext.namespace,
+                        identityResult.clientId,
+                        identityResult.tenantId,
+                    );
+                }
 
                 return {
                     clientId: identityResult.clientId,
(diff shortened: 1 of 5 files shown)


src/commands/aksContainerAssist/oidcSetup.ts
```diff
@@ -38,6 +38,11 @@ const AKS_NAMESPACE_CONTRIBUTOR_ROLE_ID = "289d8817-ee69-43f1-a0af-43a45505b488"
 const ACR_PUSH_ROLE_ID = "8311e382-0749-4cb8-b61a-304f252e45ec";
 const ACR_TASKS_CONTRIBUTOR_ROLE_ID = "fb382eab-e894-4461-af04-94435c366c3f";
 
+// Managed-namespace annotation keys for AKS desktop / project tooling to
+// discover which federated workload identity is wired up for the namespace.
+const WORKLOAD_IDENTITY_CLIENT_ID_ANNOTATION = "aks-project/workload-identity-id";
+const WORKLOAD_IDENTITY_TENANT_ID_ANNOTATION = "aks-project/workload-identity-tenant";
```
Tatsinnit, line 45: 💡Important suggestion:  Do we have official `msdn` or `aks-desktop` doc which we can add it in the comments please? This will be very beneficial for long term maintenance when this changes and folks will have official doc to refer to.

src/commands/aksContainerAssist/oidcSetup.ts
```diff
@@ -781,6 +803,50 @@ async function createFederatedCredential(
...
+    };
+
+    try {
+        const poller = client.managedNamespaces.createOrUpdate(clusterResourceGroup, clusterName, namespace, {
+            ...existing,
+            properties: mergedProperties,
+        });
+        await poller.pollUntilDone();
+    } catch (error) {
+        logger.warn(
```
Tatsinnit: 💡 Regarding `error-handling` idea share - this is good to log the warning but I feel early to mid stage devops will hardly look at this log, shall we add the `vscode.warningMessage` api to bubble anything valuable, so that user can see it as notification and do something about it.

Tatsinnit (dismissed):

src/commands/aksContainerAssist/oidcSetup.ts
bosesuneha: There's no public MSDN doc for these keys yet since it's an internal contract between the two repos, so I've documented the aks-desktop file path where the same constants are used for future reference.

src/commands/aksContainerAssist/oidcSetup.ts
bosesuneha: Agreed, added notification message as well

Tatsinnit (approved):

PR #2280, opened 2026-07-02
Update VSIX publishing to use service connection

bosesuneha:
Refactor the VSIX publishing process to utilize a service connection instead of a token parameter, enhancing security and simplifying the pipeline configuration.


Tatsinnit:
💡 pretty cool, I wonder if we test this post kickstart review!! Let’s discuss on sidelines. Thanks

Tatsinnit (approved):

PR #2314, opened 2026-07-16
feat(skills): contribute CA containerization skills

benjaminbob21:
Bundle the five Containerization Assist skills (analyze-repo, generate-dockerfile, fix-dockerfile, generate-k8s-manifests, deploy-to-aks) from the containerization-assist-mcp package into dist/skills at build time and register them via contributes.chatSkills, each gated by a new aks.skillsEnabledPreview preview flag (default off). Coexists with the in-repo kickstart-* skills; the containerization-assist-mcp dependency is already present on this branch.


package.json
```diff
@@ -158,6 +158,11 @@
                     "default": true,
                     "description": "Set to true to enable deployment related file generation capability using AI tool. (Preview feature)"
                 },
+                "aks.skillsEnabledPreview": {
```
Tatsinnit: :bulb: we already previous had a CA level flag for featureI wonder why do we need this or how are we planning larger design?

package.json
```diff
@@ -1095,7 +1100,12 @@
             { "path": "./skills/kickstart-phase-acceleration/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" },
             { "path": "./skills/kickstart-collaborator-voice/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" },
             { "path": "./skills/kickstart-file-generation/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" },
-            { "path": "./skills/kickstart-safeguard-checklist/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" }
+            { "path": "./skills/kickstart-safeguard-checklist/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" },
+            { "path": "./dist/skills/analyze-repo/SKILL.md", "when": "config.aks.skillsEnabledPreview" },
+            { "path": "./dist/skills/generate-dockerfile/SKILL.md", "when": "config.aks.skillsEnabledPreview" },
```
Tatsinnit: :bulb:  can you please share the reference of the `./dist` dir structure usage?

package.json
```diff
@@ -1095,7 +1100,12 @@
             { "path": "./skills/kickstart-phase-acceleration/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" },
             { "path": "./skills/kickstart-collaborator-voice/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" },
             { "path": "./skills/kickstart-file-generation/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" },
-            { "path": "./skills/kickstart-safeguard-checklist/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" }
+            { "path": "./skills/kickstart-safeguard-checklist/SKILL.md", "when": "config.aks.kickstartEnabledPreview == true" },
```
Tatsinnit, line 1098: 💡 Why is the kickstart changes for CA specific PR please =) ?

package.json
benjaminbob21: The other flag is for kickstart, decided to separate them cos for now kickstart doesnt use these skills that are shipped from CA. But moving forward these CA skills should replace some of the kickstart skills and they would all be under the same flag. I'm fine to consolidate now under a single flag if suggested though, thanks

package.json
benjaminbob21: The change is just a comma for this line

package.json
benjaminbob21: This is the reference for contribute.chatSkills https://code.visualstudio.com/api/references/contribution-points#contributes.chatSkills. It requires path to be relative to the extension root and to resolve to a file that exists inside the package extension. The Kickstart skills live in the repo under skills, so they point there directly. The CA skills don't, they ship inside the containerization-assist-mcp npm package, so at build time we copy them into ./dist/skills via CopyPlugin and that's why we point to ./dist. The ./dist is what gets packaged into the VSIX so I tried to find a way to get it in there.

package.json
```diff
@@ -1109,7 +1119,7 @@
         "package": "vsce package",
         "webpack": "npm run build:webview && webpack --mode production",
         "webpack-dev": "npm run build:webview && webpack --mode development --watch",
-        "test-compile": "npm run build:webview && tsc -p ./ && cp resources/yaml/aks-deploy.template.yaml out/src/commands/aksContainerAssist/",
+        "test-compile": "npm run build:webview && tsc -p ./ && node -e \"const fs=require('fs');const path=require('path');const templateSrc=path.join('resources','yaml','aks-deploy.template.yaml');const templateDstDir=path.join('out','src','commands','aksContainerAssist');fs.mkdirSync(templateDstDir,{recursive:true});fs.copyFileSync(templateSrc,path.join(templateDstDir,path.basename(templateSrc)));fs.rmSync(path.join('dist','skills'),{recursive:true,force:true});fs.mkdirSync('dist',{recursive:true});fs.cpSync(path.join('node_modules','containerization-assist-mcp','skills'),path.join('dist','skills'),{recursive:true});\"",
```
bosesuneha: can you move this to a separate script so that it's readable?

benjaminbob21: Alright I will do that, thank you

Tatsinnit:
Hi @benjaminbob21, great work on the build setup and script extraction!

I'd like to suggest consolidating `aks.kickstartEnabledPreview` and `aks.skillsEnabledPreview` into a single flag in this PR, rather than waiting for a future refactor. Here's the reasoning:

Why now is better than later:

- Both flags control access to the same conceptual feature: AKS agent/skill capabilities
- You're already planning to consolidate them when CA skills replace kickstart skills
- Consolidating now avoids a multi-phase refactor + deprecation cycle later
- Users see one clear setting for "enable AKS preview skills"
- Implementation: Keep `aks.skillsEnabledPreview` (or rename to `aks.agentSkillsEnabledPreview` for clarity), and update all kickstart skill when clauses to use it. ~15 lines of change in this PR.

Benefits:

- No future flag rename/migration needed
- Clearer semantics in settings UI
- Single source of truth for preview features
- Reduced maintenance burden when skills eventually merge

Would you be open to this in this PR? Happy to discuss if you see constraints I'm missing!

Also, let's also document if the previous flag is not documented properly. Thank you!

bosesuneha:
@benjaminbob21  I feel we can have the skills available by default without having them gated behind a flag. As @Tatsinnit pointed out it might get confusing to maintain so many similar flags. For better clarity, could you please add the motivation behind adding the skills in the extension?

benjaminbob21:
> Hi @benjaminbob21, great work on the build setup and script extraction!
> 
> I'd like to suggest consolidating `aks.kickstartEnabledPreview` and `aks.skillsEnabledPreview` into a single flag in this PR, rather than waiting for a future refactor. Here's the reasoning:
> 
> Why now is better than later:
> 
> * Both flags control access to the same conceptual feature: AKS agent/skill capabilities
> * You're already planning to consolidate them when CA skills replace kickstart skills
> * Consolidating now avoids a multi-phase refactor + deprecation cycle later
> * Users see one clear setting for "enable AKS preview skills"
> * Implementation: Keep `aks.skillsEnabledPreview` (or rename to `aks.agentSkillsEnabledPreview` for clarity), and update all kickstart skill when clauses to use it. ~15 lines of change in this PR.
> 
> Benefits:
> 
> * No future flag rename/migration needed
> * Clearer semantics in settings UI
> * Single source of truth for preview features
> * Reduced maintenance burden when skills eventually merge
> 
> Would you be open to this in this PR? Happy to discuss if you see constraints I'm missing!
> 
> Also, let's also document if the previous flag is not documented properly. Thank you!

Thanks Tats! Sounds great to me. Happy to make the change. Just some quick things to highlight, the current kickstart flag not only enables the skills but also the kickstart agent itself so might not totally be accurate to have it named `aks.skillsEnabledPreview` or anything with a skills naming. I expect that the kickstart agent shouldn't be enabled without the kickstart skills so they can't be under separate flags. I'm leaning towards Suneha's idea to just have the skills available by default without having them gated behind a flag. 
Another option will be to just have the 5 skills gated behind the `kickstartEnabledPreview` flag or maybe some name more nuanced than that to accomodate all. Glad to have some more of your thoughts on this via teams chat for better conversations and get @bosesuneha opinions on this too.

bosesuneha:
@benjaminbob21 can you please rebase this branch onto the latest upstream main?

benjaminbob21:
> @benjaminbob21 can you please rebase this branch onto the latest upstream main?

Done, although it says some workflows are awaiting approval for now

bosesuneha (approved):
lgtm

PR #2329, opened 2026-07-21
fix(container-assist): skip build-output dirs when generating manifests

### LINKED ISSUE #2319: Kubernetes deployment manifest generation error
When generating manifests for the github.com/Azure-Samples/contoso-air demo app, it creates 7 manifests and creates files in the k8s directory as well as the .next directory (the sample app is a Next.js app).

It is not clear why the process creates manifests across the two directories. Some of the manifests seem to overlap with ingress manifests only being created in the `.next` directory (see image below of all the files it created) 

Image

> Note the file paths of all the files

Image

bosesuneha:
Fixes #2319.

Manifest generation iterates over modules returned by the SDK's `analyzeRepo`, whose walker doesn't skip build-output dirs. For Next.js apps the `.next/` standalone output ships its own `package.json`, so it's reported as an extra module and a duplicate manifest set is written to `<root>/.next/k8s/` alongside `<root>/k8s/`.

Filter out modules whose path resolves inside an excluded directory (`.next/`, `dist/`, `node_modules/`, etc.) in `analyzeRepository`, so all downstream steps (Dockerfile, manifests, workflow) ignore them.

### DIFF
--- a/src/commands/aksContainerAssist/fileOperations.ts
+++ b/src/commands/aksContainerAssist/fileOperations.ts
@@ -84,6 +84,16 @@ const EXCLUDED_DIRS = new Set([
 /** Maximum directory depth for Dockerfile and manifest scanning. */
 const SCAN_MAX_DEPTH = 3;
 
+export function isExcludedModulePath(modulePath: string, rootPath: string): boolean {
+    const relative = path.relative(rootPath, modulePath);
+
+    if (relative === "" || relative.startsWith("..") || path.isAbsolute(relative)) {
+        return false;
+    }
+
+    return relative.split(path.sep).some((segment) => EXCLUDED_DIRS.has(segment));
+}
+
 /**
  * Searches recursively (up to 3 levels deep) for Dockerfiles under rootPath.
  * Returns absolute paths sorted shallowest-first; excluded dirs are skipped.
--- a/src/commands/aksContainerAssist/fileOperations.ts
+++ b/src/commands/aksContainerAssist/fileOperations.ts
@@ -57,7 +57,7 @@ export async function checkExistingFiles(folderPath: string): Promise<ExistingFi
 }
 
 /** Directories to skip during recursive scans (build artifacts, dependencies, etc.) */
-const EXCLUDED_DIRS = new Set([
+export const EXCLUDED_DIRS = new Set([
     "node_modules",
     ".git",
     "dist",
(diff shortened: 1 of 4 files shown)


src/commands/aksContainerAssist/fileOperations.ts
```diff
@@ -84,6 +84,16 @@ const EXCLUDED_DIRS = new Set([
...
+export function isExcludedModulePath(modulePath: string, rootPath: string): boolean {
+    const relative = path.relative(rootPath, modulePath);
+
+    if (relative === "" || relative.startsWith("..") || path.isAbsolute(relative)) {
+        return false;
+    }
+
+    return relative.split(path.sep).some((segment) => EXCLUDED_DIRS.has(segment));
+}
+
```
pauldotyu: Would it make sense (or is it even possible) to find `.gitignore` and exclude directories within it?

Tatsinnit: 💡 Leaving few points and thought here. Using `.gitignore` wouldn't work well here because:

 1. **Project-configuration-specific** — `.gitignore` varies per repository and developer workflow. It's meant for version control intent, not tool behavior.
 2. **Tool predictability** — The containerization tool needs consistent, predictable behavior across all projects. If a team removes a directory from `.gitignore` for local reasons (e.g., tracking a build artifact for CI), the tool would suddenly include it, breaking user expectations without a code change.
 3. **Purpose mismatch** — `.gitignore` answers "what shouldn't be in git?" (version control question), not "what should the build tool skip?" (build/analysis question). These aren't always the same.

The use of `EXCLUDED_DIRS` approach is the right call because it covers **universal build artifacts** that should never be analyzed for containerization, regardless of a project's git configuration:

 - Language/framework standard outputs (dist, `target`, `build`, `out`)
 - Dependency/virtual environments (node_modules, `.venv`, `venv`)
 - Standard metadata/config directories (.git, .github, .vscode)

This keeps the tool reliable and predictable. If there are additional common patterns we discover should be excluded, we can extend the list, but sourcing from .gitignore would create instability.

Tatsinnit (dismissed):
💡 Suggestion please: Can we please generate test `vsix` and get it tested please as well, just inc are we missed a behaviour and re-assurance please. thank you so much!!

src/commands/aksContainerAssist/fileOperations.ts
bosesuneha: Thanks @Tatsinnit, good points. My thinking is to layer .gitignore on top of the static list, not replace it. This list controls which dirs we skip when analyzing the repo and where we don't generate manifest files.
The static list stays the source of truth and is always applied, so node_modules, .next, dist, etc. are always excluded. We only add .gitignore dirs on top of that to catch project specific directories we don't need to scan. It also trims what we send to the model during analysis, so we save some context tokens.
Let me know what you think.

bosesuneha:
Here's the vsix for testing purposes: 
[vscode-aks-tools-pr2329.vsix.zip](https://github.com/user-attachments/files/30472704/vscode-aks-tools-pr2329.vsix.zip)

Need to rename (not unzip) the file to .vsix extension by removing .zip extension before testing.

src/commands/aksContainerAssist/fileOperations.ts
pauldotyu: All sounds good to me @bosesuneha and @Tatsinnit! Saving tokens and giving devs some level of control (in this case, via .gitignore) is a good thing. Let's also make sure this logic is documented somewhere in the docs 😊

Tatsinnit (approved):

PR #2330, opened 2026-07-21
Container Assist: optional ACR and per-artifact selection

### LINKED ISSUE #2318: Make ACR dependency optional when generating Kubernetes deployment manifests
When using extension for manifest generation only, it asks for an attached ACR. In some cases, a user may already have a container image elsewhere (i.e., GitHub Container Registry, etc.) and should not be a requirement.

IMO, this should only be a requirement if the user has selected to generate manifests and a GitHub workflow file (which would create an image build pipeline).

As of today, users may want to deploy from an existing container image, but this isn’t easy to do without manual modifications to the Kubernetes manifests.

Image

bosesuneha:
Addresses: #2318

## What

Make ACR optional in the Container Assist flow and let users generate each artifact independently.

- **Optional ACR for manifests**: when generating manifests without a workflow, users can point at an existing image reference (GHCR, Docker Hub, etc.) instead of an ACR. ACR remains required for workflow generation (it builds and pushes the image).
- **Per-artifact selection**: the action picker is now a 3-item multi-select — Dockerfile, Kubernetes manifests, and GitHub workflow — each selectable independently.

## Key changes

- **`types.ts`**: replaced the single `GenerateDeployment` action with discrete `GenerateDockerfile` and `GenerateManifests` actions; added a `DeploymentArtifacts` selection shape.
- **`azureSelections.ts`**: `AzureContext.acrName`/`acrResourceGroup` are now optional and mutually exclusive with a new `customImage` field; added an image-source picker (ACR vs. existing image reference).
- **`containerAssistService.ts`**: `generateDeploymentFiles` now takes a per-module `imageRepositoryFor(moduleName)` resolver (replacing the single `acrLoginServer`) and an `artifacts` flag to generate the Dockerfile and manifests independently.
- **`aksContainerAssist.ts`**: reworked orchestration for independent selection; renamed the local `has*` flags to `want*` (they now describe desired outputs, not just selected actions); removed the old `processContainerAssistAction` switch dispatcher in favor of direct calls.
- **`workflowGenerator.ts`**: require ACR context when building the workflow configuration.
- **`command.test.ts`**: updated QuickPick assertions for the 3-item picker.


src/commands/aksContainerAssist/aksContainerAssist.ts
```diff
@@ -18,11 +18,17 @@ import { showWizardExitConfirmation } from "./wizardUtils";
 export { showWizardExitConfirmation } from "./wizardUtils";
 
 export async function containerizeApp(context: IActionContext, target: unknown): Promise<void> {
```
Tatsinnit, line 20: 💡 Caution and please check, both the function are public scoped and the return is now changed so jsut to note to keep a wider check if another place is not effected by this, might as well add a quick note or comment for this function and its operation for maintainability.

src/commands/aksContainerAssist/aksContainerAssist.ts
```diff
@@ -242,21 +248,24 @@ async function findProjectRoot(startPath: string, workspaceRoot: string): Promis
 async function showContainerAssistQuickPick(
     defaultActions: ContainerAssistAction[] = [],
 ): Promise<ContainerAssistAction[] | undefined> {
-    const deploymentPicked = defaultActions.includes(ContainerAssistAction.GenerateDeployment);
```
Tatsinnit, line 245: 💭 Please note: I see you have removed these t2 variables and used the direct assignment I have 3-4 key thoughts which this change will effect: (majority of these changes int his file happened with this inline choice)

- Reduced maintainability (biggest concern): Previously the variable were used in other code down and only one place was assigned but now if there is change in need all the above will be then changed 1 by 1.
- Harder to read
- Risk of inconsistencies
- More difficult debugging

General rule of thumb I think is for the case to keep a variable if it:

- is used 2 or more times
- has a meaningful name that improves readability
- avoids repeating a long expression
- caches an expensive computation
- represents an important business concept

For, Inline it if it:

- is used once
- is trivial
- doesn't improve readability by having a separate name

Old cod was below: 

```*.ts
    const deploymentPicked = defaultActions.includes(ContainerAssistAction.GenerateDeployment);
    const workflowPicked = defaultActions.includes(ContainerAssistAction.GenerateWorkflow);
```

src/commands/aksContainerAssist/aksContainerAssist.ts
```diff
@@ -290,11 +305,13 @@ async function executeContainerAssistActions(
         return;
     }
 
-    const hasDeployment = selectedActions.includes(ContainerAssistAction.GenerateDeployment);
-    const hasWorkflow = selectedActions.includes(ContainerAssistAction.GenerateWorkflow);
-    const hasBothActions = hasDeployment && hasWorkflow;
+    const wantDockerfile = selectedActions.includes(ContainerAssistAction.GenerateDockerfile);
```
Tatsinnit, line 312: 💡 I am not quite sure if this is agent code or known changes, from `has` to `want` what is the thinking behind it please? and also the property used changed, might be please update the PR description for details of these changes.

src/commands/aksContainerAssist/aksContainerAssist.ts
```diff
@@ -307,114 +324,126 @@ async function executeContainerAssistActions(
...
+            workspaceFolder,
+            projectRoot,
+            azureContext,
+            hasBothActions: false,
         });
+        return;
+    }
 
-        if (!workflowResult?.workflowPath) {
+    // Deployment artifacts + workflow: the workflow depends on a Dockerfile and manifests.
```
Tatsinnit: 💡 Some parts of this major function seems like could be extracted into small single responsible private decision making functs?

src/commands/aksContainerAssist/aksContainerAssist.ts
```diff
@@ -307,114 +324,126 @@ async function executeContainerAssistActions(
...
     deploymentFiles?: string[];
     workflowPath?: string;
-    /** Absolute paths to any generated Kubernetes manifests (populated by GenerateDeployment). */
+    /** Absolute paths to any generated Kubernetes manifests. */
     manifestPaths?: string[];
     /** Primary module name from SDK analysis (e.g. package.json "name"). */
     primaryModuleName?: string;
 }
 
-async function processContainerAssistAction(
```
Tatsinnit, line 381: 💡 This function is renamed or removed? I think the description of this PR is not detailing any of these renaming decision or removing of function changes. Lets make sure we capture this better for archiving and maintainability please.

src/commands/aksContainerAssist/azureSelections.ts
```diff
@@ -602,18 +602,111 @@ export async function promptForWorkflowName(appName: string): Promise<string | u
...
 
+/** The resolved container image source: either an ACR or a user-supplied image reference. */
+interface ImageSourceSelection {
+    acrName?: string;
+    acrResourceGroup?: string;
+    customImage?: string;
+}
+
+/** Prompts for a full container image reference (registry/repository[:tag]) to use in the manifests. */
+async function promptForCustomImage(): Promise<string | undefined> {
```
Tatsinnit, line 640: How about something like this:


```*.ts
function validateImageReference(value: string): string | undefined {
    const image = value.trim();

    if (!image) {
        return l10n.t("Image reference is required");
    }

    if (/\s/.test(image)) {
        return l10n.t("Image reference cannot contain spaces");
    }

    return undefined;
}

async function promptForCustomImage(): Promise<string | undefined> {
    const image = await vscode.window.showInputBox({
        prompt: l10n.t(
            "Enter the full container image reference to use in the manifests (e.g. ghcr.io/org/app:latest).",
        ),
        placeHolder: "ghcr.io/org/app:latest",
        ignoreFocusOut: true,
        validateInput: validateImageReference,
    });

    if (image === undefined) {
        return showWizardExitConfirmation(promptForCustomImage);
    }

    return image.trim();
}
```

src/commands/aksContainerAssist/azureSelections.ts
```diff
@@ -602,18 +602,111 @@ export async function promptForWorkflowName(appName: string): Promise<string | u
...
+    const image = await vscode.window.showInputBox({
+        prompt: l10n.t(
+            "Enter the full container image reference to use in the manifests (e.g. ghcr.io/org/app:latest).",
+        ),
+        placeHolder: "ghcr.io/org/app:latest",
```
Tatsinnit, line 645: 💡 Why one `ghcr` sample or is it jsut a sample?

src/commands/aksContainerAssist/azureSelections.ts
```diff
@@ -602,18 +602,111 @@ export async function promptForWorkflowName(appName: string): Promise<string | u
...
+async function selectAcrImageSource(
+    sessionProvider: ReadyAzureSessionProvider,
+    subscriptionId: string,
+    cluster: Cluster,
+): Promise<ImageSourceSelection | undefined> {
+    const acr = await selectClusterAcr(sessionProvider, subscriptionId, cluster);
+    return acr ? { acrName: acr.name, acrResourceGroup: acr.resourceGroup } : undefined;
+}
+
 /**
```
Tatsinnit: 💡 for thoughts and suggestions:

```*.ts
/**
 * Resolves the container image source for the generated manifests.
 *
 * When `requireAcr` is true, an Azure Container Registry must be selected.
 * Otherwise, users may either select an attached ACR or provide an existing
 * container image reference (for example GHCR or Docker Hub).
 */
async function selectImageSource(
    subscriptionId: string,
    cluster: Cluster,
    requireAcr: boolean,
): Promise<ImageSourceSelection | undefined> {
    if (requireAcr) {
        return selectAcrImageSource(sessionProvider, subscriptionId, cluster);
    }

    const sourceType = await promptForImageSource(
        sessionProvider,
        subscriptionId,
        cluster,
        requireAcr,
    );

    if (!sourceType) {
        return undefined;
    }

    switch (sourceType) {
        case "acr":
            return selectAcrImageSource(sessionProvider, subscriptionId, cluster);

        case "custom": {
            const customImage = await promptForCustomImage();
            return customImage ? { customImage } : undefined;
        }
    }
}

async function promptForImageSource(
    subscriptionId: string,
    cluster: Cluster,
    requireAcr: boolean,
): Promise<"acr" | "custom" | undefined> {
    const picked = await vscode.window.showQuickPick(
        [
            {
                label: l10n.t("$(cloud) Use an Azure Container Registry"),
                description: l10n.t("Select an ACR attached to the cluster"),
                sourceType: "acr" as const,
            },
            {
                label: l10n.t("$(link) Use an existing image reference"),
                description: l10n.t(
                    "Point manifests at an image you already have (e.g. GHCR, Docker Hub)",
                ),
                sourceType: "custom" as const,
            },
        ],
        {
            title: l10n.t("Container Image Source"),
            placeHolder: l10n.t("Select the container image source for the manifests"),
        },
    );

    if (!picked) {
        return showWizardExitConfirmation(() =>
            promptForImageSource(
                sessionProvider,
                subscriptionId,
                cluster,
                requireAcr,
            ),
        );
    }

    return picked.sourceType;
}

```

src/commands/aksContainerAssist/containerAssistService.ts
```diff
@@ -278,37 +284,41 @@ export class ContainerAssistService {
 
     async generateDeploymentFiles(
         folderPath: string,
-        acrLoginServer?: string,
+        imageRepositoryFor?: (moduleName: string) => string | undefined,
```
Tatsinnit, line 287: 💡 Important to note, this change will then make a interesting assumption that `imageRepositoryFor` itself is optional.

If present, it's a function that:
- takes a `moduleName: string`
- returns either a `string or undefined`.

If every implementation can provide the function (even if it sometimes returns no result), I'd write:

```*.ts
imageRepositoryFor: (moduleName: string) => string | undefined;
```

If some implementations genuinely don't support image repository resolution at all, then your original declaration is a good fit:

```*.ts
imageRepositoryFor?: (moduleName: string) => string | undefined;
```

That distinction *"feature unavailable"* versus *"feature available but no result for this module"* is usually what determines whether the function itself should be optional.

src/commands/aksContainerAssist/aksContainerAssist.ts
bosesuneha: This is just a better rename from `has` to `want`. Earlier we used (hasDeployment/hasWorkflow) to track which actions were selected. Now (wantDockerfile/wantManifests/wantWorkflow/wantDeploymentArtifacts). tracks individually the three options for Dockerfile, Manifest and workflow for based on which outputs the user wants in the quick pick. I've added this rationale to the PR description.

src/commands/aksContainerAssist/aksContainerAssist.ts
bosesuneha: This is removed, not renamed. It was a switch dispatcher over the old two-action enum. With the new separate dockerfile and manifest branching its no longer needed, so orchestration now calls generateDeploymentFiles / generateWorkflowFile directly. Added in the PR description.

src/commands/aksContainerAssist/azureSelections.ts
bosesuneha: This is just an example placeholder to signal the expected registry/repository[:tag] format. I'm open to a more neutral sample if you prefer

src/commands/aksContainerAssist/aksContainerAssist.ts
bosesuneha: Checked, these are command entry points registered only in extension.ts, and their signature is unchanged. Only the default actions they pass differ. I've noted in the PR description.

src/commands/aksContainerAssist/aksContainerAssist.ts
bosesuneha: Done, pulled the defaultActions.includes(...) checks out into named consts (dockerfilePicked, manifestsPicked, workflowPicked), matching the earlier style.

src/commands/aksContainerAssist/aksContainerAssist.ts
bosesuneha: Yup I've made the model-selection and workflow-generation logic into two separate helpers

src/commands/aksContainerAssist/azureSelections.ts
bosesuneha: Thanks for the suggestion, made validateImageReference a separate function

src/commands/aksContainerAssist/azureSelections.ts
bosesuneha: I've now split out promptForImageSource, so selectImageSource just decides ACR vs. existing image and delegates

src/commands/aksContainerAssist/containerAssistService.ts
bosesuneha: I'd kept it optional on purpose. Some flows like dockerfile-only, and the tests pass no resolver at all,so the `string | undefined` return covers the no repo case as well.

Tatsinnit (approved):

Tatsinnit (approved):

PR #2337, opened 2026-07-29
fix(ci): fix Windows smoke-test renderer OOM

davidgamero:
## Summary

Supersedes #2336, which raised the extension-host V8 heap to fix the `build (windows-latest)` smoke-test OOM. That bump targets the wrong process; this PR reverts it and fixes the real cause.

## Root cause

The OOM is in the Electron **renderer**, not the extension host (which exits `code: 0`), so `NODE_OPTIONS` could never reach it. Memory isn't the constraint either — 16 GB runner, ~1.4 GB peak. Two bugs drive the crash:

1. **Leaked webview.** `webview.test.ts` never disposes its `StyleTestPanel`. The orphaned frame starves the extension-host event loop on the 4-core runner until the renderer OOMs.
2. **Masked assertion.** Assertions run in floating `import("chai").then(...)` callbacks mocha never awaits, hiding a real failure: `rules` is empty because VS Code 1.130 wraps default styles in a CSS `@layer`, so the flat `filter(isStyleRule)` over `_defaultStyles.cssRules` returns `[]`.

## Fix

- `webview.test.ts`: dispose the panel in `finally`; await assertions.
- `BasePanel.ts`, `InspektorGadgetPanel.ts`: return the `WebviewPanel` from `show()`.
- `TestStyleViewer.tsx`: recurse into grouping rules in `getCssRules`.
- `runTests.ts`: revert the `--max-old-space-size=8192` bump.
- `dependabot.yml`: add `cooldown: default-days: 7`.

## Validation

`lint:all` and `tsc` clean. Windows smoke step drops 680–718s → **58s**, 225 passing, ext host exit 0, no OOM/unresponsive markers — event-loop starvation resolved.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,11 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
+    # Bake newly released versions for 1 week before proposing them. This gives an
+    # internal evaluation window and avoids proposing versions that are not yet
+    # mirrored in the internal package feed proxy.
+    cooldown:
+      default-days: 7
     groups:
       typescript-eslint:
         patterns:
@@ -12,6 +17,8 @@ updates:
     directory: "/webview-ui"
     schedule:
       interval: "monthly"
+    cooldown:
+      default-days: 7
     groups:
       typescript-eslint:
         patterns:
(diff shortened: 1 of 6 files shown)


davidgamero:
windows CI passed on this PR in <60 sec, whereas it was hanging for >10min on previous PRs with the starved eventloop from the webview leak

Tatsinnit (dismissed):

Tatsinnit (approved):

PR #2353, opened 2026-07-30
feat(argocd): rename to "Create Argo CD Application" and simplify flow

### LINKED ISSUE #2322: Error when launching Argo CD dashboard UI
After a user applies an Argo CD application manifest, the extension asks if they would like to launch the dashboard to view the status of the deployment. The extension assumes that the user installed Argo CD using the default URL of localhost:8080 but if folks installed using something else like localhost:9000, the Entra SSO breaks as the AKS extension’s integration opens the web UI on port 8080 which causes a mismatch on how SSO was configured for the global domain. If the argocd-server service is set to ClusterIP and global.domain set to local host in the argocd-cm configmap, the extension should port-forward based on the on configured port 

Image
Port-forward after applying application manifest

Image
Redirect URI mismatch between what is configured in the app registration and port that Argo CD is being served on

### LINKED ISSUE #2323: Argo CD application manifest guidance
The Argo CD extension leans heavily into the separation of code and K8s manifest and expects you to have manifests in a separate repository. The messaging pops up in the README.md that gets generated when the ArgoCD application manifest is generated. 

This is a best practice; however, this conflicts with the extensions existing practice of placing generated Kubernetes deployment manifests in the application source repository.

### LINKED ISSUE #2324: Argo CD menu terminology
While navigating the UI, I noticed a menu item labeled `Create Argo CD GitOps pipeline`

I'd like to raise this as a terminology/cleanup issue because it conflates two distinct DevOps concepts: CI/CD Pipelines and GitOps Controllers.

**Proposed changes:**

Can the menu item/label be updated to something that aligns with Argo CD's actual domain model?

Some options I had in mind are:
- `Create Argo CD Application` (Argo CD natively calls these "Applications")
- `Configure GitOps Sync`
- `Set up GitOps Repository`

**Rationale:**
- Argo CD is a GitOps controller (an application lifecycle manager). It operates via a *reconciliation loop*: watching a Git repository for manifest changes and continuously syncing the cluster to match that desired state.
- To me, a pipeline refers to a linear, stepwise process (e.g., `checkout → build → test → lint → package`). This belongs in CI tools like GitHub Actions, Jenkins, or Tekton.
- Describing Argo CD as a "pipeline" reinforces the misconception that it handles the build/test phases of software delivery. It does not. It only handles the deployment/sync phase via GitOps principles.

### LINKED ISSUE #2325: UX feedback on ArgoCD Application Manifest Generation
I spent some time testing the new ArgoCD application manifest generation feature in the VS Code extension. Here is my feedback on how to improve the flow

**The Wizard / Pop-ups**
The initial pop-up is really dense. It needs to be much simpler to understand what’s going to happen and what I need to do next.

Image

_Dialog that pops up after clicking to create Argo CD GitOps pipeline_

Also, having 4 buttons is confusing. I noticed that "Open Config Repository..." and "Continue to Create ArgoCD Pipeline" seem to do the same thing. Just make one primary action clear instead of cluttering it.

Let's drop the "Hollywood Principle" messaging entirely. Users coming to this tool likely already know what GitOps is; we don't need to teach them principles here, just help them get their manifests generated.

**Repo Setup**

The options for setting up the Git repo are unclear. Most people are going to want a remote URL anyway; just ask for that and skip the extra steps.

Image

_Git repo location_

"Browse GitHub repos...": The list seems sporadic. It’s hard to tell what criteria it's using to show repos. I had to try a few times just to find my own contoso-air repo. That needs to be more reliable or clearly sorted.

After picking the repo, stop making assumptions about where the manifest goes. Ask me for the path explicitly so I don't have to manually modify YAML later.

The "Application source repository annotation" field can be confusing to some; especially when putting the ArgoCD manifest in the same repo as the app code. In that case, this step isn't necessary.

**Output**
If we are generating a README.md and application.yaml, should forego the creation of the README. The comments in the YAML already explain what's happening. Most devs are just going to delete it immediately anyway.

**Private Repos / Auth**
Finally, what’s the flow for private repos? Does it prompt for a PAT and handle the repo credentials, or do I have to manually set those up in ArgoCD first? Ideally, this extension should help create the GitHub OAuth App or handle the PAT logic so we don't have to leave VS Code.

**Missing features**
- The ability to setup Application manifest that point to Helm charts or Kustomization directories with the ability to set parameter overrides for Helm charts and Kustomization patches

Tatsinnit:
## Summary

Addresses UX, terminology, and a dashboard-launch bug on the Argo CD Application feature, resolving four issues: #2322, #2323, #2324, #2325.

The command is now simpler and less opinionated: no wall-of-text modal, no forced "separate repo" guidance, explicit paths instead of hidden assumptions, an optional lightweight README, and correct Argo CD terminology. It also no longer breaks Entra ID SSO when the Argo CD UI is served on a non-default port.

Fixes #2322
Fixes #2323
Fixes #2324
Fixes #2325

**VSIX FIle:**

*  [vscode-aks-tools-2.3.0-argo-fixes.vsix.zip](https://github.com/user-attachments/files/30530473/vscode-aks-tools-2.3.0-argo-fixes.vsix.zip)

### Rename command (#2324)

- **"Create Argo CD GitOps Pipeline" → "Create Argo CD Application"** across the command title, localized strings, setting description, Copilot chat button, and docs.
- Rationale: Argo CD manages "Applications" (a GitOps controller), not CI "pipelines." The old name conflated two distinct DevOps concepts.

### Drop the "separate repo" prescription (#2323)
- Removed the "manifests MUST live in a separate GitOps config repo" / Hollywood Principle messaging from the generated YAML template and README.
- Manifests may now live in the **same repository as the app or a separate one** — the user chooses via the repo URL and path.
- Removed the redundant `aks-extension/source-repo` annotation.

### Simplify the UX (#2325)
- **Removed the dense blocking pre-generate modal.** The command goes straight to prompts; a non-modal "Learn More" toast appears only *after* the manifest is created, so it never gates the flow.
- **Explicit paths** — added prompts for the manifest path (default `k8s`) and the output path (default `argocd`) instead of assuming `apps/<name>/`, so users no longer have to correct the YAML afterward.
- **Single repo prompt** — collapsed the separate "config repo" and "source repo" prompts into one; removed the app-source/deployment-manifest detection logic.
- **Optional README** — now off by default via a QuickPick. When enabled, it's a trimmed, opinion-free `<app-name>-README.md` covering just install → apply → UI → status.
- Output is now `<output>/<app-name>.yaml` (+ optional `<app-name>-README.md`).

### Fix Argo CD dashboard launch on a non-default port (#2322)
- The post-apply "Open UI" flow previously hard-coded a `localhost:8080` port-forward. When Argo CD is served on a different port (`argocd-server` as ClusterIP with `argocd-cm` `url` / `global.domain` set to e.g. `localhost:9000`), opening 8080 broke Entra ID SSO because the app-registration redirect URI is tied to the configured port.
- The extension now reads the configured port from the `argocd-cm` ConfigMap (`url`, falling back to `global.domain`) and uses it for **both** the local side of the port-forward and the opened URL, defaulting to `8080` when none is set.

## Review fixes applied in this branch
- CodeQL "incomplete URL substring sanitization" in tests (word-boundary match).
- Output-path input-box cancel now aborts instead of silently writing to the workspace root.
- Hardened the user-supplied output path: normalises `/` and `\` separators and rejects absolute paths / `..` traversal that would escape the workspace.
- Updated UX copy, the Copilot plugin message, and docs to reference the actual generated filename (`<app-name>.yaml`).
- Fixed cross-platform test assertions (Windows drive-letter casing).

## Tests
- Extracted the write logic into an exported `writeArgoCDArtifacts()` and exported the manifest/README builders + `resolveOutputDirUri` + `parseArgoCDPort` for testability.
- Added unit tests covering placeholder substitution, opinion-free content, output-folder creation (nested/recursive, pre-existing, root), output-path safety (traversal/absolute/backslash), and Argo CD port parsing.
- Full suite: **253 passing**.

## Not included (follow-ups from #2325)
These are larger items intentionally left out of this PR:
- Helm chart / Kustomize source support with parameter overrides and patches.
- Improving the reliability/sorting of the "Browse GitHub repos…" list.

## Testing done
- `npm run lint` and `tsc --noEmit` pass clean.
- `npm test` — 253 passing.
- Manually exercised the flow in the Extension Development Host: renamed command appears, no pre-generate modal, explicit path prompts, optional README (both branches), non-modal success toast, correct folder creation (including nested output paths), and Argo CD UI port-forward on the configured port.


src/commands/aksArgoCD/argoCDDeployment.ts
bosesuneha: Can we return whitespace trimmed filename here? It currently accepts whitespace at the end of the filename

src/commands/aksArgoCD/argoCDDeployment.ts
```diff
@@ -95,262 +64,81 @@ function buildArgoCDAppYaml(params: {
...
-
----
-
-## 🔗 Step 3 — Connect this config repo to Argo CD
-
-\`\`\`bash
-# Register the GitOps config repo
-# For private repos add --ssh-private-key-path or --username / --password
-argocd repo add ${params.configRepoUrl}
+# then open https://localhost:8080
```
bosesuneha: the port 8080 is still hardcoded here, we can instead point the user to use the post deploy action command or look up the port first

src/commands/aksArgoCD/argoCDDeployment.ts
Tatsinnit: Good catch — fixed in a89efaee.

`promptRequired()` was validating against `v.trim()` but returning the raw input, so trailing whitespace passed validation and flowed into both the generated filename (`my-app .yaml`) and the manifest `metadata.name`, which Kubernetes then rejects.

```ts
-    return vscode.window.showInputBox({ ... });
+    const result = await vscode.window.showInputBox({ ... });
+    return result?.trim();
```

Used `?.trim()` rather than `.trim()` so the cancel contract is preserved — `undefined` stays `undefined` instead of collapsing to `""`.

src/commands/aksArgoCD/argoCDDeployment.ts
Tatsinnit: You are right, and this one was worse than it looks — thanks for catching it.

That `8080` directly contradicted the #2322 fix shipping in this same PR: the code resolves the port from `argocd-cm`, and then the generated README sent users straight back to 8080, re-breaking the Entra ID SSO redirect the PR set out to fix.

The README is written at manifest-generation time, when there is not necessarily a reachable cluster, so it cannot resolve the port itself. Went with your first suggestion (point at the command) and added the manual lookup as a fallback:

```diff
-kubectl -n argocd port-forward svc/argocd-server 8080:443
-# then open https://localhost:8080
+Use the extension's **AKS: Argo CD Post-Deploy Actions** command. It reads the
+port Argo CD is actually configured on and port-forwards to match — which
+matters when the UI is served behind Entra ID SSO, because the app
+registration's redirect URI is tied to that exact port.
+
+kubectl -n argocd get cm argocd-cm -o jsonpath='{.data.url}'
+kubectl -n argocd port-forward svc/argocd-server <port>:443
```

Added a regression test asserting the README contains no `8080` port-forward and does reference both the command and `argocd-cm`, so this cannot silently come back.

Fixed in a89efaee.

Tatsinnit:
Thanks for this, went back through every review comment and verified each against the code on the branch rather than assuming. Summary:

### Fixed now (a89efaee)

| Comment | Fix |
|---|---|
| @bosesuneha : trim the filename | `promptRequired()` validated `v.trim()` but returned the raw input, so trailing whitespace reached `my-app .yaml` and `metadata.name`. Now returns `result?.trim()`, keeping undefined-on-cancel. |
| @bosesuneha : 8080 still hardcoded | The generated README contradicted the #2322 fix in this same PR. Now points at **AKS: Argo CD Post-Deploy Actions** and shows the `argocd-cm` lookup, with a regression test. |

The 8080 one was the more significant of the two — the code resolved the port correctly, then the docs we generate told users to use 8080 anyway, which would have re-broken the SSO redirect this PR exists to fix.

### Already addressed earlier in the branch

Re-verified these are genuinely resolved in the current tree, not just claimed:

- **CodeQL — incomplete URL substring sanitization**: now asserted with `new RegExp(\`\\b${escapeRegExp(repoUrl)}\\b\`)` rather than a bare `includes()`.
- **Copilot — `application.yaml` in UX strings + cancel writing to workspace root**: strings use `<app-name>.yaml`; the output-path prompt aborts via `if (outputPath === undefined) return;`.
- **Copilot — `outputPath` traversal / separators**: `resolveOutputDirUri()` normalises `\` to `/`, rejects absolute paths (POSIX, drive-letter, UNC) and any `..` segment.
- **Copilot — plugin message filename**: `pluginResponses.ts` says `<app-name>.yaml`.
- **Copilot — docs still using the old command name**: docs are updated; the only remaining hits are stale `out/` build artifacts, which are not source.

### Tests

`266 passing, 0 failing`. `tsc --noEmit` and `eslint` clean.

docs/book/src/features/argocd-gitops-integration.md
```diff
@@ -59,29 +58,26 @@ The integration provides four commands, all prefixed with **AKS:**
 
 | Command | Where it appears | Description |
 |---------|-----------------|-------------|
-| **AKS: Create Argo CD GitOps Pipeline** | Command Palette, Explorer folder context menu | Scaffold an annotated Argo CD Application manifest in a config repo |
+| **AKS: Create Argo CD Application** | Command Palette, Explorer folder context menu | Generate an annotated Argo CD Application manifest pointing at your Kubernetes manifests |
```
pauldotyu:
```suggestion
| **AKS: Create Argo CD Application** | Command Palette, Explorer folder context menu | Generate an annotated Argo CD Application manifest pointing at a repo which contains Kubernetes manifests or a Helm chart to be deployed |
```

src/commands/aksArgoCD/argoCDDeployment.ts
```diff
@@ -95,262 +64,94 @@ function buildArgoCDAppYaml(params: {
...
-No admin password is required — your browser will redirect to the standard
-Microsoft sign-in page automatically.
-
-\`\`\`bash
-# Port-forward to access the UI locally (if no LoadBalancer ingress is configured)
-kubectl -n argocd port-forward svc/argocd-server 8080:443
-# Open: https://localhost:8080  — sign in with your Microsoft account
-\`\`\`
+If it is not installed, see the [Argo CD getting started guide](https://argo-cd.readthedocs.io/en/stable/getting_started/)
+or install the Azure-managed Argo CD extension.
```
pauldotyu: Should we also include the URL to the extension doc? https://learn.microsoft.com/azure/azure-arc/kubernetes/conceptual-gitops-argocd

src/commands/aksArgoCD/argoCDDeployment.ts
```diff
@@ -95,262 +64,94 @@ function buildArgoCDAppYaml(params: {
...
-| Resource | URL |
-|----------|-----|
-| Argo CD docs | <https://argo-cd.readthedocs.io/en/stable/> |
-| Getting started | <https://argo-cd.readthedocs.io/en/stable/getting_started/> |
-| App of Apps pattern | <https://argo-cd.readthedocs.io/en/stable/operator-manual/cluster-bootstrapping/> |
-| Sync waves (ordering) | <https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/> |
-| Azure Workload Identity | <https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview> |
-| Notifications | <https://argo-cd.readthedocs.io/en/stable/operator-manual/notifications/> |
+- [Argo CD docs](https://argo-cd.readthedocs.io/en/stable/)
+- [Getting started](https://argo-cd.readthedocs.io/en/stable/getting_started/)
```
pauldotyu, line 154: Maybe also include https://learn.microsoft.com/azure/azure-arc/kubernetes/conceptual-gitops-argocd

src/tests/suite/aksArgoCD/argoCDApplyApp.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        assert.strictEqual(parseArgoCDPort("https://localhost:9000"), 9000);
+    });
+
+    it("parses a port from a full http URL", () => {
+        assert.strictEqual(parseArgoCDPort("http://localhost:8081"), 8081);
+    });
+
+    it("parses a port from a bare host:port value (global.domain)", () => {
+        assert.strictEqual(parseArgoCDPort("localhost:9000"), 9000);
+    });
```
pauldotyu: Hey @Tatsinnit please correct me if I'm wrong, but this particular test doesn't look like it would read the global.domain value from a mocked configmap. If that wasn't the intent, then maybe we just remove the `(global.domain)` from the test description.

pauldotyu (commented):
Just some small nits. Overall, looks great!

docs/book/src/features/argocd-gitops-integration.md
Tatsinnit: Applied your suggestion in f0f391a9 — thanks.

Worth noting for anyone reading later: the generated Application sets only `repoURL` + `path`, and Argo CD auto-detects and renders a Helm chart found at that path, so the wording is accurate. The "Helm support" listed as a follow-up in the PR description refers specifically to extension-driven parameter overrides (`spec.source.helm`), not basic chart rendering.

src/commands/aksArgoCD/argoCDDeployment.ts
Tatsinnit: Good call — added in f0f391a9. It was inconsistent that the OSS getting-started guide was linked while the Azure-managed extension was only named:

```diff
-or install the Azure-managed Argo CD extension.
+or install the [Azure-managed Argo CD extension](https://learn.microsoft.com/azure/azure-arc/kubernetes/conceptual-gitops-argocd).
```

src/commands/aksArgoCD/argoCDDeployment.ts
Tatsinnit: Added to **Further reading** in f0f391a9:

```diff
 - [Argo CD docs](https://argo-cd.readthedocs.io/en/stable/)
 - [Getting started](https://argo-cd.readthedocs.io/en/stable/getting_started/)
+- [GitOps with Argo CD on AKS and Arc-enabled Kubernetes](https://learn.microsoft.com/azure/azure-arc/kubernetes/conceptual-gitops-argocd)
```

src/tests/suite/aksArgoCD/argoCDApplyApp.test.ts
Tatsinnit: You are right, ❤️  that test never reads a ConfigMap. It only exercises `parseArgoCDPort` against a bare `host:port` string; the `(global.domain)` was meant to convey *"this is the shape `global.domain` has"* (bare host:port, unlike `url` which is a full URL), but it reads as if the ConfigMap lookup is under test.

Rather than just drop the parenthetical, your comment pointed at a real gap: the `url` → `global.domain` fallback in `detectArgoCDConfiguredPort()` had **no test at all**. Fixed both in f0f391a9.

Renamed for accuracy:
```diff
-it("parses a port from a bare host:port value (global.domain)", ...)
+it("parses a port from a bare host:port value, the form global.domain uses", ...)
```

And added real coverage with a faked kubectl, including the case I think is easiest to regress — `url` set but with no explicit port, which must still fall through to `global.domain`:

```ts
it("prefers the port from argocd-cm 'url'", ...)                              // 9000
it("falls back to 'global.domain' when 'url' is not set", ...)                   // 9443
it("falls back to 'global.domain' when 'url' carries no explicit port", ...)     // 8443
it("defaults to 8080 when neither key is configured", ...)                 // 8080
```

Tatsinnit:
All remaining review comments addressed in f0f391a9. Thanks @pauldotyu and @bosesuneha.

### Substantive

**`manifestPath` was unvalidated** — this came from Copilot's *suppressed low-confidence* section, so it had not been surfaced. `appName` and `namespace` both had validators; `manifestPath` had none, despite being the value interpolated **unquoted** into `spec.source.path`. It now rejects absolute paths, `..` traversal, control characters, and values that would break the YAML scalar (leading indicators like `*`/`{`/`-`/`#`, or an embedded `": "`), and normalises `\` to `/` since git paths use forward slashes.

Copilot overstated one part — `showInputBox` is single-line, so newline injection was never possible — but the core point stood.

**Missing coverage on the port fallback** — @pauldotyu's note about the misleading `(global.domain)` test name led to a real gap: `detectArgoCDConfiguredPort()` had no tests. Added four, including `url` set but portless falling through to `global.domain`.

### Nits

- Stale `localhost:8080` in the `openArgoCDUI` doc comment (also from the suppressed set) — same class of prose-vs-code drift as the README issue.
- Applied the suggested command-table wording.
- Linked the Azure-managed extension to Microsoft Learn where it is mentioned and under Further reading.

### Tests

`278 passing, 0 failing` (was 266 — 12 new). `tsc --noEmit` and `eslint` clean.

---

One observation worth recording: the highest-value comment on this PR — the hardcoded `8080` in the generated README, which would have reintroduced the exact Entra SSO redirect mismatch #2322 exists to fix — was in Copilot's **suppressed low-confidence** section, and only reached us because @bosesuneha found it independently. Two of the three suppressed comments turned out to be legitimate. Worth expanding that section on future reviews.

package-lock.json
```diff
@@ -12569,7 +12411,7 @@
             "version": "1.2.5",
             "resolved": "https://registry.npmjs.org/word-wrap/-/word-wrap-1.2.5.tgz",
             "integrity": "sha512-BN22B5eaMMI9UMtjrGd5g5eCYPpCPDUy0FJXbYsaT5zYxjFOckS53SQDE3pWkVoWpHXVb3BrYcEN4Twa55B5cA==",
-            "devOptional": true,
+            "dev": true,
```
bosesuneha, line 12414: is this change intentional?

bosesuneha (approved):

PR #2372, opened 2026-08-05
docs: correct menu paths, add generated reference, fix navigation

bosesuneha:
Corrects the documented cluster menu paths, adds a generated reference section, and fixes navigation. Rebased on `main`; #2370 and #2371 have merged.

Touches `package.json` — two command titles, see section 4.

## 1. Generated reference and navigation

Adds a `Reference` section generated from `package.json` by `scripts/generate-docs-reference.js`, with a `DO NOT EDIT` header: commands and their menu placement, settings, pinned versions. Menu placement is not derivable by hand — 66 commands, two menu modes, nested up to four levels.

`SUMMARY.md` listed `show-properties-azureportal-start-stop.md` four times under four labels and `inspektor-gadget.md` twice, so three sidebar entries led to a page that did not match the label clicked. Collapsed to one each; no URLs change.

`features.md` was a stale `2.1.0` release spotlight with 20 generically captioned screenshots, linking to 2 of its 25 feature pages. Rewritten as an index.

`README.md` re-listed all 25 feature pages by hand — a third copy after `SUMMARY.md` and `features.md`, already drifted. Replaced with pointers to the index and the generated reference.

## 2. Menu path corrections

`aks.simplifiedMenuStructure` has defaulted to `true` since #2168, grouping cluster commands under **Develop & Deploy**, **Troubleshoot & Diagnose** and **Manage Cluster**. The guides still described the classic flat menu, so readers were told to look for items now one level down.

Paths come from the generated reference, not from reading the UI, and are written as `Right-click your AKS cluster > Submenu > Command` so `docs-check` can validate them.

Beyond the missing group level:

| Page | Was | Actually |
|---|---|---|
| `aks-compare-cluster` | cluster node | subscription node |
| `show-properties-azureportal-start-stop` | "Show in Azure Portal" | "Show In Azure Portal" |
| `manage-cluster-operations` | "Managed Cluster Operations" | classic label; default is "Manage Cluster" |
| `kaito-*` | "Deploy an LLM with KAITO" | "Deploy a LLM with KAITO" |
| `container-assist-integration` | `AKS: Run Container Assist (Preview)` | no menu contribution; replaced with the three real commands |
| `inspektor-gadget` | "moved to the main context menu" | describes the classic layout, now backwards |

"Abort Last Operation" was listed as a cluster operation. It is a webview action in `ClusterPropertiesPanel.ts`, not a command, and there are two of them. `simplified-menu-structure.md` advised keeping the default `false` while stating above that it was `true`.

[...]

PR #2168, opened 2026-05-21
feat(menu): default to structured menu with one-click switch back to …

Tatsinnit:
# Menu Restructure— Structured menu by default with one-click classic fallback

## Summary

**Testing VSIX:**  
[vscode-aks-tools-1.7.8-restructured-menu.vsix.zip](https://github.com/user-attachments/files/28363539/vscode-aks-tools-1.7.8-restructured-menu.vsix.zip)

Promote the role-based **Structured Menu** (Develop & Deploy, Troubleshoot & Diagnose, Manage Cluster) to be the default AKS cluster context-menu experience in release **`v2.0.0`**, and ship a mutually exclusive **"Take me back to Classic Menu"** command so users can flip between modes from the menu itself — no settings.json edits required.

---

The default user-visible cluster menu layout changes. No command IDs, settings keys, or APIs are removed — but the out-of-the-box right-click experience is different enough that semver-major signals it clearly to users and downstream consumers.

---

### 1) Default UX
- `aks.simplifiedMenuStructure` default: **`false` → `true`**.
- Right-click on an AKS cluster now collapses 15+ top-level items into:
  - 🚀 **Develop & Deploy**
  - 🔍 **Troubleshoot & Diagnose**
  - ⚙️ **Manage Cluster**
- Always-visible top-level: **Show in Portal**, **Cluster Properties**.

### 2) New "menu mode" toggle commands
| Command ID | Title | Visible when |
|---|---|---|
| `aks.switchToClassicMenu` | **AKS: Take me back to Classic Menu** | Structured menu active |
| `aks.switchToStructuredMenu` | **AKS: Switch to Structured Menu** | Classic menu active |

- Surfaced at the **top** of the AKS cluster context menu (group `0_menuMode@1`).
- Also surfaced in the **Command Palette**.
- Both surfaces are gated on `config.aks.simplifiedMenuStructure` so the two commands are **mutually exclusive everywhere** — only one is visible at a time, never both, never neither.
- Handler updates the setting at `Global` scope and prompts a window reload.

### 3) Version Target
- When we will release we will bump `package.json` and `package-lock.json`: `1.7.8` → **`2.0.0`**.

---

## Files Changed

- **New:** [src/commands/menuToggle/menuToggle.ts](src/commands/menuToggle/menuToggle.ts) — `switchToClassicMenu` / `switchToStructuredMenu` handlers.
- [package.json](package.json) — version bump, default flip, two new command contributions, `view/item/context` + `commandPalette` menu entries.
[...]

PR #2373, opened 2026-08-06
docs: correct factual errors and remove stale screenshots

bosesuneha:
Corrects documentation that did not match the shipped extension, removes stale screenshots, and brings the release notes and telemetry disclosure up to date.

27 markdown files, 23 images deleted. Every claim was checked against `package.json`, `src/` or `resources/` rather than taken from the existing text.

## 1. Factual corrections

| Page | Was wrong |
|---|---|
| `merge-save-kubeconfig` | Documents two commands as ours. Both come from `ms-kubernetes-tools.vscode-kubernetes-tools` |
| `draft-integration` | Listed Draft Dockerfile and Draft Deployment as palette commands. Neither is in `contributes.commands` — they open only from links in the GitHub workflow screen. `aksDraftValidate` and `draftArgoCDDeployment` were undocumented |
| `container-assist-github-workflow` | Five Action versions stale. The templates pin by commit SHA, so tags in prose go stale on every bump; the tables now name actions without versions and link to the generated [Pinned versions](docs/book/src/reference/pinned-versions.md) |
| `container-assist-github-workflow`, `container-assist-integration` | Document an "Annotate namespace" step. No template contains it; `workflowTemplate.test.ts:223` asserts its absence |
| `container-assist-azure-resources` | "9 role assignments" summed two mutually exclusive paths. Maximum is 5 — one in stage 1, four in stage 2 |
| `aks-diagnostics` | Two detector titles renamed, one reworded, `Storage` missing |
| `run-kubectl-command`, `k8s-api-health-points` | Labels did not match `webviewDefinitions/kubectl.ts`. Both pages now show the command each entry runs |
| `argocd-gitops-integration` | Apply runs with `--validate=false` (`argoCDApplyApp.ts:745`), omitted. The Copilot integration is an Azure AI Agent plugin, not a chat skill |
| `simplified-menu-structure` | Described the classic menu as a flat list of every command. Seven submenus sit on the cluster node |
| `inspektor-gadget` | Three sections named a command without its menu path; one heading did not match the menu label |
| `telemetery.md`, `README.md` | `telemetry.enableTelemetry` was deprecated in VS Code 1.61. The setting is `telemetry.telemetryLevel` set to `off` |

Plus typo fixes: `auomatically`, `ip-resrouces`, `form`/`from`, `yout`/`your`, `Successfull`, `Kuberentes`, a doubled `the the`.

## 2. Screenshot cleanup

23 images deleted. `docs/book/src/resources/` goes 100 → 77, 12.6 MB → 9.7 MB.

11 were referenced by no page. The other 12 were still referenced but show the **classic** cluster menu; the default switched to the grouped menu in #2168 (2026-05-30) and all 12 were last touched before that. The text breadcrumbs added in #2372 replace them and are validated by `docs:check`.

Webview and result-panel screenshots are kept — they show output that text cannot convey.

## 3. Release notes

Notes stopped at **2.1.0** while the extension shipped **2.5.0**. Replaced `whats-new-2.1.0.md` with `whats-new-2.5.0.md` covering 2.2.0 → 2.5.0, following the convention of keeping only the current release's file. Content derived from `git log 2.1.0..2.5.0` — **please check the highlights read the way you want them to.**

## 4. Telemetry disclosure

| Issue | Evidence |
|---|---|
| Per-message webview telemetry undisclosed | `BasePanel.ts:162-164` emits an event per webview message |
| `clusterCreationSuccess` filed under Copilot handlers | `CreateClusterPanel.ts:300` uses the plain `command` event for the normal flow |
| Application Insights transport unmentioned | `reporter.ts` uses `@vscode/extension-telemetry` with `meta.aiKey` |
| `README.md` and `telemetery.md` disclosed different data | README listed only "which commands are executed" |

Adds an explicit "what we do not collect" section.

## Verification

`docs:check` **0 errors** across `identifiers`, `orphans`, `menu-paths` and `menu-syntax`; `docs:reference:check` up to date; lychee 0 errors; Prettier clean; `make -C docs/book build` succeeds on the pinned mdBook 0.4.27.

Note that `docs:check` currently **fails on `main`** with 2 `menu-syntax` errors in `merge-save-kubeconfig.md`. This PR fixes both, so it is a prerequisite for wiring the docs checks into CI.

Remaining: 15 `coverage` warnings for commands with no prose yet. Follow-up, not in this PR.


docs/book/src/features/simplified-menu-structure.md
```diff
@@ -56,9 +48,23 @@ Both commands are available in:
...
 After running either command, VS Code prompts you to reload the window. The new menu layout takes effect after the reload.
 
-## Backward compatibility
+## Changing it in Settings instead
+
+If you would rather set this directly, the menu is controlled by
+`aks.simplifiedMenuStructure`. Set it to `false` for the classic menu, where commands
+and their submenus — **Run AKS Diagnostics**, **Deployment Tools**, **Managed Cluster
+Operations**, **Troubleshoot Network Health**, **Troubleshoot Resource Utilization**,
+**Improve security of my cluster**, **Deploy a LLM with KAITO** and **Collect TCP Dump**
```
Tatsinnit: `Collect TCP Dump` doesn't belong in this list — it's the label of `aks.tcpDataCollectionSubMenu`, which has **zero references in any menu bucket**. It's the dead submenu the generator added in #2371 already reports:

```
dead submenus (declared, never referenced): aks.tcpDataCollectionSubMenu
```

In both menu modes TCP shows up as *commands* under Troubleshoot Network Health, never as a top-level submenu:

```
simplified: cluster|Troubleshoot & Diagnose > Troubleshoot Network Health > Collect TCP Dumps
classic:    cluster|Troubleshoot Network Health > Collect TCP Dumps
```

So someone switching to the classic menu won't find a `Collect TCP Dump` submenu. Dropping it makes the count seven — worth updating the PR body too, which says "Eight submenus sit on the cluster node".

Worth noting why `docs:check` passed here: this is a comma-separated list of bold labels rather than a `**A** > **B**` breadcrumb, so `menu-paths` never parses it and `menu-syntax` doesn't match the pattern either. That's a gap in the checker rather than anything you did wrong — and the same blind spot I raised on #2371, so it might be worth a follow-up there.

Tatsinnit (dismissed):
Checked this out and verified each claim against `package.json`, `src/` and `resources/` rather than taking the description at its word. Nearly all of it holds, and the citations are precise enough to be falsifiable — which made checking straightforward.

## Verified

| Claim | Checked against |
|---|---|
| Draft Dockerfile/Deployment unreachable | Registered at `extension.ts:147-148`, absent from `contributes.commands`. Exact. |
| Action versions stale | Container-assist templates pin `checkout@v7.0.0`, `login@v3.0.0`, `use-kubelogin@v1.3`, `aks-set-context@v5.0.0`, `k8s-deploy@v6.0.0`. Doc now matches. |
| "Annotate namespace" doesn't exist | `workflowTemplate.test.ts:223` — `assert.ok(!result.includes("Annotate namespace"), ...)`. Exact line. |
| "9 role assignments" -> max 5 | User-ns path = 3, managed-ns path = 4, mutually exclusive, +1 in stage 1 -> 5. |
| `aks-diagnostics` detectors | All 7 match `package.json`, including the previously-missing `Storage`. |
| kubeconfig commands aren't ours | No `aks.*kubeconfig` command exists. |
| Argo CD apply uses `--validate=false` | `argoCDApplyApp.ts:745`. |
| Only **Check Argo CD Status** on the cluster menu | `cluster\|Develop & Deploy > AKS: Check Argo CD Status` is the only Argo entry. |
| Telemetry gaps | `BasePanel.ts` fires `sendTelemetryEvent("command", ...)` per webview message; `CreateClusterPanel.ts` sets `eventName = commandId === "aks.createCluster" ? "command" : "aks.ghcp"`; `reporter.ts` -> `new TelemetryReporter(meta.aiKey)`. All three exact. |
| `telemetry.enableTelemetry` deprecated | Correctly replaced with `telemetryLevel` in both README and `telemetery.md`. |
| 23 images deleted, 9.4M | `docs/book/src/resources` goes 81 -> 58. Exact. |
| `docs:check` 0 errors, `reference:check` up to date | Both reproduce. |

The telemetry disclosure work in particular is the kind of thing that usually goes unwritten — good to see the per-message event and the AI transport called out.

## One correction needed

Left inline: the classic-menu submenu list includes `Collect TCP Dump`, which is a declared-but-unreferenced submenu. The real count is seven, not eight.

## Minor

The image counts "156 down to 133" don't reproduce for me — I measure 81 -> 58 in `docs/book/src/resources`, and 109 -> 86 repo-wide tracked. The **delta of 23 and the 9.4M are both exact**, so it looks like only the absolutes are off, possibly counted with a built `docs/book/book/` present. Description-only; the change itself is right.

## Verdict

Approving. Everything else checks out; the one item above is a single-line fix.

Separately, and not for this PR: `aks.tcpDataCollectionSubMenu` being declared but never referenced is a **product** bug rather than a docs one — dead contribution surface in `package.json`. #2372 left it alone as a product decision, which is fair, but it's worth its own issue.

docs/book/src/features/simplified-menu-structure.md
bosesuneha: Both addressed in e6105982. `Collect TCP Dump` is gone, so it's seven submenus now, in the page and in the body.

Your point about why `docs:check` passed goes wider than menu syntax. Bold command labels aren't checked against `contributes.commands` at all. I'll track that in #2382.

Tatsinnit (approved):

PR #2376, opened 2026-08-07
fix(kickstart): prefer az acr build and safeguards validation in cloudshell

davidgamero:
kickstart agent should always prefer az acr build command to use remote builds instead of docker daemon local builds

add deployment safeguards validation to skills for manifest generation in kickstart


bosesuneha (dismissed):

skills/kickstart-discover/SKILL.md
```diff
@@ -26,6 +26,16 @@ Never assume a flat repo. Apps often live in nested or monorepo layouts (`src/<s
...
 Use `codebase`/`search` to confirm each path actually exists — do not infer it from the language alone. Surface this map to the user and let them correct it before proceeding.
 
+## Build environment
+
+All container images are built with **`az acr build`** — server-side on the ACR remote task builders. Never `docker build`. Kickstart runs in Azure Cloud Shell, which has no Docker daemon, and a single remote build path keeps behavior identical everywhere.
+
+Cloud Shell constraints to honor:
+
+- Clone into `~/clouddrive/` (persistent, ~5 GB) rather than the ephemeral home dir.
+- `az acr build` needs the Phase 2 ACR to exist and the caller to hold `AcrPush` (or `Container Registry Tasks Contributor`) — see `/kickstart-handoff`.
```
Tatsinnit: Purely as some thoughts 💭 am I missing a planned Cloud Shell mode here? I went looking and couldn't find one — `rg -in "cloudshell|clouddrive"` comes back empty across the repo before this PR, and the launch path reads as local to me: `extension.ts:117` -> `kickstartLaunch.ts:27` calls `workbench.action.closePanel`, `kickstartChat.ts:39` opens local Copilot chat, and `kickstart-handoff:52` suggests `brew install Azure/kubelogin/kubelogin`.

If it does run locally, the `~/clouddrive/` line on 35 might send people somewhere that doesn't exist — Discover also covers the "use my current workspace" path, so it isn't only the sample-clone flow that sees it.

Worth noting the existing text at `kickstart-generate:48` already justified preferring ACR on "no local daemon available" grounds, which gets you the same outcome without depending on where it runs. But entirely possible there's a direction here I don't have visibility into.

skills/kickstart-safeguard-checklist/SKILL.md
```diff
@@ -6,9 +6,88 @@ disable-model-invocation: true
...
+- **Scope**: labels only. The policy does not inspect `metadata.annotations` — the safeguards docs name the annotation field explicitly where they mean it (e.g. the AppArmor rule). Even so, don't put `kubernetes.azure.com/*` in an annotation; the prefix is reserved by convention regardless of which field is enforced.
+- `azure.workload.identity/*` labels and annotations are **not** covered by this rule and are required for Workload Identity.
+- [ ] Pass / Fail
+
+### A5: allowed-images-only
+- **Check**: Every image resolves to a registry the cluster permits — normally the Phase 2 ACR (`<acr>.azurecr.io/...`). Flag any third-party image (`docker.io/...`, `ghcr.io/...`) that must first be brought in with `az acr import`.
+- [ ] Pass / Fail
+
+### A6: reserved-system-pool-taints  *(mutating)*
+- **Check**: No app workload declares a `CriticalAddonsOnly` toleration, and no user node pool config sets that taint. AKS uses it to keep customer pods off the system pool.
```
Tatsinnit: I went digging on this one specifically because the rest of Part A is sourced so tightly, and I think A6 may be node-pool-only rather than covering pod tolerations.

The constraint template (`Azure/azure-policy`, `samples/KubernetesService/restricted-taints/template.yaml`) binds:

```rego
node := input.review.object
taints := {x | x = node.spec.taints[_].key}
not is_system_pool(node)
```

with `is_system_pool` keyed off the node's `kubernetes.azure.com/mode` label. Both fixtures in that directory are `kind: Node` with `spec.taints`, and the mutation policy's description says the taint "should not be applied to user pools". I couldn't find anywhere that pod tolerations get evaluated.

If that reading is right, the first half of this check ("No app workload declares a `CriticalAddonsOnly` toleration") wouldn't have a policy behind it, and A6 would be effectively N/A for generated manifests — which would mean A2 and A3 are the two carrying the mutating weight.

That said, the public constraint template may not be the whole story on an Automatic cluster. If you've seen it actually strip a toleration in practice I'd genuinely like the pointer, and I'll happily be wrong here.

skills/kickstart-generate/SKILL.md
```diff
@@ -27,7 +27,74 @@ Load these for detailed patterns as you author each artifact:
...
+    - maxSkew: 1
+      topologyKey: kubernetes.io/hostname
+      whenUnsatisfiable: ScheduleAnyway
+      labelSelector:
+        matchLabels: { app: <service> }
+  ```
+- **Readiness *and* liveness probes** on every container — use the real health path and port from the structure map, never a guessed `/healthz`.
+- **Unique Service selectors** — each Service must select exactly one workload. In a monorepo do **not** reuse `app: <appName>` across services; use `app: <serviceName>` (or `app.kubernetes.io/name` + `app.kubernetes.io/component`) so no two Services overlap.
+- **No AKS-specific labels** — never set `kubernetes.azure.com/*` labels on your own objects. (`azure.workload.identity/*` labels and annotations are fine and required.)
+- **No reserved system-pool taints/tolerations** — do not add a `CriticalAddonsOnly` toleration to app pods; it would place them on the system pool and AKS strips it anyway.
```
Tatsinnit: Small one, related to the A6 thread: I don't think "it would place them on the system pool" holds — a toleration only permits scheduling onto a tainted node, it doesn't attract pods there. You'd still need a `nodeSelector` or node affinity for them to actually land on the system pool. So a stray toleration on its own probably wouldn't move anything.

Not a big deal in practice since you don't want app pods tolerating it either way, but since this is an instruction the agent acts on, the reasoning it's given might be worth tightening.

skills/kickstart-generate/SKILL.md
```diff
@@ -37,18 +104,27 @@ Load these for detailed patterns as you author each artifact:
...
-   - Otherwise build in ACR (also catches missing `COPY` sources): `az acr build --registry <acr> --image kickstart-validate/<svc>:check -f <dockerfilePath> <buildContext>`.
-2. **Inspect contents** (when built locally): `docker run --rm kickstart-validate-<svc>:check ls -la <workdir>` — confirm the entry point and expected files landed where the app runs from. A build that succeeds but places files in the wrong dir is exactly the failure this step catches.
-3. If the build fails or the entry point is missing, fix the Dockerfile/paths and rebuild — do not proceed to Review with a broken image.
+1. **Build in ACR** from the service's own build context. This also catches missing `COPY`/`ADD` sources:
+   ```bash
+   az acr build --registry <acr> --image kickstart-validate/<svc>:check -f <dockerfilePath> <buildContext>
+   ```
+2. **Assert the entry point at build time.** Add to the Dockerfile's final stage so a wrong path fails the ACR build itself:
+   ```dockerfile
+   RUN test -f <entrypointPath>
```
Tatsinnit: Does this hold for a distroless or `FROM scratch` final stage? Neither has a shell, so `RUN` would fail the build outright — and those are common finals for Go and Rust, both of which show up in the Store Demo sample this skill recommends. (The sample's own Dockerfiles land on `alpine:3.22` and `debian:bookworm-slim`, so the sample itself is fine; generated ones might not be.)

Also worth flagging that `test -f` proves the path exists, whereas the `docker run --rm ... ls -la <workdir>` it replaces was catching the case the old text explicitly called out — build succeeds but files land in the wrong directory. I'm not sure there's a clean server-side equivalent, so this may just be a tradeoff worth naming rather than something to fix.

skills/kickstart-deploy/SKILL.md
```diff
@@ -11,12 +11,16 @@ Deploy using Azure CLI and `kubectl`. Execute each step via `run_in_terminal`, c
 ## Steps
 
 1. **Build and push**: `az acr build --registry <acr> --image <image>:<tag> -f <dockerfilePath> <buildContext>`
+   `az acr build` runs **server-side on the ACR remote task builders** — always use it, never `docker build` + `docker push`. Kickstart runs in Azure Cloud Shell, which has no Docker daemon, and a single remote build keeps the pushed digest and the deployed image identical.
    Use the build context and Dockerfile path from the structure map — never assume repo root (`.`). For monorepos, build each service from its own context. Tag with a version (e.g. v1.0.0), never `:latest`.
+   Requires `AcrPush` or `Container Registry Tasks Contributor` (verified in Pre-Deploy Check). Keep the command in the foreground so its streamed log holds an idle Cloud Shell session open, and make sure `.dockerignore` excludes bulk (the whole context is uploaded on each build).
```
Tatsinnit: "(verified in Pre-Deploy Check)" is accurate here in Phase 7 — but Phase 4 now runs `az acr build` for validation too, and per `kickstart-handoff:89` that wants `Container Registry Tasks Contributor`, which isn't probed until Phase 6.

As far as I can tell it's never granted either: `kickstartProvision.ts:337,740` grant only `AcrPull` to the cluster's kubelet identity, and `kickstartAzureBackend.ts:815` is a `stage.warn(...)` rather than a gate. So someone without that role might hit a 403 at Phase 4, two phases before the remediation command and `/kickstart-pim-activation` escalation are loaded.

I realise dropping the local fallback was deliberate, so I'm not suggesting bringing `docker build` back — would moving just that one ACR permission probe earlier be easier than keeping a fallback path?

Tatsinnit (dismissed):
Approving, (sharing some thoughts for your eyes please) the direction is right and the Part A rewrite is genuinely good work. I checked the numbers against MS Learn and they're exact: A2's `500m`/`2048Mi` defaults and `100m`/`100Mi` minimums, the QoS capping, A3's weight-100 anti-affinity and `maxSkew: 1`/`ScheduleAnyway`, the `app` -> `app.kubernetes.io/name` -> `default-antiaffinity-applabel` priority, A8 checking provisioner rather than name, fail-open Gatekeeper, all-or-nothing enforcement. Splitting policy-enforced (Part A) from best-practice (Part B) is a real improvement over what was there.

**Caveat on everything below: I'm not 100% sure on these, and I don't have full context on where kickstart is headed.** They're things I couldn't reconcile while reading, not blockers — please treat them as "worth a second look" rather than findings, and push back freely where I've got it wrong. Not gating the merge on any of them.

One that has no line to attach to: `AGENTS.md:17-18` and `kickstart-guide.md:77-78` still say "build & inspect the image" / "verify image builds", which reads like the `docker build` + `docker run ... ls` loop this PR replaces. Since `AGENTS.md` gets loaded as instructions it might be worth pulling those two files in — though possibly that's already queued behind this.

Tatsinnit (approved):
Re-reviewed after `a809fc8e`. You've addressed all of it, and in a couple of places more thoroughly than I'd suggested.

**Cloud Shell framing** — gone from `kickstart-discover`, `kickstart-deploy`, `kickstart-generate` and the reviewer agent, and the `~/clouddrive/` instruction with it. "Avoids depending on a local Docker daemon" carries the change without asserting where kickstart runs, which is exactly right.

**A6** — the correction is more precise than my comment was:

> **Applies to**: Node and node-pool configuration, not Pod tolerations. The policy evaluates node `spec.taints` and excludes system pools by their `kubernetes.azure.com/mode` label.

That matches the constraint template exactly, including the `kubernetes.azure.com/mode` exclusion I hadn't mentioned. The table row, the reviewer agent's mutating-set, and the `deploy` "toleration is stripped" line are all consistent now.

**Toleration semantics** — "permits, but does not force, scheduling onto a matching tainted node" is the correct statement.

**`RUN test -f`** — removed, and the replacement picks up both halves of what I raised: no shell in distroless/`scratch`, *and* that path-existence never proved the entry point landed in the right directory. Reconciling final-stage `COPY`/`ADD` against `WORKDIR` and `CMD` is a better check than the one it replaces.

**Phase ordering** — the ACR permission preflight as step 1 of Generate, wired to `aks.checkDeploymentPermissions` with the `acr-push`/`acr-tasks` probes and PIM escalation, resolves the 403-two-phases-early problem properly. Better than the fallback I'd floated.

**`AGENTS.md` / `kickstart-guide.md`** — both updated, so nothing contradicts the new flow.

One left, and it's the one I flagged as lowest-confidence, so genuinely fine to leave:

`kickstart-generate/SKILL.md:50` still reads "`app.kubernetes.io/managed-by` is the convention already used elsewhere in this extension". As far as I can find there's one authored use — `resources/yaml/argocd-application.template.yaml:30`, value `argocd` — plus read-only selectors for `Microsoft.ArgoCD`. The extension's documented tool-provenance schema is `aks-project/deployed-by: vscode` (`container-assist-integration.md:226,245`), described as shared with aks-desktop. Not blocking, and I have no strong view on which prefix wins — just that the justification sentence claims more precedent than exists. Happy for it to land as-is.

Approving.

PR #2382, opened 2026-08-12
docs: document the checks that gate a pull request

bosesuneha:
`CONTRIBUTING.md` was CLA and code-of-conduct boilerplate only. No build steps, no test steps, and no mention of the checks that actually fail a PR. It was also **byte-identical** to `docs/book/src/contributing.md`. Expanded the root file, which GitHub surfaces natively in the PR and issue UI, and reduced the book copy to CLA, code of conduct and links, so the two are no longer duplicates waiting to drift.

Four commits, reviewable in order:

| | Commit |
|---|---|
| 1 | `docs:` document the checks that gate a pull request |
| 2 | `docs:` correct the gate list and add a `titles` check |
| 3 | `ci:` run the documentation checks on pull requests |
| 4 | `docs:` describe the Docs Check workflow now that it exists |

Commit 3 is the only workflow change and is self-contained, so it can be reverted on its own if you would rather not adopt the gate yet.

## The gate list

| Workflow | Command | Runs on |
|---|---|---|
| Build | `npm run lint:all`, `npm test` | every PR |
| Fuzzing Tests | `npm run test:fuzz` | every PR |
| Docs Check | `npm run docs:check`, `docs:reference:check` | PRs touching `docs/`, `scripts/`, `package.json`, `package.nls.json`, `resources/` |
| PR Size Checker | — | every PR, 1200 lines |
| Prettier Check | `npm run prettier-format` | **only PRs touching `.ts` / `.tsx`** |

Fuzzing was missing entirely. Prettier needs the caveat: `format-check.yml` triggers on `**/*.ts` and `**/*.tsx`, but the job it runs checks `.json`, `.css` and `.md` too, so a docs-only PR is never format-checked.

## Development page

| Issue | Reality |
|---|---|
| No Node version documented anywhere | CI pins **22** (`build.yml:48`). No `engines.node`, no `.nvmrc`, so nothing enforces it locally |
| `install:all` requirement undocumented | Both installs pass `--legacy-peer-deps`; a bare `npm install` fails on peer resolution |
| `prepare`, `package`, `test:fuzz`, `lint`, `eslint-inspector` undocumented | Including the CI gates |
| `webpack` described as packaging the extension | It does not — `package` does |
| VSIX section said to run `vsce package` | The repo wraps it as `npm run package`, which is what CI uses |
| `test-compile` description incomplete | `prepare-test-assets.js` copies the CA skills into `dist/skills` as well as the deploy template |
| `npm test` needs a display on Linux | CI wraps it as `xvfb-run -a npm run test` |
| `[skip pr-size]` | The workflow reads the most recent commit, not the whole branch |

## New `titles` check

`identifiers` validates command IDs and `menu-paths` validates breadcrumbs, but a command named by its **title alone** was checked by neither. A page could keep using a name the extension had dropped and still pass with zero errors.

Not hypothetical — the Draft page named `AKS: Create Argo CD GitOps Pipeline` for months after #2353 renamed it, through a clean `docs:check` run. With this check:

```
E  docs/book/src/features/draft-integration.md:14
   no command is named "AKS: Create Argo CD GitOps Pipeline"
```

Scoped to bold labels beginning `AKS:`. Any bold string could be a command name, but most are ordinary emphasis and guessing produces false positives that train people to ignore the check. Both the palette form (`AKS: Create a GitHub Workflow`) and the bare menu form (`Create a GitHub Workflow`) are accepted.

It found a fabricated command name in the first draft of its own documentation.

## Docs Check workflow

`docs:check` and `docs:reference` were not invoked by anything. `website.yaml` only builds the book, and only on push to `main`, so the drift they exist to prevent could recur on the next PR — and did, per the example above.

No install step; the scripts use only node builtins. Runs in ~20s. Was blocked until #2373 fixed the two `menu-syntax` errors on `main`; `main` is green as of 8a4367f2.

Supersedes #2410, which is the same commit. Kept here so the workflow and the page describing it land together, rather than leaving one of them wrong in between.

## Verification

```
docs:check            0 errors across all six checks
test:scripts          63 passing (11 new)
docs:reference:check  up to date
prettier              clean
lychee                209 OK, 0 errors
mdbook                builds
size                  356 / 1200
```


docs/book/src/development/development.md
```diff
@@ -1,28 +1,49 @@
...
 
-- [`dev-webview`](./webview-development.md#developing-the-ui): for concurrent development/debugging of webview UX.
-- `build-webview`: bundles and minifies the webview UX for consumption by the extension.
-- `webpack`: builds and packages the extension.
+- [`dev:webview`](./webview-development.md#developing-the-ui): for concurrent development/debugging of webview UX.
+- `build:webview`: bundles and minifies the webview UX for consumption by the extension.
+- `webpack`: builds the webview UX and then bundles the extension in production mode. This does not produce a `vsix`; use `package` for that.
+- `package`: packages the extension into a `vsix` (`vsce package --no-dependencies`).
 - `test`: runs automated tests.
+- `test:fuzz`: runs the fuzzing test suite.
```
Tatsinnit, line 26: Worth catching when you rebase: this branch's `package.json` has 17 scripts, but `main` has 21. These four landed with #2371 and aren't documented here:

```
docs:check              node scripts/docs-check.js
docs:reference          node scripts/generate-docs-reference.js
docs:reference:check    node scripts/generate-docs-reference.js --check
test:scripts            mocha 'scripts/**/*.test.js'
```

Since the PR is `CONFLICTING` and stacked on #2373, a rebase is required anyway — and at that point this page drops from 0 undocumented scripts to 4, which undoes the main thing the change achieves. `test:scripts` probably belongs right here next to `test` and `test:fuzz`; the `docs:*` three fit either here or under a docs heading.

docs/book/src/development/development.md
```diff
@@ -1,28 +1,49 @@
...
+
+## Checks that gate a pull request
+
+Run these before pushing; each has a corresponding CI job that will fail the PR.
+
+- `prettier-format`: formats the repository with Prettier. The **Prettier Check** workflow fails if anything is unformatted.
+- `lint:all`: lints both the extension and `webview-ui`. The **Build** workflow runs a lint step.
+- `lint-fix:all`: the same, applying autofixes.
+
+Pull requests are also limited to **1200 changed lines** by the **PR Size Checker** workflow. Include `[skip pr-size]` in the commit message only when a larger change is genuinely unavoidable.
```
Tatsinnit: One thing worth being explicit about in a section titled "Checks that gate a pull request": `test:scripts` and `docs:check` are **not** gates. I grepped every file under `.github/workflows` and nothing invokes either.

That's consistent with #2371's own note ("No CI gate is added yet; that is follow-up work"), so it isn't a contradiction — but it does mean the docs-tooling tests added there currently gate nothing. Since this page is now the canonical answer to "what will fail my PR", a line distinguishing *run these locally* from *these actually gate* would keep it accurate as gates get added later.

CONTRIBUTING.md
```diff
@@ -13,3 +11,49 @@ provided by the bot. You will only need to do this once across all repos using o
...
+User-facing documentation lives in `docs/book/` and is published to
+[azure.github.io/vscode-aks-tools](https://azure.github.io/vscode-aks-tools). Only
+`docs/book/**` is published, so that is the copy to edit.
+
+Pages under `docs/book/src/reference/` are generated from `package.json` and carry a
+`DO NOT EDIT` header. Regenerate them rather than editing by hand:
+
+```sh
+node scripts/generate-docs-reference.js
+node scripts/docs-check.js
```
Tatsinnit: Small consistency thing: `main` has `npm run docs:reference` and `npm run docs:check` wrappers for exactly these two (added in #2371), and every other command in this file is written as `npm run X`. Once rebased, these could be:

```sh
npm run docs:reference
npm run docs:check
```

Invoking via `node` works fine either way — just reads as inconsistent next to `npm run install:all`, `npm run prettier-format` and `npm run lint:all` above.

Tatsinnit (dismissed):
Reviewed the last commit (`87eeff41`) only, as directed — `CONTRIBUTING.md`, `docs/book/src/contributing.md`, `docs/book/src/development/development.md`.

## Verified

Checked every claim against the workflows and `package.json`:

| Claim | Evidence |
|---|---|
| CI pins Node 22 | `build.yml:48` -> `node-version: 22`; `1es-pipeline.yml` -> `nodeVersion default: 22.x`. Both cited lines exact. |
| Nothing enforces it locally | `engines` is `{"vscode":"^1.110.0"}` only; no `.nvmrc` or `.node-version`. |
| Bare `npm install` fails | `install:all` = `npm install --legacy-peer-deps && cd webview-ui && npm install --legacy-peer-deps`. |
| `webpack` doesn't package | `npm run build:webview && webpack --mode production` — no `vsix`. |
| `package` is the real one | `vsce package --no-dependencies`, and `build.yml:63` runs `npm run package`. |
| `test-compile` copies the template | `... && cp resources/yaml/aks-deploy.template.yaml out/src/commands/aksContainerAssist/`. |
| husky via `prepare` | `"prepare": "husky"`. |
| Build runs lint + test | `build.yml:53` `npm run lint:all`, `:55`/`:58` `npm run test`. |
| The two files were byte-identical | `diff` on the parent commit confirms it. |

I went looking specifically for the **`[skip pr-size]`** escape hatch, since documenting one that didn't exist would be worse than the gap it fills — it's real, at `restrict-pr-size.yml:47`, and was previously undocumented. Good catch.

All links resolve, including the two book-relative ones in `docs/book/src/contributing.md`.

## Main thing to catch during the rebase

Left inline on the script list: this branch's `package.json` has **17** scripts, `main` has **21**. The four added by #2371 aren't here yet, so after the rebase this page goes from 0 undocumented scripts back to 4 — undoing the main thing it achieves. Cheap to fix while rebasing, easy to miss.

## Minor

**"9 of 17 npm scripts undocumented"** — I count **8**: `lint:all`, `lint-fix:all`, `prettier-format`, `package`, `test:fuzz`, `lint`, `eslint-inspector`, `prepare`. After the change it's 0 of 17, so the improvement stands either way; just the before-number is off by one.

## Verdict

Approving the content. The corrections are accurate and well-sourced, and expanding the root `CONTRIBUTING.md` — which GitHub surfaces natively in the PR and issue UI — while reducing the book copy to a pointer is the right call for two files that were byte-identical and already drifting.

Can't merge yet regardless: `CONFLICTING`, and stacked on #2373 which is still open. Sequence looks like #2373 -> rebase this -> add the four missing scripts.

.github/workflows/docs-check.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      # Fails if package.json changed without regenerating docs/book/src/reference/.
+      # Fix with: npm run docs:reference
+      - name: Generated reference is up to date
+        run: npm run docs:reference:check
+
+      # Command IDs that do not exist, and menu breadcrumbs that do not match the
+      # menu package.json actually declares. Warnings do not fail the run.
+      - name: Documentation checks
+        run: npm run docs:check
```
Tatsinnit, line 64: FYI, not blocking: this PR adds 11 tests to `scripts/docs-check.test.js`, but nothing runs them. The workflow runs `docs:reference:check` and `docs:check`; Build runs `npm test`, which compiles only `src/` and loads `out/src/tests/suite`; Fuzzing runs only the compiled fuzz tests. So `npm run test:scripts` is invoked by no workflow, and a broken assertion in `scripts/docs-check.test.js` or `scripts/lib/menu-graph.test.js` still gets green CI.

The snag is that `test:scripts` needs mocha, which would mean an install step here and lose the "no install, ~20s" property that makes this workflow cheap. Probably easier to hang it off Build, which already runs `install:all`, than to add an install to this job.

Worth doing at some point since the parser tests are the main defence against regressions the current docs corpus doesn't happen to expose — but a follow-up is fine.

CONTRIBUTING.md
```diff
@@ -13,3 +11,59 @@ provided by the bot. You will only need to do this once across all repos using o
...
+npm run test:fuzz         # Fuzzing Tests workflow
+npm run docs:check        # Docs Check workflow
+npm run prettier-format   # Prettier Check workflow
+```
+
+**Build** and **Fuzzing Tests** run on every pull request. **Docs Check** runs on any
+pull request touching `docs/`, `scripts/`, `package.json`, `package.nls.json` or
+`resources/`. **Prettier Check** only runs when the pull request touches a `.ts` or
+`.tsx` file, but when it does run it checks `.json`, `.css` and `.md` too, so format
+everything before pushing rather than only what triggered it.
```
Tatsinnit, line 44: Small factual thing: the Prettier job's glob does include `.json` and `.md`, but `.prettierignore` excludes `docs/`, `**/*.json` and `**/*.md`, so those are never actually checked. Confirmed with `prettier --file-info`:

```
package.json           -> { "ignored": true }
docs/book/src/README.md -> { "ignored": true }
webview-ui/src/index.css -> { "ignored": false, "inferredParser": "css" }
src/extension.ts        -> { "ignored": false, "inferredParser": "typescript" }
```

So in practice it checks TS/TSX/CSS outside the ignored paths. The same wording is at `development.md:39`, and `npm run prettier-format` respects the same ignore file, so "format everything before pushing" can't fully happen either.

Doesn't change the advice — running `prettier-format` is still right — just the reason given for it.

Also minor, further down at line 68: "The **Docs Check** workflow runs both" follows `docs:reference` and `docs:check`, but the workflow runs `docs:reference:check` (it detects staleness rather than regenerating). Might be worth adding `docs:reference:check` to the pre-PR list too, since that's the one that actually fails.

scripts/docs-check.js
```diff
@@ -197,6 +198,81 @@ checks.identifiers = (report) => {
...
+ * Scoped to labels beginning `AKS:` on purpose. Every bold string is a candidate
+ * command name in principle, but most are ordinary emphasis, and guessing which
+ * is which produces false positives that train people to ignore the check. The
+ * `AKS:` prefix is only ever written when a command is meant, so it identifies
+ * the intent without needing an opt-out marker.
+ *
+ * Both forms are accepted. The Command Palette shows `category: title`, menus
+ * show the bare title, and pages legitimately use either.
+ */
+const COMMAND_LABEL = /\*\*(AKS:\s*[^*\n]+)\*\*/g;
```
Tatsinnit, line 219: FYI on where this check's edges are — the scoping rationale in the comment above is sound, I just want the limits written down somewhere since the docs page presents it fairly broadly.

**Inline code isn't scanned.** There's a live example: `docs/book/src/installation.md:9` says `` `AKS: Select tenant...` `` while `package.json` contributes `%Select Tenant...%` with `category: AKS`, i.e. `AKS: Select Tenant...`. Lowercase `t`. `docs:check` reports zero errors today because the label isn't bold. Inline code seems about as strong a "this is a command name" signal as bold, and `withoutFencedCode` already deliberately leaves it alone for `coverage`.

**`knownBare` accepts some names that don't exist.** Stripping `AKS:` from every command *and* submenu means these all pass:

```
AKS: Delete Cluster            # title is bare "Delete Cluster", no category
AKS: Develop & Deploy          # a submenu, never a command
AKS: Launch Kickstart Agent... # real title has no ellipsis
```

Only 12 of 66 commands carry `category: AKS`, so the bare set is doing a lot of work. Building the exact permitted display forms per command — the literal title, plus `category: title` only where a category exists, submenus excluded — would tighten it without changing what legitimately passes.

Neither is a regression; the check is strictly better than what was there. Just noting it catches the bold-and-renamed case rather than the general one.

Tatsinnit (approved):
Approving. This addresses everything from my earlier review — the four missing scripts are documented, `test:scripts` is listed, the contributor commands use the `npm run docs:*` aliases, and the new Docs Check workflow turns documentation drift into a real gate rather than something to remember. The `titles` check is a genuinely good addition; catching `AKS: Create Argo CD GitOps Pipeline` surviving a clean run after #2353 is exactly the kind of gap that's invisible until someone goes looking.

Three non-blocking FYIs, left inline. None of them affect the extension at runtime, and none is a reason to hold this up.

- Docs Check doesn't run `npm run test:scripts`, so the 11 new tests aren't gated by anything.
- The Prettier wording overstates what that job actually checks — `.prettierignore` excludes JSON and Markdown.
- The `titles` check is narrower than the surrounding prose implies. Worth knowing where the edges are; fine as a follow-up.

Verified along the way: the gate list matches the workflows (`build.yml:48` Node 22, `:53` lint, `:55`/`:58` test, fuzzing on every PR), `[skip pr-size]` reads the most recent commit as described (`restrict-pr-size.yml:45-47`), and `prepare-test-assets.js` does copy both the deploy template and the CA skills into `dist/skills`.

PR #2383, opened 2026-08-13
Route 1ES npm traffic through the CFS proxy

bosesuneha:
## Summary
- configure the 1ES job to use the CFS npm proxy
- redirect registry URLs from existing lockfiles at runtime without modifying them

### DIFF
--- a/.azurepipelines/1es-pipeline.yml
+++ b/.azurepipelines/1es-pipeline.yml
@@ -37,6 +37,9 @@ extends:
             - stage: Stage
               jobs:
                   - job: HostJob
+                    variables:
+                        NPM_CONFIG_REGISTRY: https://packagefeedproxy.microsoft.io/npm/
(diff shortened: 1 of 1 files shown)


.azurepipelines/1es-pipeline.yml
```diff
@@ -37,6 +37,9 @@ extends:
             - stage: Stage
               jobs:
                   - job: HostJob
+                    variables:
```
Tatsinnit, line 40: 💡 Thank you for this, I am guessing there is a missing bit about using this from the install side of code. Leaving this as a comment as fyi we can chat up. 

lets make sure we check for a sample so for now we could wither run sample: (Purely as a thought and what do you think?)

```
- script: |
      set -e
      npm config get registry
      npm config get replace-registry-host
  displayName: Verify npm registry configuration
```

Tatsinnit (approved):

PR #2393, opened 2026-08-24
fix(deps): replace unmaintained decompress with targeted archive extraction

Tatsinnit:
## Why

`decompress@4.2.1` carries **three critical advisories with no fix available**. It was last published in 2017 and is unmaintained:

| Advisory | Issue |
|---|---|
| [GHSA-mp2f-45pm-3cg9](https://github.com/advisories/GHSA-mp2f-45pm-3cg9) | Archive extraction can create files and links outside the target directory |
| [GHSA-h39j-r5qq-r9mm](https://github.com/advisories/GHSA-h39j-r5qq-r9mm) | Arbitrary file write via archive extraction (Zip Slip) |
| [GHSA-jwp9-9v96-94mx](https://github.com/advisories/GHSA-jwp9-9v96-94mx) | Arbitrary hardlink creation during archive extraction |

`npm audit` reports these as `critical` / `No fix available`, so no version bump resolves them — the dependency has to go.

It is used in exactly one place, `binaryDownloadHelper.ts`, to unpack the `kubelogin` and `kubectl-gadget` release archives.

## What changed

| Format | Before | After |
|---|---|---|
| `.zip` | `decompress` | [`yauzl`](https://github.com/thejoshwolfe/yauzl) — the reader behind `extract-zip`, `@vscode/vsce` and Electron |
| `.tar.gz` | `decompress` | [`tar`](https://github.com/isaacs/node-tar) — npm's own implementation |

**This PR adds zero packages to the dependency tree and removes 51.** Both `tar@7.5.21` and `yauzl@3.4.0` were already resolving transitively (via `containerization-assist-mcp` and `@vscode/vsce` respectively) at exactly these versions, so they simply become direct dependencies and dedupe. `npm ci` produces the same tree you're already running.

`yauzl` ships no types, and rather than pull in `@types/yauzl` this adds `src/types/yauzl.d.ts` — a ~50-line ambient declaration covering only the API actually used. Verified to type-check rather than degrade to `any`.

### The vulnerability class is removed, not mitigated

This is the part worth reviewing. The old code extracted the **entire archive** to disk, then moved the one file it wanted:

```ts
await decompress(downloadFilePath, downloadFolder);          // writes every path in the archive
const unzipped = path.join(downloadFolder, spec.pathToBinaryInArchive);
await moveFile(unzipped, binaryFilePath);
```

Every path inside the archive got written, which is exactly what makes Zip Slip and hardlink attacks work. The new code extracts **only the requested entry, streamed directly to a path we compute ourselves**:

```ts
await extractBinaryFromArchive(downloadFilePath, spec.pathToBinaryInArchive, partialFilePath);
```

A malicious archive has no path it can influence, so the bug class is structurally unreachable rather than patched.

### Two bugs the rewrite would otherwise have introduced

**Windows path separators.** Archive entries are always `/`-separated, but `pathToBinaryInArchive` is built with `path.join`, which yields `bin\windows_amd64\kubelogin.exe` on Windows. The old code sidestepped this by extracting to real filesystem paths; comparing entry names directly would have broken Windows silently. `toArchiveEntryPath()` normalises before comparing.

**Atomicity.** Writing straight to `binaryFilePath` would let a failed extraction leave a truncated binary behind — and `getToolBinaryPath()` opens with an `fs.existsSync(binaryFilePath)` cache check, so the next run would hand back the corrupt file. Extraction now goes to a `.partial` file and renames only on success, with cleanup on failure.

### Call sites unchanged

Archive format is inferred from the file extension, so `kubeloginDownload.ts` and `kubectlGadgetDownload.ts` are untouched. An unrecognised extension fails with a clear error.

Making it explicit (`archiveFormat: "zip" | "tar.gz"` on `ArchiveDownloadSpec`) would move that to compile time and is probably worth doing if more tools get added — it just wasn't worth widening this change. Happy to add it here if preferred.

## Verification

| Check | Result |
|---|---|
| `tsc -p ./ --noEmit` | exit 0 |
| `eslint` | exit 0 |
| `prettier --check` | clean |
| `webpack --mode production` | exit 0 |
| `npm ls decompress` | `(empty)` |
| `npm audit` | **3 critical → 0 critical** |
| Packages added / removed | **0 / 51** |

Beyond static checks, I exercised the real production `getToolBinaryPath()` with `vscode` stubbed, downloading actual GitHub releases at the versions this extension pins (`kubelogin v0.2.19`, `kubectl-gadget v0.53.2`):

```
PASS  kubelogin .zip   (nested entry bin/darwin_arm64/kubelogin)  sha=match mode=755 cleaned=true
PASS  gadget .tar.gz   (root entry kubectl-gadget)                sha=match mode=755 cleaned=true
PASS  missing entry rejected
```

Both digests are **byte-identical** to reference extraction via `unzip -p` and `tar xzO`, and both binaries execute — `kubelogin --version` reports `v0.2.19`, matching the configured release tag. Cache-hit path, archive cleanup, `.partial` cleanup, missing-entry and unsupported-extension cases all verified.

## On the lockfile diff

`package-lock.json` loses ~580 lines. That's a genuine reduction, not churn — `decompress` pulled in **51 packages**, many long-deprecated:

`decompress-tar`, `decompress-tarbz2`, `decompress-targz`, `decompress-unzip`, `seek-bzip`, `unbzip2-stream`, `tar-stream`, `buffer-alloc`, `buffer-fill`, `pinkie`, `pinkie-promise`, `is-natural-number`, `strip-dirs`, `fd-slicer`, plus nested duplicate copies of `readable-stream`, `safe-buffer`, `string_decoder`, `isarray` and an old `yauzl@2.10.0`.

The `package.json` diff itself is six lines. Note `npm install` also wanted to reformat the unrelated `chatSkills` array (~110 lines); since `.prettierignore` excludes `**/*.json` nothing enforces either style, so I reverted that and hand-applied only the dependency changes to keep this reviewable.

## Out of scope

`npm audit` still reports findings against `undici` and `hono`, both pre-existing and unrelated. Separately, `hono@4.13.3` (pulled in by #2380) is currently unavailable on the Microsoft package feed proxy — its latest is `4.13.2` — which breaks `npm ci` locally for anyone on that registry. Not caused by this PR, but worth a look.

### DIFF
--- a/src/commands/utils/helper/binaryDownloadHelper.ts
+++ b/src/commands/utils/helper/binaryDownloadHelper.ts
@@ -65,9 +65,28 @@ function extractFileFromZip(archivePath: string, entryPath: string, destPath: st
                 return;
             }
 
-            let found = false;
+            // Settle exactly once, always closing the archive first. An open handle
+            // leaks a file descriptor and, on Windows, blocks the caller from
+            // deleting the archive afterwards.
+            let settled = false;
+            const finish = (error?: Error) => {
+                if (settled) {
+                    return;
+                }
+                settled = true;
+                try {
+                    zipFile.close();
+                } catch {
+                    // Already closed, or never fully opened; nothing to release.
+                }
+                if (error) {
+                    reject(error);
+                } else {
+                    resolve();
+                }
+            };
 
-            zipFile.on("error", reject);
+            zipFile.on("error", finish);
 
             zipFile.on("entry", (entry: yauzl.Entry) => {
                 if (toArchiveEntryPath(entry.fileName) !== wantedEntry) {
@@ -75,25 +94,20 @@ function extractFileFromZip(archivePath: string, entryPath: string, destPath: st
                     return;
                 }
 
-                found = true;
+                // First match wins: we stop requesting entries, so a duplicate name
+                // later in the archive is never read.
                 zipFile.openReadStream(entry, (streamError, readStream) => {
                     if (streamError || !readStream) {
-                        reject(streamError ?? new Error(`Failed to read ${wantedEntry} from archive.`));
+                        finish(streamError ?? new Error(`Failed to read ${wantedEntry} from archive.`));
                         return;
                     }
 
-                    pipeline(readStream, fs.createWriteStream(destPath)).then(() => {
-                        zipFile.close();
-                        resolve();
-                    }, reject);
+                    pipeline(readStream, fs.createWriteStream(destPath)).then(() => finish(), finish);
                 });
             });
 
-            zipFile.on("end", () => {
-                if (!found) {
-                    reject(new Error(`Archive does not contain an entry named ${wantedEntry}.`));
-                }
-            });
+            // Only reached when no entry matched, since a match stops the iteration.
+            zipFile.on("end", () => finish(new Error(`Archive does not contain an entry named ${wantedEntry}.`)));
 
             zipFile.readEntry();
         });
@@ -111,8 +125,11 @@ async function extractFileFromTarball(archivePath: string, entryPath: string, de
     await tar.list({
         file: archivePath,
         onReadEntry: (entry) => {
-            if (toArchiveEntryPath(entry.path) !== wantedEntry) {
-                // Leave it unconsumed; tar drains entries we don't attach to.
+            // First match wins. A tar is append-only, so the same path may legally
+            // appear more than once; attaching twice would start two concurrent
+            // writes to `destPath` and interleave them. Entries we don't attach to
+            // are drained by tar, so parsing continues either way.
+            if (writeCompleted || toArchiveEntryPath(entry.path) !== wantedEntry) {
                 return;
             }
 
@@ -196,8 +213,13 @@ async function downloadTool(
                 error: `Failed to extract ${downloadSpec.pathToBinaryInArchive} from ${downloadFilePath}: ${error}`,
             };
         } finally {
-            // Remove the archive whether or not extraction succeeded.
+            // Remove the archive whether or not extraction succeeded, and drop the
+            // download-once marker along with it. `download.once()` keeps an
+            // in-memory Completed flag per destination: deleting the file while that
+            // flag stands would make every retry skip the download and then fail on
+            // the missing archive, for the lifetime of the window.
             fs.rmSync(downloadFilePath, { force: true });
+            download.clear(downloadFilePath);
         }
     } else {
         await moveFile(downloadFilePath, binaryFilePath);
--- a/src/commands/utils/helper/binaryDownloadHelper.ts
+++ b/src/commands/utils/helper/binaryDownloadHelper.ts
@@ -4,6 +4,9 @@ import * as fs from "fs";
 import { moveFile } from "move-file";
 import { Errorable, failed } from "../errorable";
 import path from "path";
+import { pipeline } from "stream/promises";
+import * as tar from "tar";
+import * as yauzl from "yauzl";
 import { longRunning } from "../host";
 
 function getToolBaseInstallFolder(toolName: string): string {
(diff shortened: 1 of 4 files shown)


src/commands/utils/helper/binaryDownloadHelper.ts
```diff
@@ -73,23 +180,25 @@ async function downloadTool(
...
         } catch (error) {
+            fs.rmSync(partialFilePath, { force: true });
             return {
                 succeeded: false,
-                error: `Failed to unzip binary ${downloadFilePath} to ${downloadFolder}: ${error}`,
+                error: `Failed to extract ${downloadSpec.pathToBinaryInArchive} from ${downloadFilePath}: ${error}`,
             };
+        } finally {
+            // Remove the archive whether or not extraction succeeded.
+            fs.rmSync(downloadFilePath, { force: true });
```
bosesuneha, line 221:
```suggestion
            fs.rmSync(downloadFilePath, { force: true });
            download.clear(downloadFilePath);
```

bosesuneha, line 221: the finally deletes the archive on failure too, but download.clear() at line 208 is skipped by the early return in the catch. That leaves download.once marked Completed with the file gone, so a retry skips the download and fails until VS Code restarts. On main the archive survived a failed extraction, so retries still worked.

Tatsinnit: Confirmed, and you've got the mechanism exactly right — including that main behaved differently. Reproduced it before fixing:

```
attempt 1 (bad entry) succeeded: false
archive still on disk after failure: []          <- finally removed it
attempt 2 (good entry) succeeded: false
  error: Failed to extract ... /download/kubelogin-darwin-arm64.zip
RETRY IS PERMANENTLY BROKEN
```

`download.once` keeps `DOWNLOAD_ONCE_STATUS[dest] = Completed` in module state, so the second call took the `else` branch, returned success without downloading, and then extraction failed on a file the `finally` had already deleted. Permanent for the life of the window, since that map is in-memory.

And your point about main is right: it called `fs.unlinkSync` only after a successful `decompress`, so a failed extraction left the archive in place and a retry could still use it. The `finally` I added is what broke that.

Fixed as you suggested — clearing the marker alongside the removal so the two always move together:

```ts
} finally {
    fs.rmSync(downloadFilePath, { force: true });
    download.clear(downloadFilePath);
}
```

Same repro now:
```
attempt 1 (bad entry) succeeded: false
attempt 2 (good entry) succeeded: true
  sha256: MATCH
RETRY RECOVERS
```

Thanks — this was the most consequential of the four and I'd missed it entirely. Fixed in f7858c27.

src/commands/utils/helper/binaryDownloadHelper.ts
```diff
@@ -37,6 +40,109 @@ function isArchive(downloadSpec: DownloadSpec): downloadSpec is ArchiveDownloadS
...
+                    zipFile.readEntry();
+                    return;
+                }
+
+                found = true;
+                zipFile.openReadStream(entry, (streamError, readStream) => {
+                    if (streamError || !readStream) {
+                        reject(streamError ?? new Error(`Failed to read ${wantedEntry} from archive.`));
+                        return;
+                    }
```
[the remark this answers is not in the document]
Tatsinnit: Legitimate — the ZipFile was only closed on the success path. `openReadStream` failure, pipeline rejection, and entry-not-found all rejected with the handle still open, which leaks a descriptor and on Windows would block the caller's `finally` from deleting the archive.

Settling now goes through a single idempotent `finish()` that always closes first:

```ts
let settled = false;
const finish = (error?: Error) => {
    if (settled) return;
    settled = true;
    try { zipFile.close(); } catch { /* already closed */ }
    if (error) reject(error); else resolve();
};
```

All three paths route through it. The `found` flag is gone too — a match stops the iteration, so `end` only fires when nothing matched.

Verified with `lsof`: 12 consecutive failed extractions leave the descriptor count unchanged (27 before, 27 after). Fixed in f7858c27.

src/commands/utils/helper/binaryDownloadHelper.ts
```diff
@@ -37,6 +40,109 @@ function isArchive(downloadSpec: DownloadSpec): downloadSpec is ArchiveDownloadS
...
+    await tar.list({
+        file: archivePath,
+        onReadEntry: (entry) => {
+            if (toArchiveEntryPath(entry.path) !== wantedEntry) {
+                // Leave it unconsumed; tar drains entries we don't attach to.
+                return;
+            }
+
+            writeCompleted = pipeline(entry, fs.createWriteStream(destPath));
+        },
```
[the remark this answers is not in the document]
Tatsinnit: Legitimate. A tar is append-only so the same path can legally appear more than once, and reassigning `writeCompleted` started a second `createWriteStream` to the same destination while the first was still going.

Built an archive containing `payload` twice to check the actual behaviour:

```
fixed  -> "FIRST-COPY-AAAA\n" (16 bytes)
old    -> "SECOND-COPY-BBBBBBBBBBBBBBBB\n" (29 bytes)
```

So it wasn't hypothetical — the previous version silently emitted the second copy. Guarded so the first match wins and later duplicates are left to drain:

```ts
if (writeCompleted || toArchiveEntryPath(entry.path) !== wantedEntry) {
    return;
}
```

Worth noting the zip path was already safe here for a different reason: it stops calling `readEntry()` after a match, so a duplicate name later in the archive is never read. Fixed in f7858c27.

src/commands/utils/helper/binaryDownloadHelper.ts
Tatsinnit: Applied — this is the same finding as your other comment, taking the suggestion as written. `download.clear()` now sits next to the removal in the `finally` so the file and its download-once marker always move together. Details and the before/after repro are in the reply on the other thread. Fixed in f7858c27.

bosesuneha (approved):
lgtm

PR #2410, opened 2026-09-03
ci: run the documentation checks on pull requests

bosesuneha:
`docs:check` and `docs:reference` are not invoked by anything. `website.yaml` only builds the book, and only on push to `main`. So the drift they exist to prevent can recur on the next PR, and did: the Draft page named a command that had been renamed months earlier, and a clean `docs:check` run never said so.

Two steps, no install (the scripts use only node builtins):

| Step | Fails when |
|---|---|
| `npm run docs:reference:check` | `package.json` changed without regenerating `docs/book/src/reference/` |
| `npm run docs:check` | unknown command IDs or names, or menu breadcrumbs that do not match the menu `package.json` declares |

Triggers on `docs/**`, `scripts/**`, `package.json`, `package.nls.json` and `resources/**`, which is every input the checks read.

## Why now

This was blocked until #2373 merged. `docs:check` failed on `main` with 2 `menu-syntax` errors in `merge-save-kubeconfig.md`; #2373 fixed both. `main` is green as of 8a4367f2, so this can go in without turning it red.

Verified against `main` at 834195f3:

```
[ok] identifiers   [ok] orphans   [ok] menu-paths   [ok] menu-syntax
0 error(s), 15 warning(s)
generated reference is up to date
```

Warnings do not fail the run. The 15 are `coverage` — commands with no prose yet.

Also aligns the `harden-runner` pin with the other five workflows (v2.20.0 → v2.20.1).

## Ordering

Independent of #2382, but they complement each other: #2382 documents these as commands to run by hand and adds a `titles` check. If both land, the CONTRIBUTING line saying neither runs in CI should be dropped — happy to do that in whichever merges second.

### DIFF
--- a/.github/workflows/docs-check.yml
+++ b/.github/workflows/docs-check.yml
@@ -1,0 +1,64 @@
+name: Docs Check
+
+# The documentation restates package.json in prose: command titles, menu paths,
+# setting defaults, pinned tool versions. Nothing verified those restatements
+# before merge, so they drifted. These checks run whenever either side changes.
+#
+# The scripts have no npm dependencies, so there is no install step.
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "docs/**"
+      - "scripts/**"
+      - "package.json"
+      - "package.nls.json"
+      - "resources/**"
+      - ".github/workflows/docs-check.yml"
+  push:
+    branches:
+      - main
+    paths:
+      - "docs/**"
+      - "scripts/**"
+      - "package.json"
+      - "package.nls.json"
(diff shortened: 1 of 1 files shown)


bosesuneha:
Folded into #2382 as commit `7c19c709`, unchanged.

Keeping them apart guaranteed a window where one was wrong: #2382's CONTRIBUTING said the docs checks run nowhere, and this PR makes them run. Landing together means the workflow and the page describing it are consistent on arrival.

It is still its own commit in #2382, so it can be reverted independently if you would rather not adopt the gate yet.

PR #2420, opened 2026-09-14
fix(kickstart): fix repo picker ordering and search, plus surrounding UI defects

bosesuneha:
Fixes the Kickstart GitHub repo picker, plus two defects found in the surrounding UI while reviewing it.

## Repo picker

**Ordering** — the list sorted by `updated`, which GitHub bumps on any change to the repo record (stars, description edits, renames), not on commits. Switched to `pushed` with an explicit `desc` direction, so the order reflects recent work. Rows now show a relative last-push time. Also pages past the first 100 repos and flags when the list is capped.

**Search** — the list used `CustomDropdown`, whose type-ahead only matches from the start of the label, so `vscode-aks-tools` could not be found by typing `aks`. Adds `SearchableDropdown`, which filters fuzzily over name and description and renders name, last-push, visibility and description as separate fields instead of one concatenated string.

**List vs. paste** — both controls wrote the same state, so picking from the list also filled the text box and the `or` divider between them was meaningless. They are now tracked separately, each clears the other, and the paste field is labelled as the fallback it is.

**URL handling** — the paste field accepted anything non-empty, so a `/tree/<branch>/<path>` link copied from the address bar went straight to `git clone`. URLs are now normalized (bare `owner/repo`, missing scheme, trailing slash, GitHub web subpaths) and validated. Stays permissive about host, since the agent clones with plain `git`.

**Guided setup, same flow** — opening the wizard preselected the GitHub source and immediately requested the `repo` scope. Nothing is preselected now, the repo load is silent, and an explicit sign-in button is the only path to the consent dialog. A failed chat handoff also left the user on a permanent "Opening the Kickstart chat…" message, because the stage flipped optimistically and unmounted the only component that rendered the error, which was never posted anyway. `AppSource` is now a discriminated union so `undefined` cannot reach the agent's seed prompt.

## Cluster view

**Form state lost on "Back to setup"** — `Stage.Provisioning` unmounts `ClusterInput`, so every field lived in state that was destroyed on the way to the provisioning screen. After a failed run the user returned to an empty form, with the auto-derived resource names regenerated under a different random suffix, so they could not reproduce what they had. Form fields, the suffix, and the user-edited flags move into the reducer.

## Shared dropdown

`TextWithDropdown` accepted an `id` prop but never rendered it, so six `htmlFor` attributes pointed at nothing and clicking a label did nothing. Also completes the combobox ARIA contract, which declared `role="combobox"` with no `aria-expanded`, no `role="listbox"` on the popup and no `role="option"` on its items.


pauldotyu (approved):
Looks like a decent amount of cleanup, thanks @bosesuneha!!

